PR #18, opened 2025-02-25
Use Worker Deployment APIs

carlydf:
## What is the change?
- Converted everything to use Worker Deployment Versions / Worker Deployments instead of Deployments
- Changed ReachabilityStatus to VersionStatus (not registered -> inactive -> ramping -> current -> draining -> drained -> not registered)
- Indexed everything on Version ID instead of Build ID.

## How was it tested?
Tested using skaffold and minikube against a temporal cloud cluster.
Tested instant cutover and progressive strategies.

Not tested: Manual strategy


internal/controller/util.go
```diff
@@ -15,30 +15,39 @@ import (
 	"github.com/DataDog/temporal-worker-controller/internal/controller/k8s.io/utils"
 )
 
-func findHighestPriorityStatus(statuses []temporaliov1alpha1.ReachabilityStatus) temporaliov1alpha1.ReachabilityStatus {
+func findHighestPriorityStatus(statuses []temporaliov1alpha1.VersionStatus) temporaliov1alpha1.VersionStatus {
```
jlegrone: Dead code, please feel free to delete 🕊️

api/v1alpha1/worker_types.go
```diff
@@ -66,26 +67,37 @@ type TemporalWorkerSpec struct {
...
+	// worker deployment. It is accepting all new workflow executions except for the
+	// percent that are sent to the ramping version, if one exists.
+	VersionStatusCurrent = "Current"
+
+	// VersionStatusDraining indicates that the version has stopped accepting new workflows
+	// (is no longer ramping or current) and DOES have open workflows pinned to it.
+	VersionStatusDraining VersionStatus = "Draining"
+
+	// VersionStatusDrained indicates that the version has stopped accepting new workflows
+	// (is no longer ramping or current) and does NOT have open workflows pinned to it.
```
jlegrone, line 102: Queries for closed workflows can still be routed to a version with this status right? Please feel free to play with the wording, but I'd like to mention that in the godoc comment.

```suggestion
	// VersionStatusDrained indicates that the version has stopped accepting new workflows
	// (is no longer ramping or current) and does NOT have open workflows pinned to it.
	// This version MAY still receive query tasks associated with closed workflows.
```

api/v1alpha1/worker_types.go
```diff
@@ -94,23 +106,29 @@ type TemporalWorkerStatus struct {
 	// so it’s generally not a good idea to read from the status of the root object.
 	// Instead, you should reconstruct it every run.
 
-	// TargetVersion is the desired next version. If the deployment is nil,
+	// TargetVersion is the desired next version. If the version is nil,
```
jlegrone: This was actually referring to the `TargetVersion.Deployment` field. I'm not sure if these doc comments were actually necessary since they're just describing the behavior of the controller, but it doesn't hurt to keep them.

```suggestion
	// TargetVersion is the desired next version. If TargetVersion.Deployment is nil,
```

api/v1alpha1/worker_types.go
```diff
@@ -94,23 +106,29 @@ type TemporalWorkerStatus struct {
...
+	// wait for it to become healthy and then move it to the DefaultVersion.
+	TargetVersion *WorkerDeploymentVersion `json:"targetVersion"`
 
-	// DefaultVersion is the deployment that is currently registered with
-	// Temporal as the default. This must never be nil.
+	// TargetVersionRampPercentage is the ramp percentage of the target version.
+	TargetVersionRampPercentage float32 `json:"targetVersionRampPercentage"`
+
+	// TargetVersionRampingSinceTime is time when the target version first started ramping.
+	TargetVersionRampingSinceTime *metav1.Time `json:"targetVersionRampingSinceTime"`
```
jlegrone: Nit: remove the `Time` suffix in the field name since the type is already `metav1.Time`. This also matches the naming of the `HealthySince` field.

```suggestion
	// TargetVersionRampingSince is time when the target version first started ramping.
	TargetVersionRampingSince *metav1.Time `json:"targetVersionRampingSinceTime"`
```

README.md
```diff
@@ -28,22 +28,29 @@ default version after a deployment.
...
 
 In order to be compatible with this controller, workers need to be configured using these standard environment
 variables:
 
 - `TEMPORAL_HOST_PORT`: The host and port of the Temporal server, e.g. `default.foo.tmprl.cloud:7233`
 - `TEMPORAL_NAMESPACE`: The Temporal namespace to connect to, e.g. `default`
-- `TEMPORAL_DEPLOYMENT_SERIES`: The name of the deployment series. This must be unique to the worker deployment and should not
+- `TEMPORAL_DEPLOYMENT_NAME`: The name of the worker deployment. This must be unique to the worker deployment and should not
   change between versions.
 - `WORKER_BUILD_ID`: The build ID of the worker. This should change with each new worker rollout.
```
carlydf, line 55: change this to `WORKER_DEPLOYMENT_VERSION`

internal/controller/util.go
```diff
@@ -7,36 +7,15 @@ package controller
...
-		return 2
-	case temporaliov1alpha1.ReachabilityStatusNotRegistered:
-		return 1
-	}
-	return 0
+func computeVersionID(spec *temporaliov1alpha1.TemporalWorkerSpec) string {
+	return spec.WorkerOptions.DeploymentName + "." + computeBuildID(spec)
 }
 
 func computeBuildID(spec *temporaliov1alpha1.TemporalWorkerSpec) string {
```
carlydf, line 21: follow-up PR: include user-passed build id / version id in this as a human-readable portion

README.md
carlydf: actually, this should stay as `TEMPORAL_DEPLOYMENT_NAME` (provided by user) since the build id is generated from the worker spec. Later when we provide an option to specify custom build id, then it could make sense to offer `VERSION` too.

Regardless it's a discussion for another PR.

api/v1alpha1/worker_types.go
carlydf: done!

README.md
```diff
@@ -28,22 +28,29 @@ default version after a deployment.
 
 - [x] Registration of new worker versions
 - [x] Creation of versioned worker deployment resources
-- [x] Deletion of unreachable worker deployments
+- [x] Deletion of drained worker deployments
 - [x] Manual, Blue/Green, and Progressive rollouts of new worker versions
 - [ ] Autoscaling of worker deployments
 - [ ] Automated rollover to compatible worker versions
```
carlydf, line 34: @jlegrone do you remember what is this referring to?

README.md
```diff
@@ -28,22 +28,29 @@ default version after a deployment.
...
 - [ ] Canary analysis of new worker versions
```
carlydf, line 35: is this bullet point referring to analysis beyond just checking that the canary workflow completed successfully?

internal/controller/genstatus.go
```diff
@@ -247,98 +250,104 @@ func (r *TemporalWorkerReconciler) generateStatus(ctx context.Context, l logr.Lo
 	sort.SliceStable(childDeploys.Items, func(i, j int) bool {
 		return childDeploys.Items[i].ObjectMeta.CreationTimestamp.Before(&childDeploys.Items[j].ObjectMeta.CreationTimestamp)
 	})
-	// Track each deployment by build ID
+	// Track each k8s deployment by version ID
 	for _, childDeploy := range childDeploys.Items {
 		if buildID, ok := childDeploy.GetLabels()[buildIDLabel]; ok {
-			versions.addDeployment(buildID, &childDeploy)
-			deployedBuildIDs = append(deployedBuildIDs, buildID)
+			versionID := workerDeploymentName + "." + buildID
```
carlydf, line 256: I think this is wrong, because we don't know that all of the (k8s) deployments managed by the controller correspond to the same Temporal Worker Deployment. So it's inaccurate to construct a version ID for each one from the build ID like this.

Also, if a user wants to deploy multiple TemporalWorker objects in the same controller-managed k8s cluster but polling different Temporal Namespaces, then `Build ID` or even `Version ID` is not a unique identifier, since version is only unique within 1 temporal namespace. It seems like to manage which k8s deployment is talking to which Temporal ns, and to figure out which k8s deployment is expected to have their version registered in which worker deployment, we need to label them with more than just build id (we need Temporal Worker Deployment Name and Temporal Namespace Name as well).

README.md
```diff
@@ -28,37 +28,46 @@ default version after a deployment.
...
 - [ ] Optional cancellation after timeout for workflows on old versions
 - [ ] Passing `ContinueAsNew` signal to workflows on old versions
 
+## Terminology
+Note that in Temporal, "Worker Deployment" is sometimes referred to as "Deployment", but since the controller makes
+significant references to Kubernetes Deployment resource, within this repository we will stick to these terms:
+- **Worker Deployment Version**: A version of a deployment or service. It can have multiple Workers, but they all run the same build. Sometimes shortened to "version" or "deployment version."
```
jlegrone, line 42: Is this the same thing?

```suggestion
- **Worker Deployment Version**: A version of a deployment or service. It can have multiple task queues, but they all run the same build. Sometimes shortened to "version" or "deployment version."
```

api/v1alpha1/temporalworker_webhook.go
```diff
@@ -23,7 +23,7 @@ var _ webhook.Defaulter = &TemporalWorker{}
 
 // Default implements webhook.Defaulter so a webhook will be registered for the type
 func (r *TemporalWorker) Default() {
-	if r.Spec.WorkerOptions.DeploymentSeries == "" {
-		r.Spec.WorkerOptions.DeploymentSeries = fmt.Sprintf("%s.%s", r.GetName(), r.GetNamespace())
+	if r.Spec.WorkerOptions.DeploymentName == "" {
+		r.Spec.WorkerOptions.DeploymentName = fmt.Sprintf("%s-%s", r.GetName(), r.GetNamespace())
```
jlegrone, line 27: Is this change because of restrictions on deployment names on the Temporal server side? If so, is there a character other than `-` that could be used?

My bad for not putting a comment here before, but the `.` separator was chosen over `-` because of the RFC 1123 restriction on k8s namespace names: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names

Specifically k8s namespace names must `contain only lowercase alphanumeric characters or '-'`. So using any separator _other than_ `-` is necessary in order to disambiguate between deployment `foo` in namespace `bar-baz` and deployment `foo-bar` in namespace `baz`.

README.md
Shivs11: I think so, but in this context it's best to use `Workers` since it goes well in tandem with `same build`. Any other specific reason you had in mind to replace this with task-queues?

api/v1alpha1/temporalworker_webhook.go
Shivs11: That's right - In the temporal server right now, there is a restriction that `DeploymentName` cannot have `.` present.

Hmm, the deploymentName (or buildID for that matter of fact) cannot also start with `__` and should be validUTF8 strings - we could maybe use another separator like `/`?

README.md
jlegrone: Makes sense, I think we can revisit clarifications in the readme after this is merged. I think I just was a little unclear about the usage of "Worker" in this sentence, and maybe being more explicit that multiple task queues could be polled by the same process could help.

api/v1alpha1/temporalworker_webhook.go
jlegrone: Thanks, we'll change it to `/` for now. Also chatted with @carlydf and we're going to revisit to see if we can come up with a solution that doesn't require escaping URLs in the Temporal UI in the future though.

internal/controller/genstatus.go
Shivs11: I noticed this too as I was reviewing the code but for the sake of simplicity had chosen to ignore this. I think for the time being, we can assume a 1:1 relationship between the TemporalWorker and a Temporal Worker Deployment but you are right that we have to tackle this problem very soon

Issue #2, opened 2024-07-23
Rollout Strategy: Manual


jlegrone:
Implement a rollout strategy where the controller treats Temporal worker versioning APIs as "read-only".

In this mode the controller should still create/scale/delete deployment resources as needed to satisfy the build ID assignment rules and build ID reachability statuses for a task queue. But it should not drive updates to build ID assignment rules directly; instead these are managed by external systems or human operators.

PR #22, opened 2025-03-28
add scaledown delay before deleting deployments 

Shivs11:
With this PR:

1. Deleting a deployment is only possible when:
a) is has been scaled down to have 0 replicas
b) the deployment has been drained for deleteDelay + scaledownDelay.

2. Unit test refactor and additions for `genPlan`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ flowchart TD
       twd[TemporalWorkerDeployment 'foo']
       
       subgraph "Current/default version"
-        d5["Deployment foo-v5 (Version ns/foo.v5)"]
+        d5["Deployment foo-v5 (Version foo/ns.v5)"]
         rs5["ReplicaSet foo-v5"]
         p5a["Pod foo-v5-a"]
         p5b["Pod foo-v5-b"]
@@ -88,7 +88,7 @@ flowchart TD
       end
 
       subgraph "Deprecated versions"
-        d1["Deployment foo-v1 (Version ns/foo.v1)"]
+        d1["Deployment foo-v1 (Version foo/ns.v1)"]
         rs1["ReplicaSet foo-v1"]
         p1a["Pod foo-v1-a"]
         p1b["Pod foo-v1-b"]
@@ -104,12 +104,12 @@ flowchart TD
     twd --> dN
     twd --> d5
 
-    p1a -. "poll version ns/foo.v1" .-> server
-    p1b -. "poll version ns/foo.v1" .-> server
+    p1a -. "poll version foo/ns.v1" .-> server
+    p1b -. "poll version foo/ns.v1" .-> server
 
-    p5a -. "poll version ns/foo.v5" .-> server
-    p5b -. "poll version ns/foo.v5" .-> server
-    p5c -. "poll version ns/foo.v5" .-> server
+    p5a -. "poll version foo/ns.v5" .-> server
+    p5b -. "poll version foo/ns.v5" .-> server
+    p5c -. "poll version foo/ns.v5" .-> server
 
     server["Temporal Server"]
 ```
@@ -147,14 +147,14 @@ sequenceDiagram
     Dev->>K8s: Create TemporalWorkerDeployment "foo" (v1)
     K8s-->>Ctl: Notify TemporalWorkerDeployment "foo" created
     Ctl->>K8s: Create Deployment "foo-v1"
-    Ctl->>T: Register "ns/foo.v1" as new current version of "ns/foo"
+    Ctl->>T: Register "foo/ns.v1" as new current version of "foo/ns"
     Dev->>K8s: Update TemporalWorker "foo" (v2)
     K8s-->>Ctl: Notify TemporalWorker "foo" updated
     Ctl->>K8s: Create Deployment "foo-v2"
-    Ctl->>T: Register "ns/foo.v2" as new current version of "ns/foo"
+    Ctl->>T: Register "foo/ns.v2" as new current version of "foo/ns"
     
     loop Poll Temporal API
-        Ctl-->>T: Wait for "ns/foo.v1" to be drained (no open pinned wfs)
+        Ctl-->>T: Wait for "foo/ns.v1" to be drained (no open pinned wfs)
     end
     
     Ctl->>K8s: Delete Deployment "foo-v1"
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ flowchart TD
       twd[TemporalWorkerDeployment 'foo']
       
       subgraph "Current/default version"
-        d5["Deployment foo-v5 (Version foo/ns.v5)"]
+        d5["Deployment foo-v5 (Version ns/foo.v5)"]
         rs5["ReplicaSet foo-v5"]
         p5a["Pod foo-v5-a"]
         p5b["Pod foo-v5-b"]
@@ -88,7 +88,7 @@ flowchart TD
       end
 
       subgraph "Deprecated versions"
-        d1["Deployment foo-v1 (Version foo/ns.v1)"]
+        d1["Deployment foo-v1 (Version ns/foo.v1)"]
         rs1["ReplicaSet foo-v1"]
         p1a["Pod foo-v1-a"]
         p1b["Pod foo-v1-b"]
@@ -104,12 +104,12 @@ flowchart TD
     twd --> dN
     twd --> d5
 
-    p1a -. "poll version foo/ns.v1" .-> server
-    p1b -. "poll version foo/ns.v1" .-> server
+    p1a -. "poll version ns/foo.v1" .-> server
+    p1b -. "poll version ns/foo.v1" .-> server
 
-    p5a -. "poll version foo/ns.v5" .-> server
-    p5b -. "poll version foo/ns.v5" .-> server
-    p5c -. "poll version foo/ns.v5" .-> server
+    p5a -. "poll version ns/foo.v5" .-> server
+    p5b -. "poll version ns/foo.v5" .-> server
+    p5c -. "poll version ns/foo.v5" .-> server
 
     server["Temporal Server"]
 ```
@@ -147,14 +147,14 @@ sequenceDiagram
     Dev->>K8s: Create TemporalWorkerDeployment "foo" (v1)
     K8s-->>Ctl: Notify TemporalWorkerDeployment "foo" created
     Ctl->>K8s: Create Deployment "foo-v1"
-    Ctl->>T: Register "foo/ns.v1" as new current version of "foo/ns"
+    Ctl->>T: Register "ns/foo.v1" as new current version of "ns/foo"
     Dev->>K8s: Update TemporalWorker "foo" (v2)
     K8s-->>Ctl: Notify TemporalWorker "foo" updated
     Ctl->>K8s: Create Deployment "foo-v2"
-    Ctl->>T: Register "foo/ns.v2" as new current version of "foo/ns"
+    Ctl->>T: Register "ns/foo.v2" as new current version of "ns/foo"
     
     loop Poll Temporal API
-        Ctl-->>T: Wait for "foo/ns.v1" to be drained (no open pinned wfs)
+        Ctl-->>T: Wait for "ns/foo.v1" to be drained (no open pinned wfs)
     end
     
     Ctl->>K8s: Delete Deployment "foo-v1"
--- a/internal/controller/genstatus.go
+++ b/internal/controller/genstatus.go
@@ -29,6 +29,7 @@ import (
 	temporaliov1alpha1 "github.com/DataDog/temporal-worker-controller/api/v1alpha1"
 )
 
+// TODO (Shivam): Should we be having a map of [versionID] -> Structure?
 type deploymentVersionCollection struct {
 	versionIDsToDeployments map[string]*appsv1.Deployment
 	// map of version IDs to ramp percentages [0,100]
@@ -388,7 +389,7 @@ func (r *TemporalWorkerDeploymentReconciler) generateStatus(ctx context.Context,
 		case desiredVersionID, defaultVersionID:
 			continue
 		}
-		d, _ := versions.getWorkerDeploymentVersion(version)
+		d, _ := versions.getWorkerDeploymentVersion(version) // TODO (Shivam): How do we know these versions have been deleted by Temporal? They could just be draining...
 		deprecatedVersions = append(deprecatedVersions, d)
 	}
 
@@ -399,7 +400,7 @@ func (r *TemporalWorkerDeploymentReconciler) generateStatus(ctx context.Context,
 
 	// Ugly hack to clear ramp percentages (not quite correctly) for now
 	for _, d := range deprecatedVersions {
-		d.RampPercentage = nil
+		d.RampPercentage = nil // TODO (Shivam): All deprecatedVersions will have rampPercentage set to nil already
 	}
 	if defaultVersion != nil {
 		defaultVersion.RampPercentage = nil
--- a/internal/controller/genstatus.go
+++ b/internal/controller/genstatus.go
@@ -8,11 +8,11 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"go.temporal.io/api/serviceerror"
-	"math"
 	"sort"
 	"strings"
 	"time"
+
+	"go.temporal.io/api/serviceerror"
 
 	"github.com/go-logr/logr"
 	"go.temporal.io/api/enums/v1"
@@ -139,13 +139,11 @@ func (c *deploymentVersionCollection) addAssignmentRule(rule *taskqueue.BuildIdA
 
 func (c *deploymentVersionCollection) addVersionStatus(version string, status temporaliov1alpha1.VersionStatus) {
 	c.versionStatus[version] = status
-	return
 }
 
 func (c *deploymentVersionCollection) addDrainedSince(version string, drainedSince time.Time) {
 	t := metav1.NewTime(drainedSince)
 	c.drainedSince[version] = &t
-	return
 }
 
 func (c *deploymentVersionCollection) addTaskQueue(versionID, name string) {
@@ -400,7 +398,7 @@ func (r *TemporalWorkerDeploymentReconciler) generateStatus(ctx context.Context,
 
 	// Ugly hack to clear ramp percentages (not quite correctly) for now
 	for _, d := range deprecatedVersions {
-		d.RampPercentage = nil // TODO (Shivam): All deprecatedVersions will have rampPercentage set to nil already
+		d.RampPercentage = nil
 	}
 	if defaultVersion != nil {
 		defaultVersion.RampPercentage = nil
@@ -419,8 +417,4 @@ func (r *TemporalWorkerDeploymentReconciler) generateStatus(ctx context.Context,
 		DeprecatedVersions:   deprecatedVersions,
 		VersionConflictToken: []byte("todo"),
 	}, nil
-}
-
-func convertFloatToUint(val float32) uint8 {
-	return uint8(math.Round(float64(val)))
 }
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -76,7 +76,7 @@ flowchart TD
       twd[TemporalWorkerDeployment 'foo']
       
       subgraph "Current/default version"
-        d5["Deployment foo-v5 (Version foo/ns.v5)"]
+        d5["Deployment foo-v5 (Version ns/foo.v5)"]
```
carlydf: nit: I think the name change could go in a separate PR since it's independent of the other changes, but not a big deal

carlydf: Aside from review clarity, another good reason to make them separate is that on the off chance there is a regression introduced by the scaledown delay PR (which is a logic change), it's nice to be able to revert that commit without also reverting the name change, which has no logic change and therefore not risky / we know we want it.

internal/controller/genstatus.go
```diff
@@ -399,7 +400,7 @@ func (r *TemporalWorkerDeploymentReconciler) generateStatus(ctx context.Context,
 
 	// Ugly hack to clear ramp percentages (not quite correctly) for now
 	for _, d := range deprecatedVersions {
-		d.RampPercentage = nil
+		d.RampPercentage = nil // TODO (Shivam): All deprecatedVersions will have rampPercentage set to nil already
```
carlydf: Why do you think all deprecatedVersions will already have nil ramp?

internal/controller/genstatus.go
```diff
@@ -388,7 +389,7 @@ func (r *TemporalWorkerDeploymentReconciler) generateStatus(ctx context.Context,
 		case desiredVersionID, defaultVersionID:
 			continue
 		}
-		d, _ := versions.getWorkerDeploymentVersion(version)
+		d, _ := versions.getWorkerDeploymentVersion(version) // TODO (Shivam): How do we know these versions have been deleted by Temporal? They could just be draining...
```
carlydf, line 390: good catch! I re-read the code and you're right, all we know about the deprecatedVersions is that they are not current or ramping; they're not necessarily deleted!

If we wanted the list of deprecatedVersions to contain only versions that are not registered with Temporal, then we could make that true, however, in genplan.go:89, when that list is used I didn't treat it as a "deleted versions only list." There, we loop over all the `Status.DeprecatedVersions` and then make scaling decisions based on whether the `VersionStatus` is `Inactive`, `Ramping`, `Current`, `Drained`, or `NotRegistered`. This doesn't really make sense because in genstatus.go we exclude "default" and "target" version from that list, so the `Ramping` and `Current` cases won't actually occur. 

My comment `// TODO(carlydf): Consolidate scale up cases and verify that scale up is the correct action for inactive versions` kinda touches on this inconsistency and was me realizing something was off, but I didn't realize that the issue starts when `deprecatedVersions` is created.

carlydf: The way I'd like to fix this is:
- redefine `deprecatedVersions` to include `Draining`, `Drained`, and `NotRegistered` versions
  - when building the `deprecatedVersions` list, if we come across a version that was not in the server's `VersionSummaries` list, give it the `NotRegistered` status (I think currently those would have an empty status)
- only handle those 3 cases in the genplan loop over that list, because scaling of the target version is handled in the next code block

internal/controller/worker_controller_test.go
```diff
@@ -29,152 +39,413 @@ var (
 			},
 		},
 	}
+
+	testNamespace = "ns"
 )
 
+// Hash is always taken of this for generating the buildID.
```
carlydf: If you check out `func computeBuildID`, note that it's actually hashing the pod spec, which is `TemporalWorkerDeploymentSpec.Template`, not the whole TWD spec

internal/controller/worker_controller_test.go
```diff
@@ -29,152 +39,413 @@ var (
...
+			expectedPlan: plan{
+				TemporalNamespace:    testNamespace,
+				WorkerDeploymentName: testNamespace + deploymentNameSeparator + "baz",
+				DeleteDeployments:    nil,
+				CreateDeployment:     newTestDeploymentWithHashedBuildID(testPodTemplate, newTestWorkerSpec(3), "baz"),
+				ScaleDeployments:     make(map[*v1.ObjectReference]uint32),
+				UpdateVersionConfig:  nil,
+			},
+		},
 		"delete unreachable deployment versions": {
```
carlydf:
```suggestion
		"delete drained deployment versions": {
```

internal/controller/worker_controller_test.go
```diff
@@ -29,152 +39,413 @@ var (
...
 		Template: testPodTemplate,
 		WorkerOptions: temporaliov1alpha1.WorkerOptions{
-			TemporalNamespace: "baz",
+			TemporalNamespace: testNamespace,
 		},
 	}
 }
 
-func newTestDeployment(podSpec v1.PodTemplateSpec, desiredReplicas int32) *appsv1.Deployment {
+// This is the new Deployment object that will be created by the controller based on the hash of the TemporalWorkerDeploymentSpec
```
carlydf: I'm a bit confused on the reason for creating two separate functions, one `WithHashedBuildID` and one `WithBuildID`. I see them both used down in the tests, but am still not super sure both are necessary? But I could be missing something. I think you should be able to also omit the `deploymentName` input and instead generate a deployment name from the TWD name as we do in the non-test situation.

internal/controller/genstatus.go
Shivs11: I like your proposed idea and had something similar in mind since I realized we have separate scaling checks for "default" (the current "current") and the target version. Your proposed idea is nice, and it's something I had in mind too, but did you forget to account for the `VersionStatusInactive` case?

My point being: I think `deprecatedVersions` will have to account for Draining, Drained, NotRegistered as well as Inactive versions.

internal/controller/genstatus.go
```diff
@@ -388,7 +389,7 @@ func (r *TemporalWorkerDeploymentReconciler) generateStatus(ctx context.Context,
...
 		deprecatedVersions = append(deprecatedVersions, d)
```
Shivs11, line 391: Pasting this as a separate question:

What does the controller do in the case where we have a version registered in Temporal but no corresponding k8s deployment object for it? I looked at the code and here's what I found out:

In `genStatus`, we list all deployment versions registered with Temporal and iterate over them to store in the `deploymentVersionsCollection`. Then, we go over the `deployedVersions` list (a list containing all the current k8 deployments in that namespace) and check that if a deployedVersion is present in `deploymentVersionsCollection`, we add it to `deprecatedVersions` so that the worker-controller can take further scaling actions.

I didn't see how, in code, we account for those versions which haven't been created by the controller itself. Maybe this was intended, but I would love to be sure!

internal/controller/genstatus.go
Shivs11: was wrong here - sorry!

internal/controller/worker_controller_test.go
Shivs11: ah, was an outdated comment - if you take a peek at `newTestDeploymentWithHashedBuildID` that's what I am doing as well!

internal/controller/worker_controller_test.go
Shivs11: actually, I think this should be: "delete unregistered deployment version". What its doing is that it's seeing if we delete those deployments which are not even registered with Temporal.

internal/controller/worker_controller_test.go
Shivs11: sure:

The main reason for creating two functions was: In the case of a `targetVersion` not having a corresponding k8s deployment, `genPlan` ends up creating one where the `buildID` is the hash of spec.Template. For cases like these, the test should generate the hash the same way `genPlan` is doing to check correctness. (Check "create new k8s deployment from the pod template")

In all other cases, where we do have a corresponding k8s deployment for a version, we don't need to generate a new deployment for these versions and can rely on the test buildID's we are passing to them for evaluation. Thus, all `newTestDeploymentWithBuildID` does is return a `Deployment` with the buildID we are passing. Sure, this is not how it's done anyways inside `genPlan` because a user can't pass in a buildID (right now), but I believe for testing purposes, we want to check correctness and this does that.

One could argue that you could just consolidate `newTestDeploymentWithBuildID` and `newTestDeploymentWithHashedBuildID` by having an extra arg and deciding whether or not to generate the `buildID` based on the arg, but I had decided against that since I thought it would confuse the reader even more and take their attention away from focussing on the test itself.

I am happy to go either way, but I would first like to understand if I have resolved your confusion and your two cents on this.

internal/controller/genstatus.go
Shivs11: also: 

> when building the deprecatedVersions list, if we come across a version that was not in the server's VersionSummaries list, give it the NotRegistered status (I think currently those would have an empty status)

I don't think so - taking a peek inside `getWorkerDeploymentVersion` (used to build inputs for the `deprecatedVersions` slice), the default value for the status is `VersionStatusNotRegistered`.

https://github.com/DataDog/temporal-worker-controller/blob/a1b8e49c1168c123f22323c5e2b0a33c59135597/internal/controller/genstatus.go#L67

internal/controller/genstatus.go
```diff
@@ -29,6 +29,7 @@ import (
 	temporaliov1alpha1 "github.com/DataDog/temporal-worker-controller/api/v1alpha1"
 )
 
+// TODO (Shivam): Should we be having a map of [versionID] -> Structure?
```
Shivs11, line 32: wdyt about this @carlydf  - It seems that right now, `deploymentVersionCollection` seems to be having multiple maps whose key seems to be the versionID - I was thinking of something like:

```
type deploymentVersionCollection struct {

	versionIDsToDeployments map[string]*appsv1.Deployment
	// map of version IDs to ramp percentages [0,100]
	rampPercentages map[string]float32
	// map of version IDs to task queue stats
	stats map[string]temporaliov1alpha1.QueueStatistics
	// map of version IDs to version status
	versionStatus map[string]temporaliov1alpha1.VersionStatus
	// map of version IDs to drained since timestamps.
	drainedSince map[string]*metav1.Time
	// map of version IDs to test workflow executions
	testWorkflowStatus map[string][]temporaliov1alpha1.WorkflowExecution
	// map of version IDs to task queues
	taskQueues map[string][]string
}
```

to something like:

```
type versionInfo struct {
	// Corresponding k8s deployment
	deployment *appsv1.Deployment
	// Stats for all task queues
	stats temporaliov1alpha1.QueueStatistics
	// Ramp percentage [0,100]
	rampPercentage float32
	// Version status
	status temporaliov1alpha1.VersionStatus
	// Drained since timestamp
	drainedSince *metav1.Time
	// Test workflow executions
	testWorkflows []temporaliov1alpha1.WorkflowExecution
	// Task queues
	taskQueues []string
}

func deploymentVersionCollection struct {
	map[string]versionInfo
}
```

this might just be a `nit` but something worth considering.

internal/controller/genstatus.go
carlydf: Ok great, so we're good on setting the `NotRegistered` status, because that's the default empty status, thanks for finding that.

As for `Inactive` versions, we only care to scale up an `Inactive` version if it is the target version. If a version is deployed, but it is not the target, or current, or draining/drained within the scaledown delay, then we have no reason to scale it up. So I think it is correct to exclude `Inactive` versions from the `deprecatedVersions` list.

As long as a version goes from `NotRegistered` -> `Inactive` (Target) -> `Ramping`, then the server will track Drainage status for it, which means it will show up correctly in the `deprecatedVersions` list once it starts draining.

carlydf: There is one version lifecycle scenario that I think is an important "sad path" to consider though, which is this:

1. `TWD version xyz` is deployed to k8s --> Status: `NotRegistered`
2. Controller creates Deployment for Target`xyz`, worker polls and server starts tracking version `xyz` --> Status: `Inactive`
3. Controller runs canary test workflow on version `xyz`, it fails --> Status: `Inactive` (no change)
4. User deploys new `TWD version abc` to k8s as new target, controller scales it up and runs canary on it, which succeeds. Version `abc` status is now `Ramping` but version `xyz` remains `Inactive` but still has a Deployment with a running worker! **--> Question: When can the controller scale down the inactive failed canary deployment? From the server perspective, it will never be marked as draining/drained because it was never ramping or current. So there's no "drained since" time.**

Possible answer: If a canary fails and we're not ever going to retry that version, the controller could call `temporal worker version delete -v xyz --force` on it, which would cause the status to become `Unregistered`, at which point the controller would actually scale down the worker. However, if we delete the version by API, and then the `xyz` worker polls again, the version will just immediately become `Inactive` again. So, we do need to handle inactive versions because we need to solve the case where a version is deployed and never promoted to `Ramping`.

internal/controller/genstatus.go
carlydf: I think `deploymentVersionCollection` as written is more space-efficient, because there are multiple scenarios in which we don't have a `rampPercentage` for a version, or don't have `drainedSince` or `testWorkflows` or even a `deployment`. The version starts out with just a key and a status. If we only add the version id to the `rampPercentages` map when it has a ramp percentage, then the memory is allocated only when needed. Similarly, later on when a version is current / draining / drained instead of ramping, the `rampPercentage` does not need to be populated.

However, if the moment a version is created, we make a whole `versionInfo` struct, then the memory space for all the empty fields is allocated before it's needed (and it might never be needed, if a version fails the canary test and does not ramp or drain).

^^ that's just me guessing at the design choice though, so @jlegrone should confirm!

internal/controller/genstatus.go
carlydf: I think right now the controller does not support "adopting" k8s deployments that were deployed manually. Not sure if that's on the roadmap / if we want it to be on the roadmap. Certainly we need to figure out an onboarding story of how we recommend adopting the controller especially in brownfield situations where folks have multiple old workers they need to keep around for some period of time.

carlydf: cc @jlegrone to confirm here

internal/controller/genstatus.go
```diff
@@ -138,13 +139,11 @@ func (c *deploymentVersionCollection) addAssignmentRule(rule *taskqueue.BuildIdA
 
 func (c *deploymentVersionCollection) addVersionStatus(version string, status temporaliov1alpha1.VersionStatus) {
 	c.versionStatus[version] = status
-	return
```
carlydf, line 141: did you intend to delete these? IMO it's better to not make changes that are not relevant to your PR, so if you could revert these two deleted return statements that would be great!

internal/controller/genstatus.go
Shivs11: I agree with the first comment you (describing a happy path) of a version lifecycle.

Yeah, I think this is a really good sad path scenario and I also see this as a common path of execution for most developers out there. I say this because I imagine code to be shipped out in a manner where new versions have broken code and canaries catch these errors. Having said that, I like your proposed idea but have a question:

>  However, if we delete the version by API, and then the xyz worker polls again, the version will just immediately become Inactive again.

How does `xyz` start polling again? By that I mean the very specific worker `xyz`? IIUC, right now, versions are going to be random as well as unique and created as the hash of the spec, right? If that is the case, even if someone were to deploy the same changes in code, wouldn't the version specifically change?

Shivs11: Idea #2 - I realize that for these "stale" versions to be scaled down and deleted by the controller, the easiest way would be that the version be deleted from the Temporal Server. What if a deployment scavenger were to be present and were to remove "inactive" versions after duration `x`? We could choose `x` to be a small value, for now.  Sure, this idea would also require us implementing this scavenger in the server but that becomes a server's dependency and not the controller. The controller just does what it's been doing so far and in this way, a "stale" version becomes "not registered" eventually and the controller gets rid of the deployment.

I don't necessarily think this idea is better than yours, since the deployment would be active for time `x`, but just putting it out there as a potential option.

internal/controller/genstatus.go
Shivs11: sure, the current approach does seem more memory efficient but I wonder if the worker-controller would have size/memory constraints over which things would break (I might be wrong here) - I think the reason why I pasted the above suggestion is because I like the idea of having one structure take care of everything that there is to do with a specific versionID. Right now, we have to be wary about adding the versionID as a key in multiple maps and while we are doing it right, it's something I feel we can avoid.

We can let Jacob pitch in with his thoughts, and it's not something I think we *must* do/change!

internal/controller/genstatus.go
Shivs11: will be reverting back to this :) 
thanks!

internal/controller/genstatus.go
jlegrone: This is really good discussion, but I think starting to go beyond the scope of this change. Would one of you mind filing an issue summarizing the "skipped version" scenario and current/desired behavior of the controller?

internal/controller/genstatus.go
jlegrone: Good question -- I definitely agree that the controller should behave gracefully in the event that there is a worker version registered that is not associated with a k8s deployment that it manages. Right now there is no special handling for this case.

Like Carly said this could come up during brownfield migrations to the controller. Migrations from one k8s namespace or k8s cluster to another, or even just renaming a TWD resource might also induce a similar state.

By "behave gracefully" I mean that the controller:
- Should not panic or exhibit any other undefined behaviors
- Should continue being able to roll out _new_ versions
- Should continue scaling and deleting existing versions that it manages

Are there any other requirements that y'all think we should capture?

internal/controller/genstatus.go
jlegrone: I think testability and developer experience are more important here than memory efficiency. We can always refactor later if memory usage becomes an issue.

I think I wrote the collection this way originally because each attribute of build ID was updated from a different section of code, usually looping over a list of build IDs, and I didn't want to have to perform a bunch of nil checks for each write. But I'm sure this can be done in a much nicer way so please feel free to refactor!

jlegrone (approved):

internal/controller/worker_controller_test.go
```diff
@@ -29,152 +39,413 @@ var (
 			},
 		},
 	}
+
+	testNamespace = "ns"
```
jlegrone: Nit: it's helpful to disambiguate between temporal and k8s namespaces throughout this codebase 😄 

(this is also why it's `WorkerOptions.TemporalNamespace` instead of `WorkerOptions.Namespace`)

```suggestion
	testTemporalNamespace = "ns"
```

internal/controller/genstatus.go
Shivs11: sure thing - for the purpose of moving quicker, I shall open a separate PR with the data-structure refactor!

internal/controller/worker_controller_test.go
Shivs11: totally agree! thanks for bringing this to my attention!

internal/controller/genstatus.go
Shivs11: done! https://github.com/DataDog/temporal-worker-controller/issues/33

internal/controller/genstatus.go
jlegrone: Filed a new issue to track: https://github.com/DataDog/temporal-worker-controller/issues/34

internal/controller/worker_controller_test.go
```diff
@@ -29,152 +39,391 @@ var (
...
+			TemporalNamespace: testTemporalNamespace,
 		},
 	}
 }
 
-func newTestDeployment(podSpec v1.PodTemplateSpec, desiredReplicas int32) *appsv1.Deployment {
+// newTestDeployment creates a new Deployment object for testing purposes.
+// If buildID is empty, it will be computed from the desiredState.Template.
+func newTestDeployment(podSpec v1.PodTemplateSpec, desiredState *temporaliov1alpha1.TemporalWorkerDeploymentSpec, deploymentName string, buildID string) *appsv1.Deployment {
+	// If buildID is not provided, compute it from the template
```
carlydf, line 60: this looks like a good solution to me! and avoids having to maintain repeated code 👍

carlydf (approved):

Shivs11:
@jlegrone - seems like you are the only one who can merge PR's at this moment in time! Signed all my commits too!

PR #32, opened 2025-04-02
Helm chart

carlydf:
This is based on the `manager` skaffold profile.

There are a few TODOs where I was not sure how kustomize was substituting certain variables.

I've tested installing it in minikube with the default values and rolling out a few versions. I have not tested:
- certmanager
- prometheus
- webhook
- any of the other options that are disabled in values.yaml


helm/temporal-worker-controller/values.yaml
```diff
@@ -0,0 +1,41 @@
+namespace: temporal-system
+
+# Set to true for development environment with additional logging
+devEnv: false
+
+image:
+  repository: temporal-worker-controller
```
Shivs11, line 2: think this should be updated to the internal ECR repo on which we intend on hosting our controller images! But I guess that doesn't apply when running things locally, nevertheless, thought I should point this out so that we don't forget.

carlydf: yes this is just the default! in Temporal internally we would change this value

helm/temporal-worker-controller/values.yaml
```diff
@@ -0,0 +1,41 @@
...
+leaderElection:
+  # Leader election should always be enabled in production.
+  enabled: true
+
+webhook:
+  enabled: true
+  port: 9443
+  certDir: /tmp/k8s-webhook-server/serving-certs
+
+terminationGracePeriodSeconds: 10
```
jlegrone: Do you mind generating a schema file from these defaults as well? https://helm.sh/docs/topics/charts/#schema-files

That will allow us to add extra validation like enums, min/max, etc. (I don't see any use cases for that kind of additional validation right now, but it's nice to make it easy to add in the future).

helm/temporal-worker-controller/values.yaml
```diff
@@ -0,0 +1,41 @@
...
+    cpu: 10m
+    memory: 64Mi
+
+metrics:
+  enabled: true
+  port: 8080
+
+leaderElection:
+  # Leader election should always be enabled in production.
+  enabled: true
```
jlegrone: Are there scenarios where we would want to turn leader election _off_? I don't think I've run into any issues with this during development so far, so I'm wondering if the configuration option can be elided for now.

helm/temporal-worker-controller/values.yaml
```diff
@@ -0,0 +1,41 @@
...
+# Set to true for development environment with additional logging
+devEnv: false
+
+image:
+  repository: temporal-worker-controller
+  tag: latest
+  pullPolicy: IfNotPresent
+
+serviceAccount:
+  name: controller-manager
```
jlegrone: The service account `name` value should be left empty by default so that the Helm release name can be used. Defaulting to the release name makes it less likely for two installations of the chart in the same namespace to attempt modifying the same service account's annotations and rolebindings.

A `create` option and separate `rbac` section as recommended here would also make it easier to use existing service accounts managed outside of the chart: https://helm.sh/docs/chart_best_practices/rbac/

helm/temporal-worker-controller/values.yaml
```diff
@@ -0,0 +1,41 @@
+namespace: temporal-system
```
jlegrone:
```suggestion
```

The k8s namespace of the templated resources should match the namespace provided to the Helm install command, so I'd suggest replacing any references to `$.namespace` with `$.Release.Namespace` as per https://helm.sh/docs/topics/charts/#predefined-values

We can still document a recommended namespace in the install instructions, ie. `helm install ./helm/temporal-worker-controller --namespace temporal-system`.

helm/temporal-worker-controller/values.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+# Set to true for development environment with additional logging
+devEnv: false
```
jlegrone: I would prefer to have explicit toggles for the behaviors that we want to enable in dev environments (like log level) so that they can all be independently adjusted. Occasionally you might want more verbose logs in a prod environment, but if the only way to do that is to set `devEnv: true` then there could be other side effects that are undesirable.

A common pattern is to provide additional "profiles" in the form of values files that can optionally be passed in when installing the chart. We could provide a "development" profile that sets low resource limits, verbose log level, etc.

helm/temporal-worker-controller/templates/deployment.yaml
```diff
@@ -0,0 +1,85 @@
...
+  name: temporal-worker-controller
+  namespace: {{ .Values.namespace }}
+  labels:
+    control-plane: controller-manager
+    app.kubernetes.io/name: deployment
+    app.kubernetes.io/instance: temporal-worker-controller
+    app.kubernetes.io/component: manager
+    app.kubernetes.io/created-by: worker-controller
+    app.kubernetes.io/part-of: worker-controller
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
```
jlegrone: I think there are a few adjustments to make based on recommendations from https://helm.sh/docs/chart_best_practices/labels/#standard-labels

`app.kubernetes.io/name` should be the chart name
`app.kubernetes.io/instance` should be `$.Release.Name` (needs to be unique to every instance created within the k8s namespace)
`app.kubernetes.io/component` lgtm
`app.kubernetes.io/created-by` delete (it's [deprecated](app.kubernetes.io/created-by))
`app.kubernetes.io/part-of` delete (it's optional, and the typical use case is to indicate that something like a generic cache or database is part of a higher level application like "wordpress")
`app.kubernetes.io/managed-by` lgtm

jlegrone: Actually every resource should just get its labels from the helper for consistency.

```suggestion
{{ include "temporal-worker-controller.labels" $ | indent 4 }}
```

helm/temporal-worker-controller/templates/deployment.yaml
```diff
@@ -0,0 +1,85 @@
...
+spec:
+  selector:
+    matchLabels:
+      control-plane: controller-manager
```
jlegrone:
```suggestion
      {{- include "temporal-worker-controller.selectorLabels" $ | nindent 6 }}
```

helm/temporal-worker-controller/templates/deployment.yaml
```diff
@@ -0,0 +1,85 @@
...
+  replicas: 1
+  template:
+    metadata:
+      annotations:
+        kubectl.kubernetes.io/default-container: manager
+        {{- with .Values.podAnnotations }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
+      labels:
+        control-plane: controller-manager
```
jlegrone:
```suggestion
        {{- include "temporal-worker-controller.selectorLabels" $ | nindent 8 }}
```

helm/temporal-worker-controller/templates/rbac.yaml
```diff
@@ -0,0 +1,278 @@
...
+  - get
+  - list
+  - patch
+  - update
+  - watch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ .Release.Name }}-temporal-role
```
jlegrone: [Cluster roles are global](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#role-and-clusterrole) / not scoped to the helm release's k8s namespace, so we should include the release namespace in the name in order to prevent conflicts if there are multiple installations of the controller in the same k8s cluster.

```suggestion
  name: {{ .Release.Name }}-{{ .Release.Namespace }}-manager-role
```

helm/temporal-worker-controller/templates/_helpers.tpl
```diff
@@ -0,0 +1,50 @@
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "temporal-worker-controller.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
+{{- end }}
```
jlegrone: Nit: I prefer just using `$.Chart.Name` in templates. There's no real need for overriding the chart name and that also avoids the complexity of truncating custom name overrides to fit k8s name limits.
```suggestion
```

helm/temporal-worker-controller/templates/_helpers.tpl
```diff
@@ -0,0 +1,50 @@
...
+{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" }}
+{{- else }}
+{{- $name := default .Chart.Name .Values.nameOverride }}
+{{- if contains $name .Release.Name }}
+{{- .Release.Name | trunc 63 | trimSuffix "-" }}
+{{- else }}
+{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" }}
+{{- end }}
+{{- end }}
+{{- end }}
```
jlegrone: This helper is unused (I think)

```suggestion
```

helm/temporal-worker-controller/values.yaml
carlydf: yeah honestly I haven't either. I'll remove it

helm/temporal-worker-controller/values.yaml
carlydf: thanks! done

helm/temporal-worker-controller/values.yaml
carlydf: do you mind if I just remove that optional `--zap-devel` flag for now? I'm thinking we can always add it later?

helm/temporal-worker-controller/templates/_helpers.tpl
carlydf: Totally! I actually removed the entire helpers.tpl file when I was pruning the complexity of this chart before I saw your comments. I think it could make sense to add back the labels thing though.

helm/temporal-worker-controller/templates/_helpers.tpl
carlydf: It is! I just deleted it. Sorry I was late preparing this for a good review / you saw it a bit half-baked still

helm/temporal-worker-controller/templates/deployment.yaml
carlydf: Thanks so much for typing this out! I just implemented that guidance / corrected the silly AI labels before I saw your comments. The best practice links you're posting are v helpful too

helm/temporal-worker-controller/values.yaml
jlegrone: Yes, no need to expose the configuration option at the moment.

jlegrone (approved):

helm/temporal-worker-controller/templates/webhook.yaml
```diff
@@ -0,0 +1,83 @@
+{{- if .Values.webhook.enabled -}}
+apiVersion: admissionregistration.k8s.io/v1
```
jlegrone:
```suggestion
{{ fail "webhooks aren't supported yet" }}
```

PR #35, opened 2025-04-23
controller refreshes client when the client's certs get expired!

Shivs11:
WISOTT

### DIFF
--- a/internal/controller/clientpool/clientpool.go
+++ b/internal/controller/clientpool/clientpool.go
@@ -164,18 +164,10 @@ func (cp *ClientPool) UpsertClient(ctx context.Context, opts NewClientOptions) (
 		HostPort:  opts.Spec.HostPort,
 		Namespace: opts.TemporalNamespace,
 	}
-	if opts.Spec.MutualTLSSecret != "" {
-		cp.clients[key] = ClientInfo{
-			client:     c,
-			tls:        clientOpts.ConnectionOptions.TLS,
-			expiryTime: expiryTime,
-		}
-	} else {
-		cp.clients[key] = ClientInfo{
-			client:     c,
-			tls:        nil,
-			expiryTime: expiryTime,
-		}
+	cp.clients[key] = ClientInfo{
+		client:     c,
+		tls:        clientOpts.ConnectionOptions.TLS,
+		expiryTime: expiryTime,
 	}
 
 	return c, nil
--- a/internal/controller/clientpool/clientpool.go
+++ b/internal/controller/clientpool/clientpool.go
@@ -70,7 +70,7 @@ func (cp *ClientPool) GetSDKClient(key ClientPoolKey) (sdkclient.Client, bool) {
 				return nil, false
 			}
 			if expired {
-				cp.logger.Info("Certificate expired", "expiry", cert.Leaf.NotAfter)
+				cp.logger.Info("Certificate is expired or is going to expire soon", "expiry", cert.Leaf.NotAfter)
 				return nil, false
 			}
 		}
@@ -131,7 +131,7 @@ func (cp *ClientPool) UpsertClient(ctx context.Context, opts NewClientOptions) (
 			return nil, fmt.Errorf("failed to check certificate expiration: %v", err)
 		}
 		if expired {
-			return nil, fmt.Errorf("certificate is expired")
+			return nil, fmt.Errorf("certificate is expired or is going to expire soon")
 		}
 
 		cert, err := tls.X509KeyPair(secret.Data["tls.crt"], secret.Data["tls.key"])
@@ -193,10 +193,10 @@ func (cp *ClientPool) isCertificateExpiredSoon(certBytes []byte, howSoon time.Du
 		return true, fmt.Errorf("failed to parse certificate: %v", err)
 	}
 
-	expiryTime := cert.NotAfter.Add(-howSoon)
+	almostExpiredTime := cert.NotAfter.Add(-howSoon)
 
 	// Check if certificate is expired or will expire within the safety buffer
-	if time.Now().After(expiryTime) {
+	if time.Now().After(almostExpiredTime) {
 		return true, nil
 	}
 
--- a/internal/controller/clientpool/clientpool.go
+++ b/internal/controller/clientpool/clientpool.go
@@ -64,7 +64,7 @@ func (cp *ClientPool) GetSDKClient(key ClientPoolKey) (sdkclient.Client, bool) {
 			if len(cert.Certificate) == 0 {
 				continue
 			}
-			expired, err := cp.isCertificateExpiredSoon(cert.Certificate[0], 30*time.Minute)
+			expired, err := cp.isCertificateExpiredSoon(cert.Certificate[0], 5*time.Minute)
 			if err != nil {
 				cp.logger.Error("Error checking certificate expiration", "error", err)
 				return nil, false
@@ -126,7 +126,7 @@ func (cp *ClientPool) UpsertClient(ctx context.Context, opts NewClientOptions) (
 		}
 
 		// Check if certificate is expired before creating the client
-		expired, err := cp.isCertificateExpiredSoon(secret.Data["tls.crt"], 30*time.Minute)
+		expired, err := cp.isCertificateExpiredSoon(secret.Data["tls.crt"], 5*time.Minute)
 		if err != nil {
 			return nil, fmt.Errorf("failed to check certificate expiration: %v", err)
 		}
--- a/internal/controller/clientpool/clientpool.go
+++ b/internal/controller/clientpool/clientpool.go
@@ -30,8 +30,9 @@ type ClientPoolKey struct {
 }
 
 type ClientInfo struct {
-	Client sdkclient.Client
-	TLS    *tls.Config // Storing the TLS config associated with the client to check certificate expiration. If the certificate is expired, a new client will be created.
+	Client  sdkclient.Client
+	TLS     *tls.Config // Storing the TLS config associated with the client to check certificate expiration. If the certificate is expired, a new client will be created.
+	PEMCert []byte      // Store the original PEM-encoded certificate for expiration checks
 }
 
 type ClientPool struct {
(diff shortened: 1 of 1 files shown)


internal/controller/clientpool/clientpool.go
```diff
@@ -131,8 +173,31 @@ func (cp *ClientPool) Close() {
...
 
 	for _, c := range cp.clients {
-		c.Close()
+		c.Client.Close()
+	}
+
+	cp.clients = make(map[ClientPoolKey]ClientInfo)
+}
+
+func (cp *ClientPool) isCertificateExpired(certBytes []byte) (bool, error) {
```
carlydf: I think it would be better to refresh the certificate _before_ it's expired not _when_ it's expired.

For example, the certificate we are working with in our cicd dogfooding expires after 24h, but I think it is renewed after 16h. This means that between 16h-24h, the new cert is available (I think).

Renewing the client _before_ it expires will avoid downtime if the controller needs to make calls right at the moment of expiry. An hour or 30 minutes before expiry is probably a safe time to refresh the client IMO.

If the user renews their cert only 5 minutes before expiry, then the controller would recreate the client unnecessarily a few times, but I think that's alright.

internal/controller/clientpool/clientpool.go
```diff
@@ -96,10 +124,21 @@ func (cp *ClientPool) UpsertClient(ctx context.Context, opts NewClientOptions) (
 			err := fmt.Errorf("secret %s must be of type kubernetes.io/tls", secret.Name)
 			return nil, err
 		}
+
+		// Check if certificate is expired before creating the client
```
carlydf, line 128: Here I like that we only error if the cert is actually expired.

internal/controller/clientpool/clientpool.go
```diff
@@ -44,11 +53,30 @@ func (cp *ClientPool) GetSDKClient(key ClientPoolKey) (sdkclient.Client, bool) {
...
 	}
-	return nil, false
+
+	// Check if any certificate is expired
+	if info.TLS != nil {
+		for _, cert := range info.TLS.Certificates {
+			if len(cert.Certificate) == 0 {
+				continue
+			}
+			expired, err := cp.isCertificateExpired(cert.Certificate[0])
```
carlydf: Here I think we should try to renew before the cert expires

internal/controller/clientpool/clientpool.go
```diff
@@ -131,8 +172,33 @@ func (cp *ClientPool) Close() {
...
+	if block == nil {
+		return true, fmt.Errorf("failed to decode PEM block")
+	}
+
+	cert, err := x509.ParseCertificate(block.Bytes)
+	if err != nil {
+		return true, fmt.Errorf("failed to parse certificate: %v", err)
+	}
+
+	expiryTime := cert.NotAfter.Add(-howSoon)
```
Shivs11: renewing before the cert expires!!!

internal/controller/clientpool/clientpool.go
```diff
@@ -44,11 +53,30 @@ func (cp *ClientPool) GetSDKClient(key ClientPoolKey) (sdkclient.Client, bool) {
...
+			if len(cert.Certificate) == 0 {
+				continue
+			}
+			expired, err := cp.isCertificateExpiredSoon(cert.Certificate[0], 30*time.Minute)
+			if err != nil {
+				cp.logger.Error("Error checking certificate expiration", "error", err)
+				return nil, false
+			}
+			if expired {
+				cp.logger.Info("Certificate expired", "expiry", cert.Leaf.NotAfter)
```
carlydf:
```suggestion
			almostExpired, err := cp.isCertificateExpiredSoon(cert.Certificate[0], 30*time.Minute)
			if err != nil {
				cp.logger.Error("Error checking certificate expiration", "error", err)
				return nil, false
			}
			if almostExpired {
				cp.logger.Info("Certificate expiring soon", "expiry", cert.Leaf.NotAfter)
```

internal/controller/clientpool/clientpool.go
```diff
@@ -96,6 +124,16 @@ func (cp *ClientPool) UpsertClient(ctx context.Context, opts NewClientOptions) (
 			err := fmt.Errorf("secret %s must be of type kubernetes.io/tls", secret.Name)
 			return nil, err
 		}
+
+		// Check if certificate is expired before creating the client
+		expired, err := cp.isCertificateExpiredSoon(secret.Data["tls.crt"], 30*time.Minute)
```
carlydf:
```suggestion
		expired, err := cp.isCertificateExpiredSoon(secret.Data["tls.crt"], 0*time.Minute)
```

internal/controller/clientpool/clientpool.go
```diff
@@ -131,8 +172,33 @@ func (cp *ClientPool) Close() {
...
+
+	cert, err := x509.ParseCertificate(block.Bytes)
+	if err != nil {
+		return true, fmt.Errorf("failed to parse certificate: %v", err)
+	}
+
+	expiryTime := cert.NotAfter.Add(-howSoon)
+
+	// Check if certificate is expired or will expire within the safety buffer
+	if time.Now().After(expiryTime) {
```
carlydf: nit: I feel like the way you're setting up the `expiryTime` variable is a bit hard to read, because the real expiry time is `cert.NotAfter`, so it's confusing to subtract from that time and still call it `expiryTime`. Does that make sense?

Suggestion:

```suggestion
	almostExpiredTime := cert.NotAfter.Add(-howSoon)

	// Check if certificate is expired or will expire within the safety buffer
	if time.Now().After(almostExpiredTime) {
```

internal/controller/clientpool/clientpool.go
carlydf: wait actually this isn't quite accurate, cuz the cert could also be expired. let me edit

internal/controller/clientpool/clientpool.go
```diff
@@ -44,11 +53,30 @@ func (cp *ClientPool) GetSDKClient(key ClientPoolKey) (sdkclient.Client, bool) {
...
+			if len(cert.Certificate) == 0 {
+				continue
+			}
+			expired, err := cp.isCertificateExpiredSoon(cert.Certificate[0], 5*time.Minute)
+			if err != nil {
+				cp.logger.Error("Error checking certificate expiration", "error", err)
+				return nil, false
+			}
+			if expired {
+				cp.logger.Info("Certificate expired", "expiry", cert.Leaf.NotAfter)
```
carlydf:
```suggestion
				cp.logger.Info("Certificate expired or expiring soon", "expiry", cert.Leaf.NotAfter)
```

internal/controller/clientpool/clientpool.go
```diff
@@ -96,6 +124,16 @@ func (cp *ClientPool) UpsertClient(ctx context.Context, opts NewClientOptions) (
 			err := fmt.Errorf("secret %s must be of type kubernetes.io/tls", secret.Name)
 			return nil, err
 		}
+
+		// Check if certificate is expired before creating the client
+		expired, err := cp.isCertificateExpiredSoon(secret.Data["tls.crt"], 5*time.Minute)
```
carlydf, line 129:
```suggestion
		expired, err := cp.isCertificateExpiredSoon(secret.Data["tls.crt"], 0*time.Minute)
```

carlydf (approved):

internal/controller/clientpool/clientpool.go
```diff
@@ -44,11 +54,25 @@ func (cp *ClientPool) GetSDKClient(key ClientPoolKey) (sdkclient.Client, bool) {
...
+
+	// Check if any certificate is expired
+	if info.PEMCert != nil {
+		expired, err := cp.isCertificateExpiredSoon(info.PEMCert, 5*time.Minute)
+		if err != nil {
+			cp.logger.Error("Error checking certificate expiration", "error", err)
+			return nil, false
+		}
+		if expired {
+			cp.logger.Info("Certificate is expired or is going to expire soon")
```
jlegrone: Should this be warn level instead?

internal/controller/clientpool/clientpool.go
```diff
@@ -131,8 +183,33 @@ func (cp *ClientPool) Close() {
...
 
 	for _, c := range cp.clients {
-		c.Close()
+		c.Client.Close()
+	}
+
+	cp.clients = make(map[ClientPoolKey]ClientInfo)
+}
+
+func (cp *ClientPool) isCertificateExpiredSoon(certBytes []byte, howSoon time.Duration) (bool, error) {
```
jlegrone: Can this be a top level function instead of a method? It looks like there are no references to `cp`.

internal/controller/clientpool/clientpool.go
```diff
@@ -25,17 +29,23 @@ type ClientPoolKey struct {
 	Namespace string
 }
 
+type ClientInfo struct {
+	Client  sdkclient.Client
+	TLS     *tls.Config // Storing the TLS config associated with the client to check certificate expiration. If the certificate is expired, a new client will be created.
+	PEMCert []byte      // Store the original PEM-encoded certificate for expiration checks
```
jlegrone: Would it make sense to only store the expiration time instead of the whole cert?

jlegrone: Nit: even though this is already an internal package, it's good hygiene to avoid public symbols unless they need to be referenced by other packages. That way if a public package aliases this type, the fields do not accidentally become part of the public interface.

```suggestion
	client  sdkclient.Client
	tls     *tls.Config // Storing the TLS config associated with the client to check certificate expiration. If the certificate is expired, a new client will be created.
	pemCert []byte      // Store the original PEM-encoded certificate for expiration checks
```

jlegrone: (same thing for `ClientPoolKey` fields, though I didn't notice those before)

internal/controller/clientpool/clientpool.go
```diff
@@ -44,11 +54,25 @@ func (cp *ClientPool) GetSDKClient(key ClientPoolKey) (sdkclient.Client, bool) {
...
-		return c, true
+	info, ok := cp.clients[key]
+	if !ok {
+		return nil, false
 	}
-	return nil, false
+
+	// Check if any certificate is expired
+	if info.PEMCert != nil {
+		expired, err := cp.isCertificateExpiredSoon(info.PEMCert, 5*time.Minute)
```
jlegrone: Can we avoid re-parsing the cert each time a client is fetched? This happens once for each reconciliation loop for each worker, so it could get expensive.

internal/controller/clientpool/clientpool.go
```diff
@@ -44,11 +54,25 @@ func (cp *ClientPool) GetSDKClient(key ClientPoolKey) (sdkclient.Client, bool) {
...
+		if err != nil {
+			cp.logger.Error("Error checking certificate expiration", "error", err)
+			return nil, false
+		}
+		if expired {
+			cp.logger.Info("Certificate is expired or is going to expire soon")
+			return nil, false
```
jlegrone: I might have missed this, but how do we trigger reloading the cert (assuming the secret has already been updated)? Or is the goal only to improve diagnostics through logging?

internal/controller/clientpool/clientpool.go
```diff
@@ -121,7 +164,19 @@ func (cp *ClientPool) UpsertClient(ctx context.Context, opts NewClientOptions) (
 		HostPort:  opts.Spec.HostPort,
 		Namespace: opts.TemporalNamespace,
 	}
-	cp.clients[key] = c
+	if opts.Spec.MutualTLSSecret != "" {
```
jlegrone: Should this be checking `clientOpts.ConnectionOptions.TLS` instead? Maybe in the future this might get set without `MutualTLSSecret` also being configured?

jlegrone (approved):

PR #38, opened 2025-04-25
Reset ramping version on rollback

carlydf:
Closes https://github.com/DataDog/temporal-worker-controller/issues/37

### DIFF
--- a/internal/controller/execplan.go
+++ b/internal/controller/execplan.go
@@ -109,19 +109,24 @@ func (r *TemporalWorkerDeploymentReconciler) executePlan(ctx context.Context, l 
 			}); err != nil { // would be cool to do this atomically with the update
 				return fmt.Errorf("unable to update metadata after setting current deployment: %w", err)
 			}
-		} else if ramp := vcfg.rampPercentage; ramp > 0 { // TODO(carlydf): Support setting any ramp in [0,100]
+		} else if ramp := vcfg.rampPercentage; ramp > 0 || vcfg.unsetRamp { // TODO(carlydf): Support setting any ramp in [0,100]
 			err := awaitVersionRegistration(ctx, l, deploymentHandler, p.TemporalNamespace, vcfg.versionID)
 			if err != nil {
 				return fmt.Errorf("error waiting for version to register, did your pollers start successfully?: %w", err)
 			}
 
-			l.Info("applying ramp", "version", p.UpdateVersionConfig.versionID, "percentage", p.UpdateVersionConfig.rampPercentage)
+			desiredRampingVersion := vcfg.versionID
+			if vcfg.unsetRamp {
+				desiredRampingVersion = ""
+			}
+
+			l.Info("applying ramp", "version", desiredRampingVersion, "percentage", p.UpdateVersionConfig.rampPercentage)
 			resp, err := deploymentHandler.Describe(ctx, sdkclient.WorkerDeploymentDescribeOptions{})
 			if err != nil {
 				return fmt.Errorf("unable to describe worker deployment: %w", err)
 			}
 			if _, err := deploymentHandler.SetRampingVersion(ctx, sdkclient.WorkerDeploymentSetRampingVersionOptions{
-				Version:       vcfg.versionID,
+				Version:       desiredRampingVersion,
 				Percentage:    vcfg.rampPercentage,
 				ConflictToken: resp.ConflictToken,
 				Identity:      "temporal-worker-controller", // TODO(jlegrone): Set this to a unique identity, should match metadata.
--- a/internal/controller/genplan.go
+++ b/internal/controller/genplan.go
@@ -50,6 +50,9 @@ type versionConfig struct {
 	setDefault bool
 	// Acceptable values [0,100]
 	rampPercentage float32
+
+	// Unset ramp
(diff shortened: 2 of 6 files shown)


internal/controller/genplan.go
```diff
@@ -220,15 +223,20 @@ func getVersionConfigDiff(l logr.Logger, strategy temporaliov1alpha1.RolloutStra
...
 
-	// Don't make updates if desired default is already the default
+	// If desired default is already the default...
 	if vcfg.versionID == status.DefaultVersion.VersionID {
-		return nil
+		// If ramping version is already nil, do nothing
+		if status.RampingVersion == nil {
+			return nil
+		}
+		vcfg.unsetRamp = true
```
Shivs11: nice!

internal/controller/worker_controller_test.go
```diff
@@ -20,299 +22,252 @@ import (
...
-	}
-}
-
-// newTestDeployment creates a new Deployment object for testing purposes.
-// If buildID is empty, it will be computed from the desiredState.Template.
-func newTestDeployment(podSpec v1.PodTemplateSpec, desiredState *temporaliov1alpha1.TemporalWorkerDeploymentSpec, deploymentName string, buildID string) *appsv1.Deployment {
-	// If buildID is not provided, compute it from the template
-	if buildID == "" {
-		buildID = utils.ComputeHash(&desiredState.Template, nil)
+func TestGeneratePlan(t *testing.T) {
```
carlydf, line 31: The easiest way to review this change might be to just open the new file and read it as a whole first to understand the format, and then look at the diff.

The TLDR of the benefits here is that in this new structure:
- more of the real genstatus / genplan / utils code is reused, making this more representative
- fewer things need to be hardcoded
- there are ways to add different rollout and sunset policies for different test cases
- all of it is intended to be more readable and reusable than before!

internal/controller/worker_controller_test.go
```diff
@@ -20,299 +22,252 @@ import (
...
+		observedRampingVersion     *testWorkerDeploymentVersion
+		observedDeprecatedVersions []*testWorkerDeploymentVersion
+		expectedPlan               plan
 	}
-
-	labels := map[string]string{
-		"temporal.io/build-id": buildID,
+	testCases := make(map[string]testCase)
+
+	testCases["no action needed"] = testCase{
```
jlegrone, line 42: Nit: it's probably not very likely, but defining test cases this way makes it possible to accidentally overwrite existing cases later in the function.

Defining test cases inline with a map literal avoids this issue because the compiler will throw a duplicate key error.

internal/controller/worker_controller_test.go
jlegrone: To make the tests for GeneratePlan more discoverable, would you mind adding these to a new `genplan_test.go` file? The other test helpers that aren't specific to testing plan functionality should stay in this file though; that may also make them easier to find and reuse.

internal/controller/genplan.go
```diff
@@ -50,6 +50,9 @@ type versionConfig struct {
 	setDefault bool
 	// Acceptable values [0,100]
 	rampPercentage float32
+
+	// Unset ramp
+	unsetRamp bool
```
jlegrone, line 55: I was confused at first about whether this was referring to the current `versionID` or not; my understanding is that when `unsetRamp` is enabled, ramping should be disabled regardless of which version the ramp is associated with. Is that correct? It might help to add more explanation in the doc comment.

internal/controller/execplan.go
```diff
@@ -109,19 +109,24 @@ func (r *TemporalWorkerDeploymentReconciler) executePlan(ctx context.Context, l
...
 				return fmt.Errorf("error waiting for version to register, did your pollers start successfully?: %w", err)
 			}
 
-			l.Info("applying ramp", "version", p.UpdateVersionConfig.versionID, "percentage", p.UpdateVersionConfig.rampPercentage)
+			desiredRampingVersion := vcfg.versionID
+			if vcfg.unsetRamp {
+				desiredRampingVersion = ""
+			}
+
+			l.Info("applying ramp", "version", desiredRampingVersion, "percentage", p.UpdateVersionConfig.rampPercentage)
```
jlegrone, line 123: I think this log could potentially be confusing; it won't be possible to disambiguate between a ramp being "removed" vs. modified as expected during a normal progressive rollout based on the message, and the `version` tag being empty isn't very intuitive. I'd like to emit a different log message when ramp is being removed.

Maybe moving the code responsible for removing ramp into a separate if statement would help? Hardcoding request values of `Version: "", Percentage: 0` would also make the API call more clear IMO (I'm not sure right now if the value of `vcfg.rampPercentage` matters as long as `desiredRampingVersion` is empty).

internal/controller/genplan.go
carlydf: we're actually removing this concept in the API! now the only concept will be "0% ramp" because that is basically equivalent to "unset ramp" while being less confusing. So this will simplify :D

internal/controller/execplan.go
carlydf: right now, the API requires a 0% ramp to be passed if "unset" is happening. But we are getting rid of that with the struct change because it's confusing

PR #41, opened 2025-05-08
Skip any deployments which have been touched by an external system.

### LINKED ISSUE #13: Detect conflicting version assignment rule updates
It is possible for multiple Temporal clients to make updates to assignment rules that conflict with each other. Eg.
- Two workers deployed in different k8s namespaces or k8s clusters
- A human operator manually editing assignment rules via the `temporal` CLI
- Any other client that manages assignment rules

The controller should set ownership metadata on each assignment rule that it creates, and switch to [manual mode](https://github.com/DataDog/temporal-worker-controller/issues/2) when it observes an active assignment rule that is not managed by itself.

Switching _out_ of manual mode might require another API call to Temporal (eg. to update the ownership metadata).

robholland:
This avoids trying to fight with operators making changes via the CLI, for example.

Closes #13.


internal/controller/worker_controller.go
```diff
@@ -108,6 +108,13 @@ func (r *TemporalWorkerDeploymentReconciler) Reconcile(ctx context.Context, req
 
 	// Compute a new status from k8s and temporal state
 	status, err := r.generateStatus(ctx, l, temporalClient, req, &workerDeploy)
+	if isExternalModification(err) {
+		l.Info("Worker deployment was modified by another entity", "error", err)
+		return ctrl.Result{
+			Requeue:      true,
+			RequeueAfter: 30 * time.Second,
```
robholland: Assuming we may as well slow down here, but maybe we want to keep the usual timings?

internal/controller/worker_controller.go
```diff
@@ -108,6 +108,13 @@ func (r *TemporalWorkerDeploymentReconciler) Reconcile(ctx context.Context, req
 
 	// Compute a new status from k8s and temporal state
 	status, err := r.generateStatus(ctx, l, temporalClient, req, &workerDeploy)
+	if isExternalModification(err) {
+		l.Info("Worker deployment was modified by another entity", "error", err)
```
robholland: Info or Error? Kinda wanted Warn here, but that doesn't exist.

internal/controller/genstatus.go
```diff
@@ -270,9 +292,12 @@ func (r *TemporalWorkerDeploymentReconciler) generateStatus(ctx context.Context,
...
 	// Check if the worker deployment was modified out of band of the controller (eg. via the Temporal CLI)
-	if workerDeploymentInfo.LastModifierIdentity != "temporal-worker-controller" &&
-		workerDeploymentInfo.LastModifierIdentity != "" {
-		// TODO(jlegrone): if it was set by another client, switch to manual mode
+	if wasModifiedExternally(&workerDeploymentInfo) {
+		return nil, &externalModificationError{
+			Resource:             "TemporalWorkerDeployment",
+			Name:                 workerDeploymentName,
+			LastModifierIdentity: workerDeploymentInfo.LastModifierIdentity,
+		}
```
jlegrone: Rather than returning early here I think we should continue but with the same behavior as if the "manual" rollout mode were enabled.

If we skip updating the worker's `status`, then the k8s state and Temporal won't agree. This could be confusing, especially at a time when humans are making out of band changes and are likely involved in debugging the system.

As discussed in the sync we will probably want to continue autoscaling worker versions in this situation. That will also require continuing the reconciliation loop.

internal/controller/genstatus.go
```diff
@@ -270,9 +292,12 @@ func (r *TemporalWorkerDeploymentReconciler) generateStatus(ctx context.Context,
 	defaultVersionID = routingConfig.CurrentVersion
 
 	// Check if the worker deployment was modified out of band of the controller (eg. via the Temporal CLI)
-	if workerDeploymentInfo.LastModifierIdentity != "temporal-worker-controller" &&
-		workerDeploymentInfo.LastModifierIdentity != "" {
-		// TODO(jlegrone): if it was set by another client, switch to manual mode
+	if wasModifiedExternally(&workerDeploymentInfo) {
```
jlegrone: I think there's a race condition if configuration on the Temporal side is updated _after_ the controller calls the API to get worker deployment info. In that situation, this condition wouldn't evaluate `true` and we would proceed to potentially overwrite the out of band changes.

In a previous Temporal API version there was a `ConflictToken` field that could be used to avoid this: https://github.com/DataDog/temporal-worker-controller/blob/6d4cc8642ee14301de233ad8f8d506c874ace74f/internal/controller/execplan.go#L65

Is there a way to address the same issue with current APIs?

robholland: I've updated the code so that it uses the token we received during planning to perform the version updates. We really need to figure out a way to test this stuff though, it's hard to be confident in this currently.

internal/controller/genstatus.go
robholland: Makes sense, I've now mapped this to act like Manual rollout strategy.

api/v1alpha1/worker_types.go
```diff
@@ -122,6 +122,9 @@ type TemporalWorkerDeploymentStatus struct {
 
 	// TODO(jlegrone): Add description
 	VersionConflictToken []byte `json:"versionConflictToken"`
+
+	// ExternallyModified indicates if the worker deployment was modified outside of the controller (e.g. via CLI)
+	ExternallyModified bool `json:"externallyModified,omitempty"`
```
jlegrone: Maybe it would be useful to humans if the `LastModifierIdentity` were encoded here as a string instead of a bool?

internal/controller/genplan.go
```diff
@@ -241,6 +241,12 @@ func getVersionConfig(l logr.Logger, strategy temporaliov1alpha1.RolloutStrategy
 		return nil
 	}
 
+	// Do nothing if the Temporal Deployment has been modified out of band
+	if status.ExternallyModified {
+		l.Info("Forcing manual rollout strategy since deployment was modified externally")
+		return nil
+	}
+
```
jlegrone: I'm concerned that adding a check for `status.ExternallyModified` everywhere that we currently read `spec.RolloutStrategy` will make it more onerous to write tests and increase the risk of dangerous bugs like proceeding with operations that would have otherwise been skipped in manual mode.

What do you think of modifying `spec.RolloutStrategy` in-place when we detect an out of band change, and only setting `status.ExternallyModified` for diagnostic purposes? Modifying the spec in-memory [does not persist changes to the underlying k8s resource](https://github.com/DataDog/temporal-worker-controller/blob/18a2b1fe16df5589d37aba637899384099d3369e/internal/controller/worker_controller.go#L128-L130), but we could also refactor the source of truth for rollout strategy into a separate variable if that would be cleaner.

api/v1alpha1/worker_types.go
robholland: Yeah I prefer that, keeps the `Status` as factual.

internal/controller/genplan.go
robholland: I've adjusted `genplan` to do this in-line rather than adjusting specs as I think that makes it clearer what is happening. I'm not sure that ExternallyModified on the status object adds much value after all as what that means is entirely up to the controller anyway.

jlegrone (approved):

internal/controller/genplan.go
jlegrone: Thanks, I like that approach!

internal/controller/worker_controller.go
jlegrone: Info seems fine for now, I agree warn would be nice though.

PR #42, opened 2025-05-16
Refactor genstatus for separation of concerns.

robholland:
Split apart orchestration, Temporal API and k8s API usage.

### DIFF
--- a/internal/controller/state_mapper.go
+++ b/internal/controller/state_mapper.go
@@ -98,8 +98,8 @@ func (m *stateMapper) mapWorkerDeploymentVersion(versionID string) *v1alpha1.Wor
 		// Set ramp percentage if this is a ramping version
 		// TODO(carlydf): Support setting any ramp in [0,100]
 		// NOTE(rob): We are now setting any ramp > 0, is that correct?
-		if temporalVersion.Status == v1alpha1.VersionStatusRamping && temporalVersion.RampPercentage > 0 {
-			version.RampPercentage = &temporalVersion.RampPercentage
+		if temporalVersion.Status == v1alpha1.VersionStatusRamping && m.temporalState.RampPercentage > 0 {
+			version.RampPercentage = &m.temporalState.RampPercentage
 		}
 
 		// Set task queues
--- a/internal/temporal/worker_deployment.go
+++ b/internal/temporal/worker_deployment.go
@@ -21,12 +21,11 @@ import (
 
 // VersionInfo contains information about a specific version
 type VersionInfo struct {
-	VersionID      string
-	Status         temporaliov1alpha1.VersionStatus
-	DrainedSince   *time.Time
-	RampPercentage float32
-	TaskQueues     []temporaliov1alpha1.TaskQueue
-	TestWorkflows  []temporaliov1alpha1.WorkflowExecution
+	VersionID     string
+	Status        temporaliov1alpha1.VersionStatus
+	DrainedSince  *time.Time
+	TaskQueues    []temporaliov1alpha1.TaskQueue
+	TestWorkflows []temporaliov1alpha1.WorkflowExecution
 }
 
 // TemporalWorkerState represents the state of a worker deployment in Temporal
@@ -96,7 +95,6 @@ func GetWorkerDeploymentState(
 			versionInfo.Status = temporaliov1alpha1.VersionStatusCurrent
 		} else if version.Version == routingConfig.RampingVersion {
 			versionInfo.Status = temporaliov1alpha1.VersionStatusRamping
-			versionInfo.RampPercentage = routingConfig.RampingVersionPercentage
 		} else if drainageStatus == temporalClient.WorkerDeploymentVersionDrainageStatusDraining {
 			versionInfo.Status = temporaliov1alpha1.VersionStatusDraining
 		} else if drainageStatus == temporalClient.WorkerDeploymentVersionDrainageStatusDrained {
--- a/internal/controller/state_mapper_test.go
+++ b/internal/controller/state_mapper_test.go
@@ -100,9 +100,8 @@ func TestMapToStatus(t *testing.T) {
 		VersionConflictToken: []byte("test-token"),
 		Versions: map[string]*temporal.VersionInfo{
 			"worker.v1": {
-				VersionID:      "worker.v1",
-				Status:         temporaliov1alpha1.VersionStatusCurrent,
-				RampPercentage: 100,
+				VersionID: "worker.v1",
+				Status:    temporaliov1alpha1.VersionStatusCurrent,
 				TaskQueues: []temporaliov1alpha1.TaskQueue{
 					{Name: "queue1"},
 				},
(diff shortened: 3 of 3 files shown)


internal/controller/state_mapper.go
```diff
@@ -0,0 +1,144 @@
...
+)
+
+// StateMapper maps between Kubernetes and Temporal states
+type StateMapper struct {
+	k8sState      *k8s.DeploymentState
+	temporalState *temporal.TemporalWorkerState
+}
+
+// NewStateMapper creates a new state mapper
+func NewStateMapper(k8sState *k8s.DeploymentState, temporalState *temporal.TemporalWorkerState) *StateMapper {
```
jlegrone: Since these symbols don't seem to be used outside the package, I think they can be made private.

Not a big deal since it's already under /internal, but still good hygiene in case anything here is ever refactored into a public package.

internal/controller/state_mapper.go
```diff
@@ -0,0 +1,144 @@
...
+	return &StateMapper{
+		k8sState:      k8sState,
+		temporalState: temporalState,
+	}
+}
+
+// MapToStatus converts the states to a CRD status
+func (m *StateMapper) MapToStatus(desiredVersionID string) *v1alpha1.TemporalWorkerDeploymentStatus {
```
jlegrone: Same here
```suggestion
func (m *StateMapper) mapToStatus(desiredVersionID string) *v1alpha1.TemporalWorkerDeploymentStatus {
```

internal/temporal/worker_deployment.go
```diff
@@ -0,0 +1,233 @@
...
+type VersionStatus string
+
+const (
+	VersionStatusNotRegistered VersionStatus = "NotRegistered"
+	VersionStatusInactive      VersionStatus = "Inactive"
+	VersionStatusRamping       VersionStatus = "Ramping"
+	VersionStatusCurrent       VersionStatus = "Current"
+	VersionStatusDraining      VersionStatus = "Draining"
+	VersionStatusDrained       VersionStatus = "Drained"
+)
```
jlegrone: I'm not super comfortable with all of the `v1alpha1` API duplication. I imagine this will make refactoring take longer and create opportunities for bugs in mapping between types.

Why not use the `v1alpha1` types more directly? It seems like the business logic for querying k8s and temporal state could still be split out in the same way.

robholland: Yep, makes sense. I've switched to use `v1alpha1` types where I can.

internal/temporal/worker_deployment.go
```diff
@@ -0,0 +1,208 @@
...
+	TestWorkflows  []temporaliov1alpha1.WorkflowExecution
+}
+
+// TemporalWorkerState represents the state of a worker deployment in Temporal
+type TemporalWorkerState struct {
+	DefaultVersionID     string
+	VersionConflictToken []byte
+	RampingVersionID     string
+	RampPercentage       float32
+	RampingSince         *metav1.Time
```
jlegrone: Since we already have these fields, can we delete `VersionInfo.RampPercentage`?

jlegrone: Does `RampingSince` indicate the time that the first nonzero ramp value for the current ramping version was enabled, or the time that the ramp value was last modified?

A godoc comment clarifying this would be nice.

jlegrone (approved):
lgtm pending last two comments

internal/temporal/worker_deployment.go
robholland: We could drop the field, but I think it's better modelling to keep it. A `RampPercentage` should only be set on the target version, and I think having `VersionInfo.RampPercentage` with no need to have access to the `TemporalWorkerState` is cleaner and more intuitive.

robholland: It's when `RampingVersionID` was last set. Added a comment.

internal/temporal/worker_deployment_test.go
```diff
@@ -0,0 +1,183 @@
...
+	assert.Equal(t, rampingSince.Time.Unix(), state.RampingSince.Time.Unix())
+	assert.Equal(t, "test-controller", state.LastModifierIdentity)
+	assert.Equal(t, []byte("test-token"), state.VersionConflictToken)
+	assert.Equal(t, 2, len(state.Versions))
+
+	// Verify versions
+	assert.Equal(t, temporaliov1alpha1.VersionStatusCurrent, state.Versions["worker.v1"].Status)
+	assert.Equal(t, temporaliov1alpha1.VersionStatusRamping, state.Versions["worker.v2"].Status)
+	assert.Equal(t, float32(25.0), state.Versions["worker.v2"].RampPercentage)
+}
```
jlegrone: It seems like these tests aren't validating business logic, just setting field values?

```suggestion
```

internal/k8s/deployments_test.go
```diff
@@ -0,0 +1,185 @@
...
+
+		// Set up the refs map
+		state.DeploymentRefs["worker.v1"] = newObjectRef(deploy1)
+		state.DeploymentRefs["worker.v2"] = newObjectRef(deploy2)
+
+		return state
+	}
+
+	// Use our mock to get a valid DeploymentState
+	state := mockGetDeploymentState()
```
jlegrone: Why not hardcode state directly rather than declaring a "mock" function?

internal/k8s/deployments_test.go
```diff
@@ -0,0 +1,185 @@
...
+	assert.Equal(t, "worker-v1", state.Deployments["worker.v1"].Name)
+	assert.Equal(t, "worker-v2", state.Deployments["worker.v2"].Name)
+
+	// Verify the deployments are sorted by creation time
+	assert.Equal(t, "worker-v1", state.DeploymentsByTime[0].Name)
+	assert.Equal(t, "worker-v2", state.DeploymentsByTime[1].Name)
+
+	// Verify refs are correctly created
+	assert.Equal(t, "worker-v1", state.DeploymentRefs["worker.v1"].Name)
+	assert.Equal(t, "worker-v2", state.DeploymentRefs["worker.v2"].Name)
```
jlegrone: I'm not sure I see the value of this test since it's checking that the mock function returns a result which is hardcoded in the same test.

internal/temporal/worker_deployment.go
jlegrone: My concern is about the possibility of failing to update the value correctly in both places, leading to undefined behavior if both values are later read by different parts of the program.

Maybe we could avoid the need for double bookkeeping while supporting a similar access pattern by adding a `GetRampPercentage() (float32, bool)` method to `TemporalWorkerState`?

robholland: I'm not convinced this could drift given how it's set (the per-version one is set based on the worker state one), but I've removed the per-version field anyway and just use the worker state field when required.

internal/temporal/worker_deployment_test.go
robholland: Yes, sorry, I shall up the bar for what gets tested :)

internal/k8s/deployments_test.go
robholland: Updated to use a faked client to exercise the real code.

internal/k8s/deployments_test.go
robholland: Yep, dropped this.

PR #45, opened 2025-06-10
Rollback ramp

### LINKED ISSUE #37: Rollback from B -> A while B is ramping and A is still target should route 100% of traffic back to A
If A is current and B starts ramping, and then the user notices some problem with B so they make A the target again, we want the controller to revert the ramp of B so that 100% of traffic goes to A. 

However, the current code causes B to continue ramping forever, because the controller sees that default == target == A and thinks no changes are needed.

We need to fix this.

robholland:
Fixes #37.

### DIFF
--- a/internal/controller/execplan.go
+++ b/internal/controller/execplan.go
@@ -94,21 +94,12 @@ func (r *TemporalWorkerDeploymentReconciler) executePlan(ctx context.Context, l 
 			}); err != nil {
 				return fmt.Errorf("unable to set current deployment version: %w", err)
 			}
-			if _, err := deploymentHandler.UpdateVersionMetadata(ctx, sdkclient.WorkerDeploymentUpdateVersionMetadataOptions{
-				Version: vcfg.VersionID,
-				MetadataUpdate: sdkclient.WorkerDeploymentMetadataUpdate{
-					UpsertEntries: map[string]interface{}{
-						// TODO(carlydf): Add info about which k8s resource initiated the last write to the deployment
-						"temporal.io/managed-by": controllerIdentity,
-					},
-				},
-			}); err != nil { // would be cool to do this atomically with the update
-				return fmt.Errorf("unable to update metadata after setting current deployment: %w", err)
-			}
 		} else {
-			err := awaitVersionRegistration(ctx, l, deploymentHandler, p.TemporalNamespace, vcfg.VersionID)
-			if err != nil {
-				return fmt.Errorf("error waiting for version to register, did your pollers start successfully?: %w", err)
+			if vcfg.VersionID != "" {
+				err := awaitVersionRegistration(ctx, l, deploymentHandler, p.TemporalNamespace, vcfg.VersionID)
+				if err != nil {
+					return fmt.Errorf("error waiting for version to register, did your pollers start successfully?: %w", err)
+				}
 			}
 
 			if vcfg.RampPercentage > 0 {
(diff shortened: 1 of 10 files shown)


internal/controller/genplan.go
```diff
@@ -118,14 +118,14 @@ func (r *TemporalWorkerDeploymentReconciler) generatePlan(
...
-	// Convert version config
-	if planResult.VersionConfig != nil {
-		plan.UpdateVersionConfig = &versionConfig{
-			conflictToken:  planResult.VersionConfig.ConflictToken,
-			versionID:      planResult.VersionConfig.VersionID,
-			setDefault:     planResult.VersionConfig.SetDefault,
-			rampPercentage: planResult.VersionConfig.RampPercentage,
-		}
+	// Convert version configs
+	for _, vcfg := range planResult.VersionConfigs {
```
carlydf: Is there a reason that `planResult.VersionConfigs` should be a different type than `plan.UpdateVersionConfigs`? If the fields are identical, we could just make them the same.

Although making them separate does allow us to add or remove fields from one or the other and not both in the future, if that is desired.

Anyways, this is fine for now, I'm just wondering about the reasoning for this verbose struct copy, and whether we can make it tighter!

internal/controller/execplan.go
```diff
@@ -131,6 +131,17 @@ func (r *TemporalWorkerDeploymentReconciler) executePlan(ctx context.Context, l
 			}); err != nil { // would be cool to do this atomically with the update
 				return fmt.Errorf("unable to update metadata after setting ramping deployment: %w", err)
 			}
+		} else if ramp := vcfg.rampPercentage; ramp == 0 {
```
carlydf: I don't think you actually need this else if ramp == 0 case, because if `ramp==0`, the `SetRampingVersion` call above would handle things the same way (if you also remove the `ramp>0` condition from the above block). I think we still want to update the metadata even when setting ramp to 0, so it would be desirable to put all that in one block for that reason as well.

carlydf: If the structure was
```
if vcfg.SetCurrent {
  SetCurrent
  UpdateMetadata
} else {
  SetRamping
  UpdateMetadata
}
```
I think that works?

carlydf: or even
```
if vcfg.SetCurrent {
  SetCurrent
} else {
  SetRamping
}
UpdateMetadata
```

carlydf: Sidenote that i can't comment on because you didnt change that line. Now that there is a list of vcfgs, do you still want to `return` if there is an error setting ramp? or setting current? Or do we want to log the error and continue to the next vcfg in the list. It's too late in my work day for me to fully think this through right now, but something we should think about!

internal/planner/planner.go
```diff
@@ -286,41 +286,31 @@ func getVersionConfigDiff(
...
+) []*VersionConfig {
+	var versionConfigs []*VersionConfig
 
-	// Set default version if there isn't one yet
-	if status.DefaultVersion == nil {
-		vcfg.SetDefault = true
-		vcfg.RampPercentage = 0
-		return vcfg
-	}
+	vcfg := getVersionConfig(l, strategy, status)
```
carlydf: is this basically the vcfg for the target version?

internal/planner/planner.go
```diff
@@ -286,41 +286,31 @@ func getVersionConfigDiff(
...
+		versionConfigs = append(versionConfigs, vcfg)
 	}
 
-	// Don't make updates if desired ramping version is already the target, and ramp percentage is correct
-	if !vcfg.SetDefault &&
-		vcfg.VersionID == status.TargetVersion.VersionID &&
-		status.TargetVersion.RampPercentage != nil &&
-		vcfg.RampPercentage == *status.TargetVersion.RampPercentage {
-		return nil
+	for _, vcfg := range status.DeprecatedVersions {
```
carlydf: Lets say version A is current, and then version B is the target version and starts rolling out gradually.
Then someone notices a problem with B so they make A the new target (let's say with Immediate cutover strategy).

We want the ramp of B to be set to 0 / the ramp to be "unset".
It seems like that would only happen if B is in status.DeprecatedVersions. Is that the case? (The contents of `DeprecatedVersions` may have changed since I last read the code, or I forgot)

internal/controller/genplan.go
robholland: Removed the duplicate type.

internal/controller/execplan.go
robholland: Now I better understand the versioning Workflow I can see that it's not possible to have multiple versions with a ramp set. Also, we will never need to be SetCurrent and SetRamping at the same time, so I've removed the multiple version config, and gone back to just a single. I've applied the simplifications you mentioned, much nicer :)

internal/planner/planner.go
robholland: The ramping version would indeed have been in DeprecatedVersions, but my assumption was that it was possible for multiple of those to have a ramp set for some reason. I now understand that is not the case, so I've being more explicit and using RampingVersion as your PR did.

internal/planner/planner.go
robholland: It may be specific to the target version, such as updating the ramp or marking it current. It may (now) also be a ramp deletion (so VersionID == "").

jlegrone (approved):

carlydf (approved):
looks good!

PR #65, opened 2025-06-26
Prefix Build ID with image tag of first container, suffix with short pod spec hash

### LINKED ISSUE #53: Add image tag to build id

carlydf:
Closes #53

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,5 +1,19 @@
-## This is a work in progress :)
+# Temporal Worker Controller Documentation
+
+This directory contains extended documentation for the Temporal Worker Controller project. The documentation is organized into different categories to help you find the information you need.
 
 
-- Max TemporalWorkerDeploymentName = 63
-- Max BuildID = 63
+This documentation structure is designed to support various types of technical documentation, such as:
+
+- **Limits**: Technical constraints and limitations of the system
+- **Runbooks**: Operational procedures and troubleshooting guides (planned)
+- **Conceptual Guides**: High-level explanations of concepts and architecture (planned)
+
+## Index
+
+### [Limits](limits.md)
+Technical constraints and limitations of the Temporal Worker Controller system, including maximum field lengths and other operational boundaries.
+
+---
+
+*Note: This documentation structure is designed to grow with the project.*
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,0 +1,5 @@
+## This is a work in progress :)
+
+
+- Max TemporalWorkerDeploymentName = 63
+- Max BuildID = 63
--- a/internal/controller/k8s.io/utils/utils.go
+++ b/internal/controller/k8s.io/utils/utils.go
@@ -57,9 +57,10 @@ func deepHashObject(hasher hash.Hash, objectToWrite interface{}) {
 // ComputeHash returns a hash value calculated from pod template and
 // a collisionCount to avoid hash collision. The hash will be safe encoded to
 // avoid bad words.
+// If `short` is true, the hash is truncated to 4 digits
 //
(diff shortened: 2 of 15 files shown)


internal/controller/k8s.io/utils/utils.go
```diff
@@ -57,9 +57,10 @@ func deepHashObject(hasher hash.Hash, objectToWrite interface{}) {
 // ComputeHash returns a hash value calculated from pod template and
 // a collisionCount to avoid hash collision. The hash will be safe encoded to
 // avoid bad words.
+// If `short` is true, the hash is truncated to 4 digits
 //
 // Copied from https://github.com/kubernetes/kubernetes/blob/86fec81606b579cc478a30656c29ddb400a72dc6/pkg/controller/controller_utils.go#L1174
-func ComputeHash(template *v1.PodTemplateSpec, collisionCount *int32) string {
+func ComputeHash(template *v1.PodTemplateSpec, collisionCount *int32, short bool) string {
```
jlegrone, line 63: Mind adding a unit test in this package?

internal/k8s/deployments_test.go
```diff
@@ -203,3 +204,84 @@ func TestGetDeploymentState(t *testing.T) {
 	assert.Equal(t, "worker-v1", state.DeploymentRefs["test-worker.v1"].Name)
 	assert.Equal(t, "worker-v2", state.DeploymentRefs["test-worker.v2"].Name)
 }
+
+func TestGenerateBuildID_SameImageDifferentPods(t *testing.T) {
```
jlegrone: Could these be table tests?

docs/README.md
```diff
@@ -0,0 +1,5 @@
+## This is a work in progress :)
```
carlydf: todo: clarify plan here

internal/k8s/deployments.go
```diff
@@ -103,25 +113,61 @@ func NewObjectRef(obj client.Object) *v1.ObjectReference {
...
 
+// ComputeVersionID generates a version ID from the worker deployment spec
+func ComputeVersionID(w *temporaliov1alpha1.TemporalWorkerDeployment) string {
+	return ComputeWorkerDeploymentName(w) + versionIDSeparator + ComputeBuildID(w)
+}
+
+func ComputeBuildID(w *temporaliov1alpha1.TemporalWorkerDeployment) string {
+	if containers := w.Spec.Template.Spec.Containers; len(containers) > 0 {
+		if img := containers[0].Image; img != "" {
+			shortHashSuffix := k8sResourceNameSeparator + utils.ComputeHash(&w.Spec.Template, nil, true)
```
robholland, line 119: I think this is slightly easier to read if you build the suffix after you build the imagePrefix.

internal/k8s/deployments.go
```diff
@@ -103,25 +113,61 @@ func NewObjectRef(obj client.Object) *v1.ObjectReference {
...
+			s = reference.Path(v) // -> library/busybox
+		default:
+		}
+	}
+	return cleanAndTruncateString(s, maxLen)
+}
+
+// Truncates string to the first n characters, and then removes characters that can't be in a kubernetes resource name.
+// Pass n = -1 to skip truncation.
+func cleanAndTruncateString(s string, n int) string {
```
robholland, line 158: I feel like you should strip invalid characters, and then truncate.

internal/k8s/deployments.go
carlydf: the reason it is first, is because I need the length of the suffix in order to decide how much to truncate the prefix! and I don't want to hardcode the length of the short hash suffix, in case we want to change it later. 

Does that make sense?

PR #72, opened 2025-06-26
Remove license tooling and update module name

jlegrone:
## What was changed

- Updates go module from `github.com/DataDog/temporal-worker-controller` to `github.com/temporalio/temporal-worker-controller`
- Removes licensecheck and copyright tools that were specific to Datadog

## Why?

Repo is now owned by `temporalio`

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/DataDog/temporal-worker-controller
+module github.com/temporalio/temporal-worker-controller
```
jlegrone, line 1: Should this be updated to a vanity url a la `go.temporal.io/sdk`?

carlydf: not yet, since I don't want to block on that

PR #79, opened 2025-06-27
Wait for at least last step duration

### LINKED ISSUE #25: Guarantee that the ramps specified in progressive rollout spend at least the requested time in each step
Currently, the durations are evaluated with respect to `rampingSinceTime`. This means that with the following cutover strategy:
```
    steps:
      # Increase traffic from 1% to 10% over 15 seconds
      - rampPercentage: 1
         pauseDuration: 5s
      # Increase traffic to 50% and wait 1 minute
      - rampPercentage: 50
         pauseDuration: 1m
```
It is possible to ramp traffic to 1%, then have 1m5s failed Reconcile loops, then immediately set the target version to current, because the total ramp time has elapsed. We should instead use `rampingSinceTime` in conjunction with the last time the routing info of that deployment was changed.

jlegrone:
## What was changed

- Refactors progressive rollout logic to resume from the step that is equal to or less than the current ramp value
- Requires that the wait time for each step be >= the time since the last ramp value change
- Adds validation to require at least one step in progressive rollouts
- Adds validation to require step pause duration at least 30s
- Implements a sink for writing controller logs to test output
- Refactors test helpers out of public package

## Why?

Closes #25 

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/api/v1alpha1/temporalworker_webhook.go
+++ b/api/v1alpha1/temporalworker_webhook.go
@@ -73,16 +73,20 @@ func (r *TemporalWorkerDeployment) validateForUpdateOrCreate(ctx context.Context
 		return nil, apierrors.NewBadRequest("expected a TemporalWorkerDeployment")
 	}
 
+	return validateForUpdateOrCreate(nil, dep)
+}
+
+func validateForUpdateOrCreate(old, new *TemporalWorkerDeployment) (admission.Warnings, error) {
 	var allErrs field.ErrorList
 
-	if len(dep.GetName()) > maxTemporalWorkerDeploymentNameLen {
+	if len(new.GetName()) > maxTemporalWorkerDeploymentNameLen {
 		allErrs = append(allErrs,
-			field.Invalid(field.NewPath("metadata.name"), dep.Name, fmt.Sprintf("cannot be more than %d characters", maxTemporalWorkerDeploymentNameLen)),
+			field.Invalid(field.NewPath("metadata.name"), new.GetName(), fmt.Sprintf("cannot be more than %d characters", maxTemporalWorkerDeploymentNameLen)),
 		)
 	}
 
-	if dep.Spec.RolloutStrategy.Strategy == UpdateProgressive {
-		rolloutSteps := dep.Spec.RolloutStrategy.Steps
+	if new.Spec.RolloutStrategy.Strategy == UpdateProgressive {
+		rolloutSteps := new.Spec.RolloutStrategy.Steps
 		if len(rolloutSteps) == 0 {
 			allErrs = append(allErrs,
 				field.Invalid(field.NewPath("spec.cutover.steps"), rolloutSteps, "steps are required for Progressive cutover"),
@@ -108,7 +112,7 @@ func (r *TemporalWorkerDeployment) validateForUpdateOrCreate(ctx context.Context
 	}
 
 	if len(allErrs) > 0 {
-		return nil, newInvalidErr(dep, allErrs)
+		return nil, newInvalidErr(new, allErrs)
 	}
 
 	return nil, nil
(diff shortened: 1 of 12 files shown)


api/v1alpha1/temporalworker_webhook_test.go
```diff
@@ -2,102 +2,132 @@
 //
 // This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2024 Datadog, Inc.
 
-package v1alpha1
+package v1alpha1_test
```
jlegrone, line 5: FYI: this is needed in order to avoid import cycles after moving test helpers to a separate package.

api/v1alpha1/worker_types.go
```diff
@@ -215,6 +215,10 @@ type TargetWorkerDeploymentVersion struct {
 	// Only set when Status is VersionStatusRamping.
 	// +optional
 	RampingSince *metav1.Time `json:"rampingSince"`
+
+	// RampLastModifiedAt is the time when the ramp percentage was last changed.
```
carlydf: nit: If the ramping version changes, but the percentage stays the same (ie. 5% -> B, and then 5% -> C), would `RampLastModifiedAt` change? I'm assuming yes it does change, because that's what makes sense and what the variable name implies. Maybe it's obvious, but the comment only says "percentage" which could be misleading

carlydf (approved):
nice!

internal/planner/planner.go
```diff
@@ -338,39 +339,75 @@ func getVersionConfigDiff(
...
+func handleProgressiveRollout(
+	steps []temporaliov1alpha1.RolloutStep,
+	currentTime time.Time, // avoid calling time.Now() inside function to make it easier to test
+	rampLastModifiedAt *metav1.Time,
+	targetRampPercentage *float32,
+	vcfg *VersionConfig,
+) *VersionConfig {
+	// Set current right away if there are no steps.
+	// This is equivalent to the AllAtOnce cutover strategy.
+	if len(steps) == 0 {
```
robholland, line 361: Progressive rollouts without steps is not valid according to our validator, so do we need to handle this case?

api/v1alpha1/temporalworker_webhook.go
```diff
@@ -73,21 +73,51 @@ func (r *TemporalWorkerDeployment) validateForUpdateOrCreate(ctx context.Context
...
 
-	if len(dep.Name) > maxTemporalWorkerDeploymentNameLen {
+	if len(new.GetName()) > maxTemporalWorkerDeploymentNameLen {
 		allErrs = append(allErrs,
-			field.Invalid(field.NewPath("name"), dep.Name, fmt.Sprintf("cannot be more than %d characters", maxTemporalWorkerDeploymentNameLen)),
+			field.Invalid(field.NewPath("metadata.name"), new.GetName(), fmt.Sprintf("cannot be more than %d characters", maxTemporalWorkerDeploymentNameLen)),
 		)
 	}
 
+	if new.Spec.RolloutStrategy.Strategy == UpdateProgressive {
```
robholland: I wonder if the validation for strategies should live with the strategy code rather than living in the webhook admission system?

internal/planner/planner.go
jlegrone: I included this for robustness in case the validator is ever skipped (eg. right now the validation webhook is not deployed with our helm chart). But actually I think to be safer the behavior should be to emulate manual mode rather than all at once, so I'll make that change.

api/v1alpha1/worker_types.go
jlegrone:
```suggestion
	// RampLastModifiedAt is the time when the ramp percentage was last changed for the target version.
```

api/v1alpha1/temporalworker_webhook.go
jlegrone: I agree that would make a lot of sense, but unfortunately it's a little tricky to avoid import cycles if we want to import a validation function in the `v1alpha1` package. I refactored the rollout validation into a separate function to make it easier to move it around in the future though.

PR #80, opened 2025-07-07
Remove Ramping version from status

robholland:
## What was changed

Removed Ramping version from the status.

Also adjusted code around the assumed presence of CurrentVersion and TargetVersion which was inverted in comments and code.

## Why?

Seeing the Ramping Version in the status may confuse users.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


api/v1alpha1/worker_types.go
```diff
@@ -105,19 +105,14 @@ type TemporalWorkerDeploymentStatus struct {
 	// TargetVersion is the desired next version. If TargetVersion.Deployment is nil,
 	// then the controller should create it. If not nil, the controller should
 	// wait for it to become healthy and then move it to the CurrentVersion.
+	// This must never be nil.
 	TargetVersion *TargetWorkerDeploymentVersion `json:"targetVersion"`
```
jlegrone:
```suggestion
TargetVersion TargetWorkerDeploymentVersion `json:"targetVersion"`
```

If it's expected that it can't be nil, what do you think about enforcing that in the API?

robholland: Yes I think we should, what did you have in mind?

internal/planner/planner.go
```diff
@@ -132,49 +127,53 @@ func getDeleteDeployments(
...
-			if (time.Since(version.DrainedSince.Time) > getDeleteDelay(config.Spec)+getScaledownDelay(config.Spec)) &&
+			if (time.Since(version.DrainedSince.Time) > getDeleteDelay(spec)+getScaledownDelay(spec)) &&
 				*d.Spec.Replicas == 0 {
 				deleteDeployments = append(deleteDeployments, d)
 			}
 		case temporaliov1alpha1.VersionStatusNotRegistered:
 			// NotRegistered versions are versions that the server doesn't know about.
 			// Only delete if it's not the target version.
-			if config.Status.TargetVersion == nil || config.Status.TargetVersion.VersionID != version.VersionID {
+			if status.TargetVersion.VersionID != version.VersionID {
```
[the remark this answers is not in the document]
robholland: TargetVersion will never be `nil`.

robholland:
Co-pilot's state_mapper comment about `status.CurrentVersion =` is wrong, the function handles producing a `nil` correctly. The `mapTargetVersion` comment doesn't apply, target version can be empty, the controller unconditionally sets it.

api/v1alpha1/worker_types.go
robholland: I've added some validation around this.

jlegrone: I meant to remove the pointer value (see suggested change) -- I think that should make it required in the json schema and also avoid the need for nil checks in the controller code.

robholland: Done.

Issue #28, opened 2025-04-01
Add events to the TemporalWorkerDeployment CRD when there is a problem


carlydf:
ie. 
- Temporal namespace doesn’t exist for the TemporalConnection
- Temporal authn/authz not working/not allowed for the TemporalConnection

Shivs11:
candidate for an event: pod startup failure.

PR #82, opened 2025-07-11
Add a version limit.

### LINKED ISSUE #7: Garbage collect version assignment rules
In Temporal 1.25 by default a max of 100 assignment rules are allowed. Right now the controller becomes deadlocked after enough rollouts to hit this limit.

robholland:
This prevents the controller from trying to create deployment versions when the Temporal server may reject them.

The default limit is 75, which is 75% of the default server side limit of 100.

The user may override the limit by setting MaxVersions on the CRD. If the user raises this sufficiently that it matches the server side limit and allows deploys to build up until that point, they will start to fail as the controller is not able to make progress.

Fixes #7.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


api/v1alpha1/temporalworker_webhook_test.go
```diff
@@ -75,6 +75,22 @@ func TestTemporalWorkerDeployment_ValidateCreate(t *testing.T) {
...
 		},
+		"maxVersions below minimum": {
+			obj: testhelpers.ModifyObj(testhelpers.MakeTWDWithName("max-versions-too-low"), func(obj *temporaliov1alpha1.TemporalWorkerDeployment) *temporaliov1alpha1.TemporalWorkerDeployment {
+				maxVersions := int32(0)
+				obj.Spec.MaxVersions = &maxVersions
+				return obj
+			}),
+			errorMsg: "spec.maxVersions: Invalid value: 0: maxVersions must be at least 1",
+		},
+		"maxVersions above maximum": {
```
robholland: Not sure if we want to validate a maximum here?

internal/planner/planner.go
```diff
@@ -222,6 +230,11 @@ func shouldCreateDeployment(
 	k8sState *k8s.DeploymentState,
 	config *Config,
 ) bool {
+	// Check if we're at the version limit - if so, don't create new deployments
+	if config.Status.VersionCount >= getMaxVersions(config.Spec) {
+		return false
```
robholland, line 200: This will currently silently skip creating the deployment. We can log here, but that will most likely be pretty spammy as it will fire every reconciliation loop until deployments are freed up. I think we should still probably log though, thoughts?

robholland: A k8s error event would be an option here as well, I'm not very familiar with those.

api/v1alpha1/worker_types.go
```diff
@@ -60,6 +69,12 @@ type TemporalWorkerDeploymentSpec struct {
 
 	// TODO(jlegrone): add godoc
 	WorkerOptions WorkerOptions `json:"workerOptions"`
+
+	// MaxVersions specifies the maximum number of versions that can exist in the worker deployment
+	// before blocking new deployments. When this limit is reached, new deployments will be blocked
+	// to prevent hitting Temporal's assignment rule limit. Defaults to 75.
```
Shivs11: I suggested the above since `hitting Temporal's assignment rule limit` sounds like you are referencing an older version of worker-versioning (we don't use assignment rules in our latest version)

internal/planner/planner.go
Shivs11: yeah 100% agree on trying to alert the user when they hit this - my initial idea was to use a combination of logging statements + adding k8s events (I wish this was part of public-preview since it would fit in so nicely: https://github.com/temporalio/temporal-worker-controller/issues/28)

As an alternative, we could also use logging + modifying the existing twd by adding some labels/annotations to it? 

```
    if twd.Labels == nil {
        twd.Labels = make(map[string]string)
    }
    twd.Labels["temporal.io/version-limit-reached"] = "true"
    r.Update(ctx, twd)

```

api/v1alpha1/worker_types.go
Shivs11:
```suggestion
          // MaxVersions defines the maximum number of worker deployment versions allowed. 
          // This helps prevent hitting Temporal's default limit of 100 versions per deployment. 
          // Defaults to 75. Users can override this by explicitly setting a higher value in 
          // the CRD, but should exercise caution: once the 100-version limit is reached, 
          // Temporal attempts to delete an eligible version. If no version is eligible for 
          // deletion, new deployments get blocked which prevents
          // the controller from making progress.

```

api/v1alpha1/temporalworker_webhook.go
```diff
@@ -126,6 +130,25 @@ func validateRolloutStrategy(s RolloutStrategy) []*field.Error {
...
+		if *maxVersions < 1 {
+			allErrs = append(allErrs,
+				field.Invalid(field.NewPath("spec.maxVersions"), *maxVersions, "maxVersions must be at least 1"),
+			)
+		}
+		if *maxVersions > 500 {
+			allErrs = append(allErrs,
+				field.Invalid(field.NewPath("spec.maxVersions"), *maxVersions, "maxVersions cannot exceed 500"),
+			)
+		}
```
Shivs11: slightly confused - do we be this relaxed and allow the user to set a limit of 500? Sure, in a happy path scenario, where every new version after the 100th one gets created, a user won't notice an error. But if Temporal can't delete a version after the 100th one, because no version is eligible for deletion, then it could become a little problematic 

Maybe I am forgetting a discussion we had but if I am not, should we bring this number down?

robholland: Probably, I'm not sure if the validation is useful. We can assume that the server limit will never be changed and so bring this down to 100. Wasn't sure what our plans were for the server side limit.

robholland: Brought down to 100.

api/v1alpha1/worker_types.go
robholland: Merged your suggested changes.

internal/planner/planner.go
robholland: I don't think that is pattern used in k8s. I'll add logging for now, and we can add events for this once #28 is done.

internal/planner/planner.go
```diff
@@ -219,9 +228,16 @@ func getScaleDeployments(
 
 // shouldCreateDeployment determines if a new deployment needs to be created
 func shouldCreateDeployment(
+	l logr.Logger,
 	k8sState *k8s.DeploymentState,
 	config *Config,
 ) bool {
+	// Check if we're at the version limit - if so, don't create new deployments
+	if config.Status.VersionCount >= getMaxVersions(config.Spec) {
+		l.Error(errors.New("max versions reached"), "Max versions reached, skipping deployment creation", "versionCount", config.Status.VersionCount, "maxVersions", getMaxVersions(config.Spec))
```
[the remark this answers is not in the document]
robholland: There is no `Warn`, only `Info` and `Error`. Given the choice, I think `Error` is better.

api/v1alpha1/worker_types.go
```diff
@@ -60,6 +70,17 @@ type TemporalWorkerDeploymentSpec struct {
...
 	WorkerOptions WorkerOptions `json:"workerOptions"`
+
+	// MaxVersions defines the maximum number of worker deployment versions allowed.
+	// This helps prevent hitting Temporal's default limit of 100 versions per deployment.
+	// Defaults to 75. Users can override this by explicitly setting a higher value in
+	// the CRD, but should exercise caution: once the 100-version limit is reached,
+	// Temporal attempts to delete an eligible version. If no version is eligible for
+	// deletion, new deployments get blocked which prevents
+	// the controller from making progress.
+
```
jlegrone: Comment needs to be contiguous in order to show up in godoc

```suggestion
//
```

api/v1alpha1/worker_types.go
```diff
@@ -60,6 +70,17 @@ type TemporalWorkerDeploymentSpec struct {
 
 	// TODO(jlegrone): add godoc
 	WorkerOptions WorkerOptions `json:"workerOptions"`
+
+	// MaxVersions defines the maximum number of worker deployment versions allowed.
+	// This helps prevent hitting Temporal's default limit of 100 versions per deployment.
+	// Defaults to 75. Users can override this by explicitly setting a higher value in
+	// the CRD, but should exercise caution: once the 100-version limit is reached,
```
jlegrone: Nit: comment already says before this what the default limit is, so I think we should avoid repeating the number to avoid implying that the limit is definitely 100 when it could actually be configured differently on any given server.

Maybe the comment could also reference the dynamic config value that would need to be checked to verify server config?

```suggestion
	// the CRD, but should exercise caution: once the temporal server limit is reached,
```

internal/planner/planner.go
```diff
@@ -69,19 +70,27 @@ type Config struct {
...
 // DeleteDelay returns the delete delay from the sunset strategy
 func getDeleteDelay(spec *temporaliov1alpha1.TemporalWorkerDeploymentSpec) time.Duration {
 	if spec.SunsetStrategy.DeleteDelay == nil {
-		return 0
+		return temporaliov1alpha1.DefaultDeleteDelay
 	}
 	return spec.SunsetStrategy.DeleteDelay.Duration
 }
 
+// getMaxVersions returns the max versions limit from the spec with safe fallback
```
jlegrone: Do we need this since already setting a default value via webhook?

api/v1alpha1/temporalworker_webhook.go
```diff
@@ -126,6 +130,25 @@ func validateRolloutStrategy(s RolloutStrategy) []*field.Error {
...
+		if *maxVersions < 1 {
+			allErrs = append(allErrs,
+				field.Invalid(field.NewPath("spec.maxVersions"), *maxVersions, "maxVersions must be at least 1"),
+			)
+		}
+		if *maxVersions > DefaultServerMaxVersions {
+			allErrs = append(allErrs,
+				field.Invalid(field.NewPath("spec.maxVersions"), *maxVersions, fmt.Sprintf("maxVersions cannot exceed %d", DefaultServerMaxVersions)),
+			)
+		}
```
jlegrone: Does this mean that max versions can never be >100? I think the controller ought to support a higher limit since the server does as well.

api/v1alpha1/worker_types.go
```diff
@@ -5,10 +5,20 @@
...
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// Default values for TemporalWorkerDeploymentSpec fields
+const (
+	DefaultScaledownDelay    = 1 * time.Hour
+	DefaultDeleteDelay       = 24 * time.Hour
+	DefaultServerMaxVersions = 100
+	DefaultMaxVersions       = int32(DefaultServerMaxVersions * 0.75)
```
jlegrone: Do these to be part of the public API? We could put them in an internal package if they only need to be shared across packages within this repo.

api/v1alpha1/temporalworker_webhook.go
```diff
@@ -126,6 +130,25 @@ func validateRolloutStrategy(s RolloutStrategy) []*field.Error {
 	return allErrs
 }
 
+func validateMaxVersions(maxVersions *int32) []*field.Error {
```
jlegrone: This can be done with declarative validation comments instead, at least for minimum https://book.kubebuilder.io/reference/generating-crd.html#validation

jlegrone (approved):
lgtm pending comments

api/v1alpha1/temporalworker_webhook.go
robholland: I'm tempted to just remove this, I can't think of a sane validation here. We've no idea what the server limit is.

api/v1alpha1/temporalworker_webhook.go
robholland: Nice. Removed the manual validation.

api/v1alpha1/temporalworker_webhook.go
robholland: Removed it.

internal/planner/planner.go
robholland: It was partly due to tests where setting all that stuff is noise (because they are defaults). I've refactored the webhook default stuff so we can call `.Default(..)` from the tests and removed those runtime defaulting calls.

api/v1alpha1/worker_types.go
robholland: Shifted to internal package.

PR #86, opened 2025-07-20
Handle updates to TemporalConnection

### LINKED ISSUE #10: Handle updates to TemporalConnection configuration
The controller should recreate its Temporal client for each connection, and also reconfigure each deployment of each worker using that connection.

Shivs11:
## What was changed
- Handle updates to `TemporalConnection`, for TWD's that are associated with one, so that these updates are propagated deployments.
- This was achieved by adding the hashed contents of a `TemporalConnection` as pod annotations. Thus, if a TemporalConnection were to get updated, it would result in new deployments with the updated annotation. The new updated secret would also be mounted so that the workers starting can be started with the latest secrets.
- Fixes #10 

## Why?

- Better functionality.

## Checklist

1. Closes 

2. How was this tested:

- Added a new unit test for this.
- Going to add a new integration test for this after Carly's PR gets in.

3. Any docs updates needed?

### DIFF
--- a/api/v1alpha1/temporalworker_webhook.go
+++ b/api/v1alpha1/temporalworker_webhook.go
@@ -22,6 +22,7 @@ const (
 	defaultScaledownDelay              = 1 * time.Hour
 	defaultDeleteDelay                 = 24 * time.Hour
 	maxTemporalWorkerDeploymentNameLen = 63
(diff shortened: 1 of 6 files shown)


Shivs11:
closing this as I made a new [PR](https://github.com/temporalio/temporal-worker-controller/pull/102) since resolving all merge conflicts seemed to be a bit challenging

PR #89, opened 2025-07-22
Per version overrides for replicas/sunset strategy

### LINKED ISSUE #48: Version-specific SunsetStrategy override and ReplicaCount override

robholland:
Fixes #48


robholland:
Make sure we set owner on the patches to the TWD so they get cleaned up when the TWD does.

PR #91, opened 2025-07-23
Add migration doc draft.

### LINKED ISSUE #34: Gracefully handle worker deployment resources not created by the controller
What does the controller do in the case where we have a version registered in Temporal but no corresponding k8s deployment object for it? I looked at the code and here's what I found out:

In `genStatus`, we list all deployment versions registered with Temporal and iterate over them to store in the `deploymentVersionsCollection`. Then, we go over the `deployedVersions` list (a list containing all the current k8 deployments in that namespace) and check that if a deployedVersion is present in `deploymentVersionsCollection`, we add it to `deprecatedVersions` so that the worker-controller can take further scaling actions.

I didn't see how, in code, we account for those versions which haven't been created by the controller itself. Maybe this was intended, but I would love to be sure!

_Originally posted by @Shivs11 in https://github.com/DataDog/temporal-worker-controller/pull/22#discussion_r2020990663_

robholland:
Fixes #34

### DIFF
--- a/docs/concepts.md
+++ b/docs/concepts.md
@@ -65,8 +65,7 @@ All Pinned workflows on this version have completed. The version is ready for cl
 Requires explicit human intervention to promote versions. New versions remain in the `Inactive` state until manually promoted.
 
 **Use cases:**
-- During migration from manual deployment systems
-- Testing and validation scenarios
+- Advanced deployment scenarios that are not supported by the other strategies (eg. user wants to do custom testing and validation before making changes to how workflow traffic is routed)
 
 ### AllAtOnce Strategy
 Immediately routes 100% of new workflow executions to the target version once it's healthy and registered.
--- a/docs/migration-guide.md
+++ b/docs/migration-guide.md
@@ -151,15 +151,14 @@ The migration from unversioned to versioned workflows requires careful planning 
 
 #### Phase 2: Initial Migration
 1. **Choose lowest-risk worker** to migrate first
-2. **Create `TemporalWorkerDeployment` custom resource** with Manual strategy
+2. **Create `TemporalWorkerDeployment` custom resource** with a Progressive strategy (conservative intervals recommended)
 3. **Validate controller management** works correctly
 4. **Update deployment pipeline** for this worker
 
 #### Phase 3: Gradual Rollout
 1. **Migrate remaining workers** one at a time
-2. **Enable Progressive rollouts** for validated workers
-3. **Monitor and tune** rollout configurations
-4. **Train team** on new deployment process
+2. **Monitor and tune** rollout configurations
+3. **Train team** on new deployment process
 
 ### Recommended Migration Order
 


docs/migration-guide.md
```diff
@@ -0,0 +1,631 @@
+# Migrating to Temporal Worker Controller
+
+This guide helps teams migrate from their existing versioned worker deployment systems to the Temporal Worker Controller. It assumes you are already running versioned workers using Temporal's Worker Versioning feature and want to automate the management of these deployments.
+
+## Important Terminology
+
+To avoid confusion, this guide uses specific terminology:
+
+- **Temporal Worker Deployment**: A logical grouping in Temporal (e.g., "payment-processor", "notification-sender")
```
Shivs11: nit: can we add the following definitions from the temporal docs so that we stay precise with the terms used here?

referring to:

> A Worker Deployment is a deployment or service across multiple versions. In a rainbow deployment, more than two active Deployment Versions can run at once.
A Worker Deployment Version is a version of a deployment or service. It can have multiple Workers polling on multiple Task Queues, but they all run the same build.
A Build ID, in combination with a Worker Deployment name, identifies a single Worker Deployment Version. 


feel free to disagree, and maybe after this information goes in the outer README this might be redundant we could also just point to that area if we don't wanna repeat ourselves

wdyt

docs/migration-guide.md
```diff
@@ -0,0 +1,631 @@
...
+## Migration Strategy
+
+### The Single Resource Requirement
+
+⚠️ **CRITICAL**: For each **Temporal Worker Deployment**, you MUST create exactly **ONE** `TemporalWorkerDeployment` CRD resource and import all existing versions into it. This is not optional - it's how the system works.
+
+**Terminology Clarification:**
+- **Temporal Worker Deployment**: A logical grouping in Temporal (e.g., "payment-processor")
+- **`TemporalWorkerDeployment` CRD**: One Kubernetes custom resource per Temporal Worker Deployment
+- **Kubernetes `Deployment`**: Multiple k8s Deployments (one per version) managed by the controller
```
Shivs11: doesn't seem like we need this given we have a nice section at the top explaining these things to us

docs/migration-guide.md
```diff
@@ -0,0 +1,631 @@
...
+
+**Key Principles:**
+- **One CRD resource per Temporal Worker Deployment** - Don't create multiple `TemporalWorkerDeployment` resources for the same logical worker
+- **Use Manual strategy during import** - Prevents unwanted automatic promotions
```
Shivs11: just wanted to point out that if we are referring to the different strategies over here, we should probably have them defined in the outer README which, as it stands, is going to be a giant encyclopedia of terms for a new reader haha

docs/migration-guide.md
```diff
@@ -0,0 +1,631 @@
...
+
+### Recommended Migration Steps
+
+1. **Install the controller** in your cluster
+2. **Choose a non-critical worker** to migrate first
+3. **Create a single TemporalWorkerDeployment** with Manual strategy
+4. **Import existing versions one by one** by updating the image spec
+5. **Clean up manual deployments** after confirming controller management
+6. **Enable automated rollouts** by switching to Progressive strategy
+7. **Repeat for remaining workers** one deployment at a time
```
Shivs11: nit: this seems to be the crux of this section and maybe we should leave lines between each point over here; feels to cluttered to eye rn wdyt

docs/migration-guide.md
```diff
@@ -0,0 +1,631 @@
...
+  # IMPORTANT: Use Manual during migration to prevent unwanted promotions
+  cutover:
+    strategy: Manual
+  sunset:
+    scaledownDelay: 30m
+    deleteDelay: 2h
+  template:
+    metadata:
+      annotations:
+        prometheus.io/scrape: "true"
```
Shivs11: maybe we don't need to add this here since it is not super relevant to the migration side of things and seems prom heavy

docs/migration-guide.md
```diff
@@ -0,0 +1,631 @@
...
+   ```bash
+   # Monitor until the new version appears in status.targetVersion
+   kubectl get temporalworkerdeployment payment-processor -o yaml
+   ```
+
+   You should see:
+   ```yaml
+   status:
+     currentVersion:
+       versionID: "payment-processor.v1.5.1"
```
Shivs11: correct me if I am wrong, but this version will only be set as current if it is indeed current for the temporal server, right?

are we assuming it's the current version?

docs/migration-guide.md
```diff
@@ -0,0 +1,631 @@
...
+   ```
+
+3. **Validate workflow routing:**
+   - Start test workflows
+   - Verify they're routed to correct versions
+   - Check Temporal UI for version distribution
+
+## Common Migration Patterns
+
+⚠️ **Remember**: Each pattern below represents **separate Temporal Worker Deployments**. Each gets exactly **one** `TemporalWorkerDeployment` CRD resource.
```
Shivs11: haha this is so cursor-oriented (I also remember you mentioning you did while holding Cursor's hand); do we really need these cautions, feel like while doing a single pass I have come across them multiple times now

docs/migration-guide.md
```diff
@@ -0,0 +1,631 @@
...
+
+### Pattern 1: Microservices with Multiple Workers
```
Shivs11: I really like this section since it helps the reader go through a mental process of "is this for me"

docs/migration-guide.md
```diff
@@ -0,0 +1,631 @@
...
+
+*Symptoms:*
+```
+status:
+  targetVersion:
+    status: NotRegistered
+```
+
+*Solutions:*
+- Check worker logs for connection errors
```
Shivs11: nit: connection/initialization errors

docs/migration-guide.md
```diff
@@ -0,0 +1,631 @@
...
+status:
+  targetVersion:
+    status: Ramping
+    rampPercentage: 5
+```
+
+*Solutions:*
+- Check if gate workflow is configured and completing successfully
+- Verify progressive rollout steps are reasonable
+- Check controller logs for errors
```
Shivs11, line 635: I know there is a section at the bottom about how one can look at the controller logs; we should place a to the section there

docs/migration-guide.md
```diff
@@ -0,0 +1,631 @@
...
+- **One `TemporalWorkerDeployment` CRD per Temporal Worker Deployment**
+- **Manual strategy during migration** to prevent unwanted promotions
+- **Import existing versions sequentially** by updating the same CRD resource
+- **Enable automation only after migration is complete**
+
+**Resource Relationship:**
+- **Before**: Multiple manual Kubernetes `Deployment` resources per worker
+- **After**: One `TemporalWorkerDeployment` CRD → Controller creates multiple Kubernetes `Deployment` resources (one per version)
+
+This approach ensures a smooth transition from manual version management to controller automation without disrupting running workflows.
```
Shivs11: I think we can remove this section on the basis of redundancy

docs/migration-guide.md
```diff
@@ -0,0 +1,631 @@
...
+
+## Next Steps
+
+After successful migration:
+
+1. **Set up monitoring** for your TemporalWorkerDeployment resources
+2. **Update CI/CD pipelines** to patch TemporalWorkerDeployment image specs instead of managing Deployments directly
+3. **Configure alerting** on version transition failures
+4. **Train your team** on the new deployment process (single resource updates vs multiple Deployment management)
+5. **Document your specific configuration** patterns for future reference
```
Shivs11: this also does not feel technical focus and might just be team dependant; not sure if we wanna keep this

Shivs11 (commented):
I did a pass and left some comments

docs/migration-guide.md
Shivs11: btw took the defs out from - https://docs.temporal.io/production-deployment/worker-deployments/worker-versioning

carlydf:
@drewhoskins-temporal could you take a look at this? I think this doc should be designed/evaluated in the context of our other "communication" plans.

@robholland my broad comment is that I think for the Temporal Worker Versioning concepts, we should be linking out to the temporal docs liberally vs redefining things (kind of what Shivam said).

I think this is a good first step though!

docs/migration-guide.md
```diff
@@ -0,0 +1,631 @@
...
+
+Example CI/CD update command:
+```bash
+# Your deployment pipeline should update the CRD image field like this:
+kubectl patch temporalworkerdeployment payment-processor --type='merge' -p='{"spec":{"template":{"spec":{"containers":[{"name":"worker","image":"payment-processor:v1.6.0"}]}}}}'
+```
+
+The controller will automatically create a new Kubernetes `Deployment` (e.g., `payment-processor-v1.6.0`) and manage the rollout.
+
+## Configuration Mapping
```
jlegrone: I think this section could be lifted out of the migration guide into a separate guide on general configuration / options

docs/migration-guide.md
```diff
@@ -0,0 +1,631 @@
+# Migrating to Temporal Worker Controller
+
+This guide helps teams migrate from their existing versioned worker deployment systems to the Temporal Worker Controller. It assumes you are already running versioned workers using Temporal's Worker Versioning feature and want to automate the management of these deployments.
```
drewhoskins-temporal: Hmm, I thought we would prioritize talking about migrating existing _unversioned_ workers to this.  I don't foresee many people in this scenario given that WV is so new.

robholland: Yep, true, re-written from unversioned.

docs/migration-guide.md
robholland: Definitions moved to the concepts file for now.

docs/README.md
```diff
@@ -11,6 +11,9 @@ This documentation structure is designed to support various types of technical d
 
 ## Index
 
+### [Migration Guide](migration-guide.md)
+Comprehensive guide for migrating from existing versioned worker deployment systems to the Temporal Worker Controller. Includes step-by-step instructions, configuration mapping, and common patterns.
```
carlydf:
```suggestion
Comprehensive guide for migrating from existing versioned or unversioned worker deployment systems to the Temporal Worker Controller. Includes step-by-step instructions, configuration mapping, and common patterns.
```
right?

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
+# Temporal Worker Controller Concepts
+
+This document defines key concepts and terminology used throughout the Temporal Worker Controller documentation.
+
+## Core Terminology
+
+### Temporal Worker Deployment
+A logical grouping in Temporal that represents a collection of workers that handle the same set of workflows and activities. Examples include "payment-processor", "notification-sender", or "data-pipeline-worker". This is a concept within Temporal itself, not specific to Kubernetes.
```
carlydf:
```suggestion
A logical grouping in Temporal that represents a collection of workers that are deployed together and should be versioned together. Examples include "payment-processor", "notification-sender", or "data-pipeline-worker". This is a concept within Temporal itself, not specific to Kubernetes.
```

carlydf: soon we will have an index in the public docs that i think it would be good to link to for each term, but for now maybe just a line that says: "See https://docs.temporal.io/production-deployment/worker-deployments/worker-versioning for more details" or something would be good

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+
+**Key characteristics:**
+- Identified by a unique deployment name (e.g., "payment-processor")
```
carlydf:
```suggestion
- Identified by a unique worker deployment name (e.g., "payment-processor/staging")
  -  To ensure uniqueness across Kubernetes namespaces, the controller generates Worker Deployment names with the pattern: `{TemporalWorkerDeployment.Name}/{TemporalWorkerDeployment.Namespace}` (e.g., `payment-processor/staging`)
```
In this repo, I'm trying to always use the full name "worker deployment" instead of "deployment" so that "deployment" can === "kubernetes Deployment"

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+- Can have multiple concurrent versions running simultaneously
```
carlydf:
```suggestion
- Can have multiple concurrent worker versions running simultaneously
```

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+- Versions are identified by Build IDs (e.g., "v1.5.1", "v1.5.2")
```
carlydf:
```suggestion
- Versions of a Worker Deployment are identified by Build IDs (e.g., "v1.5.1", "v1.5.2")
```

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+- Temporal routes workflow executions to appropriate versions based on compatibility rules
```
carlydf:
```suggestion
- Temporal routes workflow executions to appropriate worker versions based on the `RoutingConfig` of the Worker Deployment that the versions are in.
```

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+
+### `TemporalWorkerDeployment` CRD
+The Kubernetes Custom Resource Definition that manages one Temporal Worker Deployment. This is the primary resource you interact with when using the Temporal Worker Controller.
+
+**Key characteristics:**
+- One CRD resource per Temporal Worker Deployment
```
carlydf:
```suggestion
- One `TemporalWorkerDeployment` CRD instance per Temporal Worker Deployment
- Identified by a unique name within its Kubernetes namespace (e.g., "payment-processor")
```
I think "CRD resource" aka "Custom Resource Definition resource" might be akin to saying "Chai Tea"? but maybe "Custom Resource" makes sense too instead of "CRD instance" (those two are synonyms)

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+- Manages the lifecycle of all versions for that deployment
```
carlydf:
```suggestion
- Manages the lifecycle of all versions for that worker deployment
```

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+- Defines rollout strategies, resource requirements, and connection details
+- Controller creates and manages multiple Kubernetes `Deployment` resources based on this spec
+
+### Kubernetes `Deployment`
+The actual Kubernetes Deployment resources that run worker pods. The controller automatically creates these - you don't manage them directly.
```
carlydf:
```suggestion
The actual Kubernetes `Deployment` resources that run worker pods. The controller automatically creates these - you don't manage them directly.
```

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+
+**Key characteristics:**
+- Multiple Kubernetes `Deployment` resources per `TemporalWorkerDeployment` CRD (one per version)
```
carlydf:
```suggestion
- Multiple Kubernetes `Deployment` resources per `TemporalWorkerDeployment` CRD instance (one per worker version)
```

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+- Named with the pattern: `{deployment-name}-{build-id}` (e.g., `payment-processor-v1.5.1`)
```
carlydf:
```suggestion
- Named with the pattern: `{worker-deployment-name}-{build-id}` (e.g., `payment-processor/staging-v1.5.1`)
```

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+- Managed entirely by the controller - created, updated, and deleted automatically
+- Each runs a specific version of your worker code
+
+### Key Relationship
+**One `TemporalWorkerDeployment` CRD → Multiple Kubernetes `Deployment` resources (managed by controller)**
```
carlydf:
```suggestion
**One `TemporalWorkerDeployment` CRD instance → Multiple Kubernetes `Deployment` resources (managed by controller)**
```

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+
+This is the fundamental architecture: you manage a single CRD resource, and the controller handles all the underlying Kubernetes `Deployment` resources for different versions.
```
carlydf:
```suggestion
This is the fundamental architecture: you make changes to the spec of your `TemporalWorkerDeployment` CRD instance / Custom Resource, and the controller handles all the underlying Kubernetes `Deployment` resources for different versions.
```

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Version States
+
+Worker deployment versions progress through various states during their lifecycle:
+
+### NotRegistered
+The version has been specified in the CRD but hasn't been registered with Temporal yet. This typically happens when:
```
carlydf:
```suggestion
The version has been specified in the `TemporalWorkerDeployment` resource but hasn't been registered with Temporal yet. This typically happens when:
```

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+- The worker pods are still starting up
+- There are connectivity issues to Temporal
+- The worker code has errors preventing registration
+
+### Inactive
+The version is registered with Temporal but isn't receiving any new workflow executions. This is the initial state for new versions when using Manual rollout strategy.
```
carlydf:
```suggestion
The version is registered with Temporal but isn't automatically receiving any new workflow executions through the Worker Deployment's `RoutingConfig`. This is the initial state for new versions before they are promoted via Versioning API calls. Inactive versions can receive workflow executions via `VersioningOverride` only.
```

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+
+### Ramping
+The version is receiving a percentage of new workflow executions as part of a Progressive rollout. The percentage gradually increases according to the configured rollout steps.
```
carlydf:
```suggestion
The version is receiving a percentage of new workflow executions. If managed by a Progressive rollout, the percentage gradually increases according to the configured rollout steps. If the rollout is Manual, the user is responsible for setting the ramp percentage and ramping version.
```

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+
+### Current
+The version is receiving 100% of new workflow executions. This is the "production" version that handles all new work.
```
carlydf:
```suggestion
The version is receiving 100% of new workflow executions. This is the "stable" version that handles all new workflows and all existing AutoUpgrade workflows running on the task queues in this Worker Deployment.
```

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+
+### Draining
+The version is no longer receiving new workflow executions but may still be processing existing workflows. The controller waits for all workflows on this version to complete.
```
carlydf:
```suggestion
The version is no longer receiving new workflow executions but may still be processing existing workflows.
```

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+
+### Drained
+All workflows on this version have completed. The version is ready for cleanup according to the sunset configuration.
```
carlydf:
```suggestion
All Pinned workflows on this version have completed. The version is ready for cleanup according to the sunset configuration.
```

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Rollout Strategies
+
+### Manual Strategy
+Requires explicit human intervention to promote versions. New versions remain in the `Inactive` state until manually promoted.
+
+**Use cases:**
+- During migration from manual deployment systems
+- High-risk production environments requiring human approval
+- Testing and validation scenarios
```
carlydf: @drewhoskins-temporal could you weigh in here when you get back? IDK how I feel about recommending Manual mode for "High-risk production environments requiring human approval" I think that might send the wrong message.

Especially because, I think one good idea for "environments requiring human approval" is for the gate workflow to be a human-in-the-loop workflow that sends a notification to a slack channel that someone has to approve (idea for a sample?...)

carlydf: maybe the right message for manual is "environments that desire manual validation of the effects of ramp before promoting to current" or something like that

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+
+### AllAtOnce Strategy
+Immediately routes 100% of new workflow executions to the new version once it's healthy and registered.
```
carlydf:
```suggestion
Immediately routes 100% of new workflow executions to the target version once it's healthy and registered.
```

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+- Non-production environments
+- Low-risk deployments
+- When you want immediate cutover without gradual rollout
+
+### Progressive Strategy
+Gradually increases the percentage of new workflow executions routed to the new version according to configured steps.
+
+**Use cases:**
+- Production deployments where you want to validate new versions gradually
+- When you want automated rollouts with built-in safety checks
```
carlydf, line 83: TODO(carlydf): test whether Gate workflow works with AllAtOnce or Manual mode

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+- Deployments that benefit from canary analysis
+
+## Configuration Concepts
+
+### Worker Options
+Configuration that defines how workers connect to Temporal:
```
carlydf:
```suggestion
Configuration that tells the controller how to connect to the same Temporal cluster and namespace that the worker is connected to:
```
technically the worker does not need the connection; the controller needs the connection so that it can talk to the same environment that the worker is polling / talking to

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+- **connection**: Reference to a `TemporalConnection` resource
+- **temporalNamespace**: The Temporal namespace to connect to
+- **deploymentName**: The logical deployment name in Temporal (auto-generated if not specified)
+
+### Cutover Configuration
```
carlydf:
```suggestion
### Rollout Configuration
```

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+Defines how new versions are promoted:
+- **strategy**: Manual, AllAtOnce, or Progressive
+- **steps**: For Progressive strategy, defines ramp percentages and pause durations
+- **gate**: Optional workflow that must succeed before promotion continues
```
carlydf:
```suggestion
- **gate**: Optional workflow that must succeed on all task queues in the target Worker Deployment Version before promotion continues
```

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+
+### Sunset Configuration
+Defines how old versions are cleaned up:
```
carlydf:
```suggestion
Defines how Drained versions are cleaned up:
```

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+- **scaledownDelay**: How long to wait after draining before scaling pods to zero
+- **deleteDelay**: How long to wait after draining before deleting the Kubernetes `Deployment`
```
carlydf:
```suggestion
- **scaledownDelay**: How long to wait after a version has been Drained before scaling pods to zero
- **deleteDelay**: How long to wait after a version has been Drained before deleting the Kubernetes `Deployment`
```

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+
+### Template
+The pod template used for all versions of this deployment. Similar to a standard Kubernetes Deployment template but managed by the controller.
```
carlydf:
```suggestion
The pod template used for the target version of this worker deployment. Similar to a standard Kubernetes Deployment template but managed by the controller.
```

carlydf: right? or am I confused and perhaps "all versions" was a correct statement and im not understanding the meaning

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+The host and port of the Temporal server, derived from the `TemporalConnection` resource.
+
+### TEMPORAL_NAMESPACE
+The Temporal namespace the worker should connect to, from `spec.workerOptions.temporalNamespace`.
+
+### TEMPORAL_DEPLOYMENT_NAME
+The deployment name in Temporal, either from `spec.workerOptions.deploymentName` or auto-generated from the CRD name.
+
+### WORKER_BUILD_ID
+The build ID for this specific version, derived from the container image tag or explicitly set.
```
carlydf:
```suggestion
The build ID for this specific version, derived from the container image tag and hash of the target pod template.
The worker *must* use this to configure its `worker.DeploymentOptions`.
```

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+The controller automatically sets these environment variables for all worker pods:
+
+### TEMPORAL_HOST_PORT
+The host and port of the Temporal server, derived from the `TemporalConnection` resource.
+
+### TEMPORAL_NAMESPACE
+The Temporal namespace the worker should connect to, from `spec.workerOptions.temporalNamespace`.
+
+### TEMPORAL_DEPLOYMENT_NAME
+The deployment name in Temporal, either from `spec.workerOptions.deploymentName` or auto-generated from the CRD name.
```
carlydf:
```suggestion
The worker deployment name in Temporal, auto-generated from the `TemporalWorkerDeployment` name and Kubernetes namespace.
The worker *must* use this to configure its `worker.DeploymentOptions`.
```

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Environment Variables
+
+The controller automatically sets these environment variables for all worker pods:
+
+### TEMPORAL_HOST_PORT
+The host and port of the Temporal server, derived from the `TemporalConnection` resource.
+
+### TEMPORAL_NAMESPACE
+The Temporal namespace the worker should connect to, from `spec.workerOptions.temporalNamespace`.
```
carlydf, line 117:
```suggestion
The Temporal namespace the worker should connect to, from `spec.workerOptions.temporalNamespace`.
The worker must connect to this Temporal namespace, but since this is user provided and not controller generated, the user does not necessarily need to access this env var to get that namespace if it already knows the namespace another way.
```

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+
+### Template
+The pod template used for all versions of this deployment. Similar to a standard Kubernetes Deployment template but managed by the controller.
+
+## Environment Variables
+
+The controller automatically sets these environment variables for all worker pods:
+
+### TEMPORAL_HOST_PORT
+The host and port of the Temporal server, derived from the `TemporalConnection` resource.
```
carlydf:
```suggestion
The host and port of the Temporal server, derived from the `TemporalConnection` resource.
The worker must connect to this Temporal endpoint, but since this is user provided and not controller generated, the user does not necessarily need to access this env var to get that endpoint if it already knows the endpoint another way.
```

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+### TEMPORAL_DEPLOYMENT_NAME
+The deployment name in Temporal, either from `spec.workerOptions.deploymentName` or auto-generated from the CRD name.
+
+### WORKER_BUILD_ID
+The build ID for this specific version, derived from the container image tag or explicitly set.
+
+## Resource Management Concepts
+
+### Rainbow Deployments
+The pattern of running multiple versions of the same worker simultaneously. This is essential for maintaining workflow determinism in Temporal, as running workflows must continue executing on the version they started with.
```
carlydf:
```suggestion
The pattern of running multiple versions of the same service simultaneously. Running multiple versions of your workers simultaneously is essential for supporting Pinned workflows in Temporal, as Pinned workflows must continue executing on the worker version they started on.
```

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+
+### Version Lifecycle Management
+The automated process of:
+1. Registering new versions with Temporal
+2. Gradually routing traffic to new versions
+3. Draining old versions once they're no longer needed
+4. Cleaning up resources for drained versions
```
carlydf:
```suggestion
3. Cleaning up resources for drained versions
```
I think this can actually be removed, because the controller doesn't do anything to make a version Drain, it just tracks to see when a version is Drained.

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+
+### Controller-Managed Resources
+Resources that are created, updated, and deleted automatically by the controller:
+- Kubernetes `Deployment` resources for each version
+- ConfigMaps and Secrets as needed
+- Service accounts and RBAC resources
```
carlydf: I don't think the controller writes to any config maps or secrets -- it reads from the secret specified in the TemporalConnection but it never writes to it

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+- Labels and annotations for tracking and management
+
+## Migration Concepts
+
+### Import Process
+The process of bringing existing manually-managed worker deployments under controller management. This involves:
+1. Creating a `TemporalWorkerDeployment` CRD with Manual strategy
```
carlydf:
```suggestion
1. Creating a `TemporalWorkerDeployment` custom resource with Manual strategy
```

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+2. Sequentially updating the image spec to register each existing version
```
carlydf:
```suggestion
2. Sequentially updating the target pod template in the `TemporalWorkerDeployment` spec to prompt the controller to create a Kubernetes Deployment with that pod spec that is owned and tracked by the controller. Do this for each of your existing Deployments.
```

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+Resources that are created, updated, and deleted automatically by the controller:
+- Kubernetes `Deployment` resources for each version
+- ConfigMaps and Secrets as needed
+- Service accounts and RBAC resources
+- Labels and annotations for tracking and management
+
+## Migration Concepts
+
+### Import Process
+The process of bringing existing manually-managed worker deployments under controller management. This involves:
```
carlydf: should we specify here that these instructions are for versioned worker deployments? The controller is never going to scale down an unversioned worker deployment, because it is not registered (or at least, it shouldnt, but I should test that)

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+1. Creating a `TemporalWorkerDeployment` CRD with Manual strategy
+2. Sequentially updating the image spec to register each existing version
+3. Cleaning up original manual Kubernetes `Deployment` resources
```
carlydf:
```suggestion
3. Cleaning up original non-worker-controller-managed Kubernetes `Deployment` resources
```

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+4. Enabling automated rollouts
+
+### Single Resource Requirement
+The critical principle that each Temporal Worker Deployment must be managed by exactly one `TemporalWorkerDeployment` CRD resource. You cannot split a single logical deployment across multiple CRD resources.
```
carlydf:
```suggestion
The critical principle is that each Temporal Worker Deployment must be managed by exactly one `TemporalWorkerDeployment` CRD instance. You cannot split a single worker deployment across multiple `TemporalWorkerDeployment` instances.
```

carlydf: suggested edit, but tbh I don't think this section is helpful, and think it might be better to remove. We've already said the one-to-one thing before

docs/concepts.md
```diff
@@ -0,0 +1,160 @@
...
+
+### Legacy Version Handling
+The process of ensuring that existing worker versions continue running during migration, maintaining workflow determinism while transitioning to controller management.
```
carlydf: how this is presented feels very "AI" to me. It's fine but if you can think of a way to reformat the "## Migration Concepts" section so that this point about legacy version handling is just a normal sentence that adds more color to "Import process" I think that would be nice

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
+# Migrating from Unversioned to Versioned Workflows with Temporal Worker Controller
```
carlydf, line 1: I'm not sure if we want to be saying "versioned workers" or "versioned workflows"

I'm leaning "versioned workers" since that's a term we use in a lot of other places

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+
+This guide helps teams migrate from unversioned Temporal workflows to versioned workflows using the Temporal Worker Controller. It assumes you are currently running workers without Temporal's Worker Versioning feature and want to adopt versioned deployments for safer, more controlled rollouts.
```
carlydf:
```suggestion
This guide helps teams migrate from unversioned Temporal workflows to versioned workflows using the Temporal Worker Controller. It assumes you are currently running workers without Temporal's Worker Versioning feature and want to adopt versioned worker deployments for safer, more controlled rollouts.
```

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+- ✅ **Basic worker configuration**: Workers connect to Temporal with namespace and task queue configuration
+- ✅ **Administrative access**: Ability to install Custom Resource Definitions and controllers
+- ✅ **Deployment pipeline**: Existing CI/CD system that can be updated to use new deployment method
+
+### Current Environment Variables
+
+Your workers are likely configured with basic environment variables like:
+
+```bash
+TEMPORAL_HOST_PORT=your-namespace.tmprl.cloud:7233
```
carlydf:
```suggestion
TEMPORAL_HOST_PORT=your-temporal-namespace.tmprl.cloud:7233
```

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+    spec:
+      containers:
+      - name: worker
+        image: my-worker:v1.2.3
+        env:
+        - name: TEMPORAL_HOST_PORT
+          value: "production.tmprl.cloud:7233"
+        - name: TEMPORAL_NAMESPACE
+          value: "production"
+        # No versioning environment variables
```
carlydf, line 75: this example has no mTLS, which is fine, but do we want to demonstrate that at some point (maybe future PR) to show that aspect of the `TemporalConnection` resource?

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+**Current Deployment Process:**
+1. Build new worker image
+2. Update Deployment with new image
+3. Kubernetes rolls out new pods, terminating old ones
+4. **Risk**: Running workflows may fail if code changes break compatibility
+
+### After: Versioned Workers with Controller
+
+```yaml
+# Single CRD manages multiple versions automatically
```
carlydf:
```suggestion
# Single Custom Resource manages multiple versions of a worker deployment automatically
```

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+      - rampPercentage: 50
+        pauseDuration: 10m
+  sunset:
+    scaledownDelay: 1h
+    deleteDelay: 24h
+  template:
+    spec:
+      containers:
+      - name: worker
+        image: my-worker:v1.2.4  # Update this to deploy new versions
```
carlydf: maybe this comment or another comment could incorporate some thing like "Updating any part of the pod spec will generate a new Target Build ID for the worker deployment"

carlydf: I just want people to know that too

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+        # Note: Controller automatically adds versioning environment variables:
+        # TEMPORAL_HOST_PORT, TEMPORAL_NAMESPACE, TEMPORAL_DEPLOYMENT_NAME, WORKER_BUILD_ID
+```
+
+**New Deployment Process:**
+1. Build new worker image  
+2. Update `TemporalWorkerDeployment` CRD with new image
```
carlydf:
```suggestion
2. Update `TemporalWorkerDeployment` CRD instance with new image
```

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+3. Controller creates new Kubernetes `Deployment` for the new version
+4. Controller gradually routes new workflows to new version
+5. Old version continues handling existing workflows until they complete
```
carlydf:
```suggestion
5. Old version continues handling existing Pinned workflows until they complete
```

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+        image: my-worker:v1.2.4  # Update this to deploy new versions
+        # Note: Controller automatically adds versioning environment variables:
+        # TEMPORAL_HOST_PORT, TEMPORAL_NAMESPACE, TEMPORAL_DEPLOYMENT_NAME, WORKER_BUILD_ID
+```
+
+**New Deployment Process:**
+1. Build new worker image  
+2. Update `TemporalWorkerDeployment` CRD with new image
+3. Controller creates new Kubernetes `Deployment` for the new version
+4. Controller gradually routes new workflows to new version
```
carlydf:
```suggestion
4. Controller gradually routes new workflows and existing AutoUpgrade workflows to new version
```

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+
+#### Phase 1: Preparation
+1. **Install the controller** in non-production environments
+2. **Update worker code** to support versioning (if needed)
+3. **Test migration process** with non-critical workers
+4. **Prepare CI/CD pipeline changes** for new deployment method
+
+#### Phase 2: Initial Migration
+1. **Choose lowest-risk worker** to migrate first
+2. **Create `TemporalWorkerDeployment` CRD** with Manual strategy
```
carlydf:
```suggestion
2. **Create `TemporalWorkerDeployment` CRD instance** with Manual strategy
```

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+3. **Validate controller management** works correctly
+4. **Update deployment pipeline** for this worker
+
+#### Phase 3: Gradual Rollout
+1. **Migrate remaining workers** one at a time
+2. **Enable Progressive rollouts** for validated workers
+3. **Monitor and tune** rollout configurations
+4. **Train team** on new deployment process
+
+### Recommended Migration Order
```
carlydf, line 163: i like this! I think its helpful for users to know they don't have to swap over all at once

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+- ✅ **Workflow continuity** - Deterministic execution preserved across deployments
+
+## Migration Strategy
+
+### Safe Migration Approach
+
+The migration from unversioned to versioned workflows requires careful planning to avoid disrupting running workflows. The key is to transition gradually while maintaining workflow continuity.
+
+**Key Principles:**
+- **Start with Manual strategy** - Prevents automatic promotions during initial setup
```
carlydf: I think we want folks to try out the thing that they will want to do long-term. So ideally they should try out Progressive, but perhaps with a more conservative rollout policy than they might do later on.

If people start with Manual, they'll have to do "wasted" work developing a system to call the upgrade APIs with the CLI or SDK, which is work that won't be re-used in production, and prone to errors.

Idk that's just my take. I think in an ephemeral environment, folks can happily play with Manual mode all they want, but my recomendation for migrating brownfield environment would be to /add/ a TWD while still deploying the unversioned pollers.

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+- **Migrate one worker deployment at a time** - Reduces risk and allows learning
+- **Test thoroughly in non-production** - Validate the approach before production migration
+- **Preserve running workflows** - Ensure in-flight workflows complete successfully
+- **Enable automation gradually** - Move to Progressive rollouts only after validation
+
+### Migration Phases
+
+#### Phase 1: Preparation
+1. **Install the controller** in non-production environments
+2. **Update worker code** to support versioning (if needed)
```
carlydf: the "if needed" feels weird here given that we said that the workers were unversioned as a requirement

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+kind: TemporalConnection
+metadata:
+  name: staging-temporal
+  namespace: default
+spec:
+  hostPort: "staging.abc123.tmprl.cloud:7233"
+  mutualTLSSecret: temporal-cloud-mtls
+```
+
+### Step 3: Prepare Your Worker Code (If Needed)
```
carlydf: same re: "if needed"

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+
+Most workers will work without changes, but you may need to update your worker initialization code to properly handle versioning:
```
carlydf: if they are unversioned, they won't work without changes

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+
+**Before (Unversioned):**
+```go
+// Worker connects without versioning
+worker := worker.New(client, "my-task-queue", worker.Options{})
+```
+
+**After (Versioned - Optional Enhancement):**
+```go
+// Worker can optionally use build ID from environment
```
carlydf: worker MUST use build id and deployment name from environment

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+buildID := os.Getenv("WORKER_BUILD_ID")
+workerOptions := worker.Options{}
+if buildID != "" {
+    workerOptions.BuildID = buildID
+    workerOptions.UseBuildIDForVersioning = true
+}
+worker := worker.New(client, "my-task-queue", workerOptions)
+```
+
+> **Note**: The controller automatically sets `WORKER_BUILD_ID` environment variable, so most workers will work without code changes.
```
carlydf: maybe this is the write place to say that WORKER_BUILD_ID and WORKER_DEPLOYMENT_NAME (double check those) are generated by the controller and must be passed to the worker's worker.Options.DeploymentOptions

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+```
+
+**After (Versioned - Optional Enhancement):**
+```go
+// Worker can optionally use build ID from environment
+buildID := os.Getenv("WORKER_BUILD_ID")
+workerOptions := worker.Options{}
+if buildID != "" {
+    workerOptions.BuildID = buildID
+    workerOptions.UseBuildIDForVersioning = true
```
carlydf: I think this is wrong too and it's actually worker.Options.DeploymentOptions{BuildId, DeploymentName}, but double check the docs or the demo worker code

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+}
+worker := worker.New(client, "my-task-queue", workerOptions)
+```
+
+> **Note**: The controller automatically sets `WORKER_BUILD_ID` environment variable, so most workers will work without code changes.
+
+### Step 4: Create Your First TemporalWorkerDeployment
+
+Start with your lowest-risk worker. Convert your existing unversioned Deployment to a `TemporalWorkerDeployment` CRD:
```
carlydf:
```suggestion
Start with your lowest-risk worker. Convert your existing unversioned Deployment to a `TemporalWorkerDeployment` CRD instance:
```
I think we want to tell people to keep their unversioned workers around, so maybe "Make a copy of your existing unversioned Deployment and convert it to a ..." would be extra clear that we don't recommend to replace them?

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+
+**Current Unversioned Deployment:**
```
carlydf:
```suggestion
**Existing Unversioned Deployment:**
```
just because the term "Current" is so overloaded in the versioning world? wdyt?

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+      - name: worker
+        image: payment-processor:v1.5.2
+        env:
+        - name: TEMPORAL_HOST_PORT
+          value: "production.tmprl.cloud:7233"
+        - name: TEMPORAL_NAMESPACE
+          value: "production"
+```
+
+**New TemporalWorkerDeployment CRD (IMPORTANT: Use Manual strategy initially):**
```
carlydf:
```suggestion
**New TemporalWorkerDeployment resource:**
```
or "instance"?

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+metadata:
+  name: payment-processor
+  labels:
+    app: payment-processor
+spec:
+  replicas: 3
+  workerOptions:
+    connection: production-temporal
+    temporalNamespace: production
+  # CRITICAL: Use Manual during initial migration
```
carlydf: idk about this, lets discuss monday?

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+  cutover:
```
carlydf:
```suggestion
  rollout:
```

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+    strategy: Manual
+  sunset:
+    scaledownDelay: 30m
+    deleteDelay: 2h
+  template:
+    spec:
+      containers:
+      - name: worker
+        image: payment-processor:v1.5.2  # Same image as current deployment
```
carlydf:
```suggestion
        image: payment-processor:v1.5.2  # Same image as current deployment to ensure no breaking changes
```

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+        resources:
+          requests:
+            memory: "512Mi"
+            cpu: "250m"
+        # Note: Controller automatically adds TEMPORAL_* env vars
+```
+
+### Step 5: Deploy the TemporalWorkerDeployment
+
+1. **Create the `TemporalWorkerDeployment` CRD:**
```
carlydf:
```suggestion
1. **Create the `TemporalWorkerDeployment` custom resource:**
```
or CRD instance, or resource

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+   ```bash
+   kubectl apply -f payment-processor-versioned.yaml
+   ```
+
+2. **Wait for the version to be registered:**
+   ```bash
+   # Monitor until status shows the version is registered and current
```
carlydf, line 306: just a note that if mode is manual, version won't become current. but if mode is progressive (which imo it should be) then it will.

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+   ```bash
+   # Monitor running workflows in Temporal UI
+   # Or use Temporal CLI to check workflow status
+   temporal workflow list --namespace production
+   ```
+
+4. **Scale down the original unversioned deployment:**
+   ```bash
+   # Gradually reduce replicas of the original deployment
+   kubectl scale deployment payment-processor --replicas=1
```
carlydf: what's the point of scaling to 1?

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+   # Only after confirming all workflows are handled by versioned workers
+   kubectl delete deployment payment-processor
+   ```
+
+### Step 7: Enable Automated Rollouts
+
+Once the initial migration is complete and validated, enable automated rollouts for future deployments:
+
+```bash
+# Update the TemporalWorkerDeployment CRD to use Progressive strategy
```
carlydf:
```suggestion
# Update the TemporalWorkerDeployment custom resource to use Progressive strategy
```

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+
+**Before (Unversioned):**
+```bash
+# Old pipeline updated Deployment directly
+kubectl set image deployment/payment-processor worker=payment-processor:v1.6.0
+```
+
+**After (Versioned):**
+```bash
+# New pipeline updates TemporalWorkerDeployment CRD
```
carlydf:
```suggestion
# New pipeline updates TemporalWorkerDeployment custom resource
```

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+
+## Configuration Reference
+
+### Rollout Strategies
+
+See the [Concepts](concepts.md) document for detailed explanations of rollout strategies. Here are the basic configuration patterns:
+
+**Manual Strategy (Default Behavior):**
+```yaml
+cutover:
```
carlydf: replace cutover with rollout for all of these

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+```
+
+**Pod Annotations and Labels:**
+```yaml
+template:
+  metadata:
+    labels:
+      team: payments
+      environment: production
+```
```
carlydf: I feel like this is out of scope, people should read k8s docs on pod spec for this

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+   - Verify they're routed to correct versions
+   - Check Temporal UI for version distribution
+
+## Common Migration Patterns
+
+Here are common scenarios when migrating from unversioned to versioned workflows:
+
+### Pattern 1: Microservices Architecture
+
+If you have multiple services with their own workers, migrate each service independently:
```
carlydf: maybe something like "and each service is versioned and patched separately"?

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+kind: TemporalWorkerDeployment
+metadata:
+  name: payment-processor
+  namespace: payments
+spec:
+  workerOptions:
+    connection: production-temporal
+    temporalNamespace: payments
+  cutover:
+    strategy: Progressive  # Conservative rollout for financial operations
```
carlydf:
```suggestion
    strategy: Progressive
```
we rec Progressive regardless, so no need to mention "conservative" imo

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+kind: TemporalWorkerDeployment
+metadata:
+  name: notification-sender
+  namespace: notifications
+spec:
+  workerOptions:
+    connection: production-temporal
+    temporalNamespace: notifications
+  cutover:
+    strategy: AllAtOnce  # Lower risk, faster rollouts acceptable
```
carlydf: maybe "desired" instead of acceptable? IMO if people don't know what they want they should do progressive, so "acceptable" is not a good enough reason to do AllAtOnce

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+kind: TemporalWorkerDeployment
+metadata:
+  name: order-processor
+  namespace: staging
+spec:
+  workerOptions:
+    connection: staging-temporal
+    temporalNamespace: staging
+  cutover:
+    strategy: AllAtOnce  # Immediate cutover for faster iteration
```
carlydf: maybe just "faster rollout"? I feel like iteration speed is dependent on so many factors, least of which is how long it takes to deploy one service to one env

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+  targetVersion:
+    status: NotRegistered
+```
+
+*Solutions:*
+- Check worker logs for connection errors
+- Verify TemporalConnection configuration  
+- Ensure TLS secrets are properly configured
+- Verify network connectivity to Temporal server
+- Check that worker code properly handles `WORKER_BUILD_ID` environment variable
```
carlydf: and WORKER_DEPLOYMENT_NAME

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+
+**2. New Workflows Still Going to Unversioned Workers**
+
+*Symptoms:*
+- Temporal UI shows workflows executing on unversioned workers
+- Versioned workers appear idle
```
carlydf, line 602: not sure what "appear idle" would look like, but the Temporal UI shows workflows *not* executing on versioned workers (you can see this by clicking "workflows" from the version line item in the UI)

docs/migration-guide.md
```diff
@@ -0,0 +1,732 @@
...
+
+*Solutions:*
+- Verify versioned workers are properly registered in Temporal UI
+- Check that new workflow starts are using the correct task queue
```
carlydf:
```suggestion
- Check that new workflows are starting on the correct task queue
```

docs/README.md
robholland: It doesn't actually cover migrating from versioned systems at all. Updated.

docs/concepts.md
```diff
@@ -0,0 +1,164 @@
+# Temporal Worker Controller Concepts
+
+This document defines key concepts and terminology used throughout the Temporal Worker Controller documentation.
+
+## Core Terminology
```
drewhoskins-temporal, line 5: Consider linking out to https://docs.temporal.io/worker-versioning

docs/concepts.md
robholland: For now I'll just delete "High-risk..." and we can figure out a better wording offline.

docs/concepts.md
robholland: Removed that, and added that we write to `TemporalWorkerDeployment` custom resources to update their status.

docs/concepts.md
robholland: Removed.

docs/concepts.md
robholland: Removed this migration process as it's not a concept. The process is out lined in the migration guide.

docs/migration-guide.md
robholland: Reworded, see what you think.

docs/concepts.md
```diff
@@ -0,0 +1,147 @@
...
+The Kubernetes Custom Resource Definition that manages one Temporal Worker Deployment. This is the primary resource you interact with when using the Temporal Worker Controller.
+
+**Key characteristics:**
+- One `TemporalWorkerDeployment` Custom Resource per Temporal Worker Deployment
+- Manages the lifecycle of all versions for that worker deployment
+- Defines rollout strategies, resource requirements, and connection details
+- Controller creates and manages multiple Kubernetes `Deployment` resources based on this spec
+
+The actual Kubernetes `Deployment` resources that run worker pods. The controller automatically creates these - you don't manage them directly.
+The actual Kubernetes Deployment resources that run worker pods. The controller automatically creates these - you don't manage them directly.
```
carlydf: duplicate line

docs/concepts.md
```diff
@@ -0,0 +1,147 @@
...
+
+**Key characteristics:**
+- Multiple Kubernetes `Deployment` resources per `TemporalWorkerDeployment` Custom Resource (one per version)
+- Named with the pattern: `{worker-deployment-name}-{build-id}` (e.g., `payment-processor/staging-v1.5.1`)
```
carlydf:
```suggestion
- Named with the pattern: `{worker-deployment-name}-{build-id}` (e.g., `staging/payment-processor-v1.5.1`)
```
I changed naming format to ns/twd-name since this doc was started!

docs/concepts.md
```diff
@@ -0,0 +1,147 @@
...
+- Managed entirely by the controller - created, updated, and deleted automatically
+- Each runs a specific version of your worker code
+
+### Key Relationship
+**One `TemporalWorkerDeployment` Custom Resource → Multiple Kubernetes `Deployment` resources (managed by controller)**
+
+make changes to the spec of your `TemporalWorkerDeployment` Custom Resource, and the controller handles all the underlying Kubernetes `Deployment` resources for different versions.
```
carlydf: should Make be capitalized as the start of a sentence?

docs/concepts.md
```diff
@@ -0,0 +1,147 @@
...
+- The worker code has errors preventing registration
+
+### Inactive
+The version is registered with Temporal but isn't automatically receiving any new workflow executions through the Worker Deployment's `RoutingConfig`. This is the initial state for new versions before they are promoted via Versioning API calls. Inactive versions can receive workflow executions via `VersioningOverride` only.
+
+### Ramping
+The version is receiving a percentage of new workflow executions. If managed by a Progressive rollout, the percentage gradually increases according to the configured rollout steps. If the rollout is Manual, the user is responsible for setting the ramp percentage and ramping version.
+
+### Current
+The version is receiving 100% of new workflow executions. This is the "stable" version that handles all new workflows and all existing AutoUpgrade workflows running on the task queues in this Worker Deployment.
```
carlydf: Really it's (100-RampPercentage)% of new workflow executions.
Here's the language we use in the SDK, if that's helpful https://pkg.go.dev/go.temporal.io/sdk@v1.36.0/internal#WorkerDeploymentRoutingConfig

In the [API](https://github.com/temporalio/api/blob/master/temporal/api/deployment/v1/message.proto#L267C1-L268C1), we say "When ramp percentage is non-zero, that portion of traffic is shifted from the Current Version to the Ramping Version" which I think is a relatively concise way to put it.

docs/concepts.md
```diff
@@ -0,0 +1,147 @@
...
+- **steps**: For Progressive strategy, defines ramp percentages and pause durations
+- **gate**: Optional workflow that must succeed on all task queues in the target Worker Deployment Version before promotion continues
+
+### Sunset Configuration
+Defines how Drained versions are cleaned up:
+- **scaledownDelay**: How long to wait after a version has been Drained before scaling pods to zero
+- **deleteDelay**: How long to wait after a version has been Drained before deleting the Kubernetes `Deployment`
+
+### Template
+The pod template used for the target version of this worker deployment. Similar to a standard Kubernetes Deployment template but managed by the controller.
```
carlydf: I think it might be more correct to say "Similar to the pod template in a standard Kubernetes Deployment, but managed by the controller."

docs/concepts.md
```diff
@@ -0,0 +1,147 @@
...
+
+## Environment Variables
+
+The controller automatically sets these environment variables for all worker pods:
+
+### TEMPORAL_HOST_PORT
```
carlydf:
```suggestion
### TEMPORAL_ADDRESS
```
changed to conform with SDK expected values

docs/concepts.md
```diff
@@ -0,0 +1,147 @@
...
+
+### TEMPORAL_NAMESPACE
+The Temporal namespace the worker should connect to, from `spec.workerOptions.temporalNamespace`.
+The worker must connect to this Temporal namespace, but since this is user provided and not controller generated, the user does not necessarily need to access this env var to get that namespace if it already knows the namespace another way.
+
+### TEMPORAL_DEPLOYMENT_NAME
+The worker deployment name in Temporal, auto-generated from the `TemporalWorkerDeployment` name and Kubernetes namespace.
+The worker *must* use this to configure its `worker.DeploymentOptions`.
+
+### WORKER_BUILD_ID
```
carlydf:
```suggestion
### TEMPORAL_WORKER_BUILD_ID
```
also changed recently

docs/concepts.md
```diff
@@ -0,0 +1,147 @@
...
+The build ID for this specific version, derived from the container image tag and hash of the target pod template.
+The worker *must* use this to configure its `worker.DeploymentOptions`.
```
carlydf, line 126: We could tell them to see our demo worker here for an example of how to connect a versioned worker: https://github.com/temporalio/temporal-worker-controller/blob/071789e70726ccd88e95c83eaa77536daad6e0a9/internal/demo/util/client.go#L37

docs/migration-guide.md
```diff
@@ -0,0 +1,708 @@
...
+- **Deployment Risk**: Code changes can break running workflows if they're not backward compatible
+- **Rollback Complexity**: Rolling back deployments can disrupt in-flight workflows
+- **Workflow Determinism Issues**: Changes to workflow logic can cause non-deterministic errors
+
+Versioned workflows with the Temporal Worker Controller solve these problems by:
+
+- ✅ **Safe Deployments**: New versions run alongside old ones, ensuring running workflows complete successfully
+- ✅ **Automated Rollouts**: Progressive rollout strategies reduce risk of new deployments
+- ✅ **Easy Rollbacks**: Can instantly route new workflows back to previous versions
+- ✅ **Workflow Continuity**: Running workflows can continue on their original version until completion
```
carlydf: would it make sense to link out to https://docs.temporal.io/production-deployment/worker-deployments/worker-versioning for the benefits (and if the docs are missing some of the things mentioned here, move that to the docs) instead of rehashing it here?

carlydf: The first section of that doc gives a good overview of the benefits, and I feel like we don't need to repeat ourselves

docs/migration-guide.md
```diff
@@ -0,0 +1,708 @@
...
+- ✅ **Basic worker configuration**: Workers connect to Temporal with namespace and task queue configuration
+- ✅ **Administrative access**: Ability to install Custom Resource Definitions and controllers
+- ✅ **Deployment pipeline**: Existing CI/CD system that can be updated to use new deployment method
+
+### Current Environment Variables
+
+Your workers are likely configured with basic environment variables like:
+
+```bash
+TEMPORAL_HOST_PORT=your-temporal-namespace.tmprl.cloud:7233
```
carlydf:
```suggestion
TEMPORAL_ADDRESS=your-temporal-namespace.tmprl.cloud:7233
```
changed recently to match SDK expectations!

docs/migration-guide.md
```diff
@@ -0,0 +1,708 @@
...
+  name: my-worker
+spec:
+  replicas: 3
+  template:
+    spec:
+      containers:
+      - name: worker
+        image: my-worker:v1.2.3
+        env:
+        - name: TEMPORAL_HOST_PORT
```
carlydf:
```suggestion
        - name: TEMPORAL_ADDRESS
```

docs/migration-guide.md
```diff
@@ -0,0 +1,708 @@
...
+  sunset:
+    scaledownDelay: 1h
+    deleteDelay: 24h
+  template:
+    spec: # Any changes to this spec will trigger the controller to deploy a new version.
+      containers:
+      - name: worker
+        image: my-worker:v1.2.4  # This is the most common value to change, as you roll out a new worker image.
+        # Note: Controller automatically adds versioning environment variables:
+        # TEMPORAL_HOST_PORT, TEMPORAL_NAMESPACE, TEMPORAL_DEPLOYMENT_NAME, TEMPORAL_WORKER_BUILD_ID
```
carlydf:
```suggestion
        # TEMPORAL_ADDRESS, TEMPORAL_NAMESPACE, TEMPORAL_DEPLOYMENT_NAME, TEMPORAL_WORKER_BUILD_ID
```

docs/migration-guide.md
carlydf: just bumping this, I think the "start with manual strategy" recommendation should be removed! but happy to hear your reasoning if you disagree

docs/migration-guide.md
```diff
@@ -0,0 +1,708 @@
...
+**After (Versioned):**
+```go
+// Worker must use the build ID/deployment name from environment
+// These are set on the deployment by the controller
+buildID := os.Getenv("TEMPORAL_WORKER_BUILD_ID")
+deploymentName := os.Getenv("TEMPORAL_DEPLOYMENT_NAME")
+if buildID == "" || deploymentName == "" {
+  // exit with an error
+}
+workerOptions := worker.Options{}
```
carlydf, line 224: maybe instead of starting with empty options, we should recommend that users start with the SDK-loaded default options so that the worker has the right connection info
https://github.com/temporalio/temporal-worker-controller/blob/071789e70726ccd88e95c83eaa77536daad6e0a9/internal/demo/util/client.go#L37

docs/migration-guide.md
robholland: Adjusted to progressive.

docs/concepts.md
robholland: The client code isn't really relevant to configuring their worker for versioning.

robholland: Honestly I think there are too many helpers/abstractions in the demo code for it be useful as an quick to understand example.

docs/concepts.md
robholland: Re-worded.

docs/migration-guide.md
robholland: That's client options, it's not related to worker options.

docs/concepts.md
robholland: I'll do this in a future PR as Alex and I re-balance content between the repo and the docs site.

docs/migration-guide.md
robholland: Added envconfig env vars, I think this is all that's needed?

carlydf: yes that's all!

docs/concepts.md
```diff
@@ -0,0 +1,144 @@
...
+All Pinned workflows on this version have completed. The version is ready for cleanup according to the sunset configuration.
+
+## Rollout Strategies
+
+### Manual Strategy
+Requires explicit human intervention to promote versions. New versions remain in the `Inactive` state until manually promoted.
+
+**Use cases:**
+- During migration from manual deployment systems
+- Testing and validation scenarios
```
carlydf:
```suggestion
- Advanced deployment scenarios that are not supported by the other strategies (eg. user wants to do custom testing and validation before making changes to how workflow traffic is routed)
```

docs/migration-guide.md
```diff
@@ -0,0 +1,678 @@
...
+
+#### Phase 1: Preparation
+1. **Install the controller** in non-production environments
+2. **Update worker code** to support versioning
+3. **Test migration process** with non-critical workers
+4. **Prepare CI/CD pipeline changes** for new deployment method
+
+#### Phase 2: Initial Migration
+1. **Choose lowest-risk worker** to migrate first
+2. **Create `TemporalWorkerDeployment` custom resource** with Manual strategy
```
carlydf:
```suggestion
2. **Create `TemporalWorkerDeployment` custom resource** with a Progressive strategy (conservative intervals recommended)
```

docs/migration-guide.md
```diff
@@ -0,0 +1,678 @@
...
+3. **Validate controller management** works correctly
+4. **Update deployment pipeline** for this worker
+
+#### Phase 3: Gradual Rollout
+1. **Migrate remaining workers** one at a time
+2. **Enable Progressive rollouts** for validated workers
+3. **Monitor and tune** rollout configurations
+4. **Train team** on new deployment process
```
carlydf:
```suggestion
1. **Migrate remaining workers** one at a time
2. **Monitor and tune** rollout configurations
3. **Train team** on new deployment process
```

PR #92, opened 2025-07-23
general bug fixes

Shivs11:
## What was changed
- general bug fixes that have somehow sneaked in `main`
- WIP

## Why?
- because bugs are not good

## Checklist

1. Closes 

2. How was this tested:
- unit-testing + testing the controller locally.

3. Any docs updates needed?

### DIFF
--- a/internal/controller/state_mapper.go
+++ b/internal/controller/state_mapper.go
@@ -5,6 +5,8 @@
 package controller
 
 import (
+	"fmt"
+
 	"github.com/temporalio/temporal-worker-controller/api/v1alpha1"
 	"github.com/temporalio/temporal-worker-controller/internal/k8s"
 	"github.com/temporalio/temporal-worker-controller/internal/temporal"
@@ -39,7 +41,9 @@ func (m *stateMapper) mapToStatus(targetVersionID string) *v1alpha1.TemporalWork
 	status.TargetVersion = m.mapTargetWorkerDeploymentVersion(targetVersionID)
 	if m.temporalState.RampingVersionID == targetVersionID {
 		status.TargetVersion.RampingSince = m.temporalState.RampingSince
-		status.TargetVersion.RampLastModifiedAt = m.temporalState.RampLastModifiedAt
+		fmt.Println("Ramping since is", m.temporalState.RampingSince)
+		fmt.Println("Ramp last modified at is", m.temporalState.RampLastModifiedAt)
+		status.TargetVersion.RampLastModifiedAt = m.temporalState.RampLastModifiedAt // This is showing to be an outdated value of Ramp last modified at is 0001-01-01 00:00:00 +0000 UTC
 		rampPercentage := m.temporalState.RampPercentage
 		status.TargetVersion.RampPercentage = &rampPercentage
 	}
--- a/internal/controller/state_mapper.go
+++ b/internal/controller/state_mapper.go
@@ -39,6 +39,8 @@ func (m *stateMapper) mapToStatus(targetVersionID string) *v1alpha1.TemporalWork
 	status.TargetVersion = m.mapTargetWorkerDeploymentVersion(targetVersionID)
 	if m.temporalState.RampingVersionID == targetVersionID {
 		status.TargetVersion.RampingSince = m.temporalState.RampingSince
+		// TODO(Shivam): Temporal server is not emitting the right value for RampLastModifiedAt.
+		// This is going to be fixed by https://github.com/temporalio/temporal/pull/8089.
 		status.TargetVersion.RampLastModifiedAt = m.temporalState.RampLastModifiedAt
 		rampPercentage := m.temporalState.RampPercentage
 		status.TargetVersion.RampPercentage = &rampPercentage
--- a/internal/controller/state_mapper.go
+++ b/internal/controller/state_mapper.go
@@ -5,8 +5,6 @@
 package controller
 
 import (
-	"fmt"
-
 	"github.com/temporalio/temporal-worker-controller/api/v1alpha1"
 	"github.com/temporalio/temporal-worker-controller/internal/k8s"
 	"github.com/temporalio/temporal-worker-controller/internal/temporal"
@@ -41,9 +39,7 @@ func (m *stateMapper) mapToStatus(targetVersionID string) *v1alpha1.TemporalWork
 	status.TargetVersion = m.mapTargetWorkerDeploymentVersion(targetVersionID)
 	if m.temporalState.RampingVersionID == targetVersionID {
 		status.TargetVersion.RampingSince = m.temporalState.RampingSince
-		fmt.Println("Ramping since is", m.temporalState.RampingSince)
-		fmt.Println("Ramp last modified at is", m.temporalState.RampLastModifiedAt)
-		status.TargetVersion.RampLastModifiedAt = m.temporalState.RampLastModifiedAt // This is showing to be an outdated value of Ramp last modified at is 0001-01-01 00:00:00 +0000 UTC
+		status.TargetVersion.RampLastModifiedAt = m.temporalState.RampLastModifiedAt // Server bug! Fixing it....
 		rampPercentage := m.temporalState.RampPercentage
 		status.TargetVersion.RampPercentage = &rampPercentage
 	}
--- a/internal/controller/state_mapper.go
+++ b/internal/controller/state_mapper.go
@@ -39,7 +39,7 @@ func (m *stateMapper) mapToStatus(targetVersionID string) *v1alpha1.TemporalWork
 	status.TargetVersion = m.mapTargetWorkerDeploymentVersion(targetVersionID)
 	if m.temporalState.RampingVersionID == targetVersionID {
 		status.TargetVersion.RampingSince = m.temporalState.RampingSince
-		status.TargetVersion.RampLastModifiedAt = m.temporalState.RampLastModifiedAt // Server bug! Fixing it....
+		status.TargetVersion.RampLastModifiedAt = m.temporalState.RampLastModifiedAt
 		rampPercentage := m.temporalState.RampPercentage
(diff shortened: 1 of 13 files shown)


internal/controller/worker_controller.go
```diff
@@ -157,13 +157,6 @@ func (r *TemporalWorkerDeploymentReconciler) Reconcile(ctx context.Context, req
 		return ctrl.Result{}, err
 	}
 
-	// TODO(jlegrone): Set defaults via webhook rather than manually
-	//                 (defaults were already set above, but have to be set again after status update)
-	if err := workerDeploy.Default(ctx, &workerDeploy); err != nil {
-		l.Error(err, "TemporalWorkerDeployment defaulter failed")
-		return ctrl.Result{}, err
-	}
-
```
Shivs11: we are already setting defaults earlier!

internal/demo/helloworld/temporal_worker_deployment.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: temporal.io/v1alpha1
 kind: TemporalWorkerDeployment
 metadata:
-  name: helloworld
+  name: superman
```
Shivs11: i'll revert this soon - idk why I chose this name maybe because I saw the movie like last week

internal/planner/planner.go
```diff
@@ -240,21 +242,17 @@ func getTestWorkflows(
...
+	// TODO maybe (Shivam): Need to attach the test workflows to the target version
 
 	return testWorkflows
 }
 
-// getTestWorkflowID generates an ID for a test workflow
-func getTestWorkflowID(taskQueue, versionID string) string {
-	return "test-" + versionID + "-" + taskQueue
-}
-
```
Shivs11, line 257: we don't need two instances of `getTestWorkflowID` in our codebase;

internal/k8s/deployments.go
```diff
@@ -290,3 +291,11 @@ func NewDeploymentWithOwnerRef(
 		},
 	}
 }
+
+// getTestWorkflowID generates a consistent ID for test workflows
+func GetTestWorkflowID(versionID, taskQueue string) string {
+	fmt.Println("versionID", versionID)
+	fmt.Println("taskQueue", taskQueue)
+
```
Shivs11: right now - `test-helloworld/default.123-taskqueue`

Shivs11: future: `test-helloworld.default:123-taskqueue`

internal/planner/planner.go
```diff
@@ -360,13 +364,15 @@ func handleProgressiveRollout(
...
 
 	// Move to the next step if it has been long enough since the last update
-	if rampLastModifiedAt.Add(currentStep.PauseDuration.Duration).Before(currentTime) {
-		if i < len(steps)-1 {
-			vcfg.RampPercentage = steps[i+1].RampPercentage
-			return vcfg
-		} else {
-			vcfg.SetCurrent = true
-			return vcfg
+	if rampLastModifiedAt != nil {
```
Shivs11, line 358: keep the check

internal/controller/state_mapper.go
```diff
@@ -39,7 +39,7 @@ func (m *stateMapper) mapToStatus(targetVersionID string) *v1alpha1.TemporalWork
 	status.TargetVersion = m.mapTargetWorkerDeploymentVersion(targetVersionID)
 	if m.temporalState.RampingVersionID == targetVersionID {
 		status.TargetVersion.RampingSince = m.temporalState.RampingSince
-		status.TargetVersion.RampLastModifiedAt = m.temporalState.RampLastModifiedAt
+		status.TargetVersion.RampLastModifiedAt = m.temporalState.RampLastModifiedAt // Server bug! Fixing it....
```
Shivs11: add a quick unit test please

internal/temporal/worker_deployment.go
```diff
@@ -82,7 +83,7 @@ func GetWorkerDeploymentState(
 	if routingConfig.RampingVersion != "" {
 		var (
 			rampingSinceTime   = metav1.NewTime(routingConfig.RampingVersionChangedTime)
-			lastRampUpdateTime = metav1.NewTime(workerDeploymentInfo.RoutingConfig.RampingVersionPercentageChangedTime)
+			lastRampUpdateTime = metav1.NewTime(routingConfig.RampingVersionPercentageChangedTime)
```
Shivs11, line 85: we have a variable above called routingConfig so it's better if we just use it

internal/controller/state_mapper.go
jlegrone: Sorry small thing but can this comment either be
- removed (if no further changes needed after bug fixed on the server side)
- reformatted as a TODO with more information about how/when this should be refactored in the future (ie. link to gh issue).

internal/controller/worker_controller.go
jlegrone: Are you sure this is no longer needed? I think the TODO comment is still relevant since we don't have the webhook configured as part of the helm chart yet.

internal/temporal/worker_deployment_test.go
```diff
@@ -82,20 +83,20 @@ func TestGetTestWorkflowID(t *testing.T) {
...
 			versionID:      "worker.app.v2",
-			expected:       "test-worker.app-queue.main-worker.app.v2",
+			expected:       "test-worker.app.v2-queue.main",
 		},
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			id := getTestWorkflowID(tt.deploymentName, tt.taskQueue, tt.versionID)
+			id := k8s.GetTestWorkflowID(tt.versionID, tt.taskQueue)
```
jlegrone: Nit: it's a bit weird for test cases to remain in this package after the function has moved. Could we either move `GetTestWorkflowID` back here or move the tests there? Putting the function in `temporal` rather than `k8s` might make the most since given it doesn't directly relate to k8s.

internal/planner/planner_test.go
```diff
@@ -1163,6 +1163,44 @@ func TestGetVersionConfig_ProgressiveRolloutEdgeCases(t *testing.T) {
...
 		},
+		"nil rampLastModifiedAt should not cause a panic": {
+			strategy: temporaliov1alpha1.RolloutStrategy{
+				Strategy: temporaliov1alpha1.UpdateProgressive,
+				Steps: []temporaliov1alpha1.RolloutStep{
+					{
+						RampPercentage: 10,
+						PauseDuration:  metav1.Duration{Duration: 30 * time.Second},
+					},
+					{
```
carlydf: nit: I think there's no need for multiple steps here, and doing only one would make the test shorter

internal/temporal/worker_deployment.go
```diff
@@ -154,8 +156,13 @@ func GetTestWorkflowStatus(
 			continue
 		}
 
+		//Adding task queue information to the current temporal state
+		temporalState.Versions[versionID].TaskQueues = append(temporalState.Versions[versionID].TaskQueues, temporaliov1alpha1.TaskQueue{
```
carlydf, line 159: I see this TaskQueues field on the base version (https://github.com/temporalio/temporal-worker-controller/blob/ad183cd4176b1682f00920e501541235203b2b7d/api/v1alpha1/worker_types.go#L185-L186) -- do these task queues need to be added there too in order for them to correctly show up when someone describes a version?

Shivs11:
@jlegrone - so regarding [this](https://github.com/temporalio/temporal-worker-controller/pull/92#discussion_r2226572271): this is technically a blocker for the controller as the controller needs this so that we can ramp a version properly; 

I already have a [PR](https://github.com/temporalio/temporal/pull/8089/files#diff-94bd372d568afd0355ed15a61ac16f60cf90c1315126cfaaf8b5cb8d2900d83f) open on the server side which shall get completely rolled out in about 2 ish weeks; 

In terms of unblocking the controller, we do have the integ tests which shall have a non-nil value for this field which shall allow us to test our controller out before we make the release. If one of us would really like to test this against temporal cloud, one would have to wait maybe another week since our test namespaces live pretty near the first "ring" of deployments.

Personally, I think we should keep the logic as-is rather than change it to account/handle server's error since this is not released to folks out yet.

carlydf (commented):
two minor comments, mostly looks good to me but I haven't seen Jacob's comments yet

internal/controller/state_mapper.go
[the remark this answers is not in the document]
Shivs11: ouch

internal/temporal/worker_deployment_test.go
Shivs11: I am happy to move this to the temporal package since I think the tests are nice where they are rn!

internal/planner/planner.go
```diff
@@ -360,13 +355,15 @@ func handleProgressiveRollout(
 	}
 
 	// Move to the next step if it has been long enough since the last update
```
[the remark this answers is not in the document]
Shivs11: eh, I think that's kinda self explanatory when one looks at the code so I am not going to alter the comment here

internal/controller/worker_controller.go
Shivs11: very sorry - should have re-read the second half of the comment!

internal/temporal/worker_deployment.go
Shivs11: we are doing that already over here where we append the list of task-queues from the temporal state variable on to the base version; the error previously was that we were not attaching these task-queues on to the temporal state variable

https://github.com/temporalio/temporal-worker-controller/blob/3e1a0e00436d1e059a9ea001f88a2b226cc0789f/internal/controller/state_mapper.go#L97

internal/controller/state_mapper.go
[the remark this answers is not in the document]
jlegrone: lol 🔥

PR #93, opened 2025-07-23
Use `rollout` rather than `cutover` in CRD JSON.

robholland:
This matches the terms used internally.


internal/demo/helloworld/changes/progressive-rollout.patch
```diff
@@ -4,8 +4,8 @@ index 926fc27..3e1fd51 100644
 +++ b/internal/demo/helloworld/temporal_worker.yaml
 @@ -12,9 +12,20 @@ spec:
      temporalNamespace: "replay-2025.ktasd"
-   # How to cut over new workflow executions to the latest worker version
```
jlegrone, line 7: Need to update this as well for the patch to apply https://github.com/temporalio/temporal-worker-controller/blob/6529dc111dc68396b2eacfa6f562644db301523f/internal/demo/helloworld/temporal_worker_deployment.yaml#L15

robholland: Serves me right for using `--assume-unchanged`. Fixed now.

PR #94, opened 2025-07-25
Update local README + demo 

Shivs11:
## What was changed
- updated the README so that folks can follow these instructions and test out our worker-controller!
- also updated the skaffold file so that building the controller image and deploying to a local k8s cluster is pain-free.

## Why?
- to partially solve #88; the first half where I clean up the conceptual overview (README present in the root of this repo) shall be coming out soon!

## Checklist

1. Closes 

2. How was this tested:
- ran all these commands locally and they work

3. Any docs updates needed?


skaffold.yaml
Shivs11, line 1: I am almost tempted to move this skaffold file inside the `demo` but don't wanna do so on the grounds that there might be some reasoning otherwise; I can't think of any and since this seems to be dealing with the demo directly it makes sense to put it there

having said that, I am curious to hear opinions/thoughts from @carlydf / @jlegrone!

internal/demo/README.md
```diff
@@ -0,0 +1,138 @@
...
+   ```
+
+### Available Rollout Strategies
+
+The controller supports three deployment strategies:
+
+🚀 **ALL-AT-ONCE**: Immediately routes all new workflows to the latest version
+- Ideal for initial deployments when no traffic exists
+- Faster deployment, appropriate when risk is minimal
+- Used by default in our demo for the unversioned → v1 transition
```
carlydf:
```suggestion
- Used by default when there is no current version, so that versioned workers can start accepting tasks from the new version's task queues immediately (this is demonstrated in the unversioned → v1 transition in our demo)
```

internal/demo/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+📈 **PROGRESSIVE**: Gradually shifts traffic using defined percentage steps
+- Best for production with existing traffic: minimizes risk by allowing observation at each step
+- Configured with `rampPercentage` and `pauseDuration` steps:
+  - `rampPercentage`: What percentage of new workflows go to the new version (e.g., 10% to v2, 90% to v1)
+  - `pauseDuration`: How long to wait at each percentage before moving to the next step
+- Demonstrated in our demo for the v1 → v2 transition with live traffic
```
carlydf:
```suggestion
- Demonstrated in our demo for the v1 → v2 transition with live traffic
- Note that if the current version is nil (as in an initial versioned worker deployment), the progressive steps will be skipped and the first version of the worker deployment will become current immediately.
```

internal/demo/README.md
```diff
@@ -0,0 +1,138 @@
...
+   skaffold dev --profile worker-controller
+   ```
+
+### Available Rollout Strategies
+
+The controller supports three deployment strategies:
+
+🚀 **ALL-AT-ONCE**: Immediately routes all new workflows to the latest version
+- Ideal for initial deployments when no traffic exists
+- Faster deployment, appropriate when risk is minimal
```
carlydf: or when trying to rapidly deploy a bug fix

skaffold.yaml
carlydf: run skaffold commands with different profiles, from top level.

PR #101, opened 2025-07-31
change default deployment name to `resource_namespace`

carlydf:
## What was changed and Why
We discussed using a `.` in the past, but the server won't be able to support that before the release of the controller. A `/` feels off if the ordering is `resource/namespace` because a resource is a sub-component of a namespace, and `/` implies file-system type hierarchy.

I had a [convo with ChatGPT](https://chatgpt.com/share/e/688bebdf-9e6c-8004-8149-dd742bcc0ae0) comparing some options just for fun, and I agree with it that an underscore separator with resource name followed by namespace name feels like the best option.

## Checklist

1. Closes 

2. How was this tested:
Integration and unit tests

3. Any docs updates needed?

### DIFF
--- a/internal/k8s/deployments.go
+++ b/internal/k8s/deployments.go
@@ -24,8 +24,8 @@ const (
 	DeployOwnerKey = ".metadata.controller"
 	// BuildIDLabel is the label that identifies the build ID for a deployment
 	BuildIDLabel             = "temporal.io/build-id"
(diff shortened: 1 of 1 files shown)


Shivs11:
can't seem to load the convo with our new best friend, so don't wanna preemptively approve this rn

carlydf:
I changed my mind, I think resource/namespace is fine actually

PR #102, opened 2025-08-01
TemporalConnection update propagation

Shivs11:
## What was changed
- Handle updates to TemporalConnection, for TWD's that are associated with one, so that these updates are propagated deployments.
- This was achieved by adding the hashed contents of a TemporalConnection as pod annotations. Thus, if a TemporalConnection were to get updated, it would result in new deployments with the updated annotation. The new updated secret would also be mounted so that the workers starting can be started with the latest secrets.

## Why?
- core functionality

## Checklist

1. Closes 
- #10 

4. How was this tested:
- Added unit tests and also tested this locally. The only thing I was not able to do was add an integration test since I was not able to configure our temporaltest server using mTLS. 

Here's how I tested this locally:
- I ran a worker pod which was using the `temporal-cloud-mtls` secret to connect to Temporal.
- I updated the namespace the worker was connected to by changing the mTLS secret. This resulted in the controller and the worker pods throwing connection errors (expected)
- I updated the local secret in the local k8s cluster (and also changed it's name to `temporal-cloud-mtls-1` since we want our controller to also refresh the client it's using)
- Noticed that a rolling deployment was conducted and everything stabalized.

Proof of a rolling update:
Screenshot 2025-08-07 at 5 41 34 PM

5. Any docs updates needed?
- None


internal/controller/worker_controller.go
```diff
@@ -112,8 +113,9 @@ func (r *TemporalWorkerDeploymentReconciler) Reconcile(ctx context.Context, req
 
 	// Get or update temporal client for connection
 	temporalClient, ok := r.TemporalClientPool.GetSDKClient(clientpool.ClientPoolKey{
-		HostPort:  temporalConnection.Spec.HostPort,
-		Namespace: workerDeploy.Spec.WorkerOptions.TemporalNamespace,
+		HostPort:        temporalConnection.Spec.HostPort,
+		Namespace:       workerDeploy.Spec.WorkerOptions.TemporalNamespace,
+		MutualTLSSecret: temporalConnection.Spec.MutualTLSSecret,
```
Shivs11, line 118: @jlegrone - curious to hear your thoughts on this:

I added the name of the `MutualTLSSecret` as a key over here since a change to the secret name, for the same connectionSpec object, should trigger our worker controller to refresh the client it is using! This was required since I realized that there could be a world where the old secret is not expired but just replaced.

However, one area where this breaks is if someone just replaces the *contents* of an existing secret without changing the name. In this way, the controller does not really get a new client.

config/rbac/role.yaml
```diff
@@ -0,0 +1,66 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
```
carlydf: I think you committed this `config` directory accidentally. I believe the `make manifests` make target generates these files in the wrong place, and that needs to be fixed, but in the meantime could you remove this directory from your PR?

internal/controller/clientpool/clientpool.go
```diff
@@ -22,8 +22,9 @@ import (
 )
 
 type ClientPoolKey struct {
-	HostPort  string
-	Namespace string
+	HostPort        string
+	Namespace       string
+	MutualTLSSecret string // Include secret name in key to invalidate cache when the secret name itself changes
```
carlydf:
```suggestion
	MutualTLSSecret string // Include secret name in key to invalidate cache when the secret name changes
```

carlydf: nit

internal/controller/worker_controller.go
carlydf: Could you solve the problem of content changing without name changing by storing a hash of the name + contents instead of just the name?

Or, if that's inefficient because it means you have to read the secret contents all the time just to generate the map key, we could do a different solution for the contents changing, where if an error occurs when calling the temporal APIs that suggests "wrong credentials", we could try reloading them once just in case it's changed. Then, do a namespace describe to make sure they work, and if they work, restart all the workers.

I don't think this needs to be solved in this PR though, but if you don't solve the "changed content, not expired, same name" thing, do just create an issue about it.

internal/controller/worker_controller.go
```diff
@@ -212,9 +214,35 @@ func (r *TemporalWorkerDeploymentReconciler) SetupWithManager(mgr ctrl.Manager)
...
 			RecoverPanic:            &recoverPanic,
 		}).
 		Complete(r)
 }
+
+func (r *TemporalWorkerDeploymentReconciler) findTWDsUsingConnection(ctx context.Context, tc client.Object) []reconcile.Request {
+	var requests []reconcile.Request
+
+	// Find all TWDs in same namespace that reference this TC
+	var workers temporaliov1alpha1.TemporalWorkerDeploymentList
```
carlydf: I think "workers" is a confusing name here, because you're listing `TemporalWorkerDeployment`s which are not workers.

internal/controller/worker_controller.go
```diff
@@ -212,9 +214,35 @@ func (r *TemporalWorkerDeploymentReconciler) SetupWithManager(mgr ctrl.Manager)
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&temporaliov1alpha1.TemporalWorkerDeployment{}).
 		Owns(&appsv1.Deployment{}).
+		Watches(&temporaliov1alpha1.TemporalConnection{}, handler.EnqueueRequestsFromMapFunc(r.findTWDsUsingConnection)).
```
carlydf, line 217: nice!

internal/planner/planner.go
```diff
@@ -85,6 +88,85 @@ func GeneratePlan(
 	return plan, nil
 }
 
+// checkAndUpdateDeploymentConnectionSpec checks if a deployment needs to be updated. This is true if the connection spec it's using is outdated.
```
carlydf: this comment reads as though the return value is a bool, but it's a *Deployment. Could you say what nil vs non-nil *Deployment means?

internal/planner/planner.go
```diff
@@ -85,6 +88,85 @@ func GeneratePlan(
...
+	connection temporaliov1alpha1.TemporalConnectionSpec,
+) *appsv1.Deployment {
+	existingDeployment, exists := k8sState.Deployments[versionID]
+	if !exists {
+		return nil
+	}
+
+	// If the connection spec hash has changed, update the deployment
+	currentHash := k8s.ComputeConnectionSpecHash(connection)
+	if currentHash != existingDeployment.Spec.Template.Annotations[k8s.ConnectionSpecHashAnnotation] {
```
carlydf, line 107: is it possible for annotations to ever be nil? like if the user doesn't provide any or something. Could you make sure to handle the case where the user provides no initial annotations?

internal/controller/worker_controller.go
Shivs11: I agree - changed this to now be `twd`

internal/planner/planner.go
Shivs11: Here is the code which builds pod annotations:

```
func ComputeConnectionSpecHash(connection temporaliov1alpha1.TemporalConnectionSpec) string {
	// HostPort is required, but MutualTLSSecret can be empty for non-mTLS connections
	if connection.HostPort == "" {
		return ""
	}

	hasher := sha256.New()

	// Hash connection spec fields in deterministic order
	_, _ = hasher.Write([]byte(connection.HostPort))
	_, _ = hasher.Write([]byte(connection.MutualTLSSecret))

	return hex.EncodeToString(hasher.Sum(nil))
}
```

Right now, pod annotations will actually never be nil. The piece of code conducting the check `connection.HostPort == ""` is actually a sign of defensive programming. In the odd world where some client were to forget to specify the hostport in the `TemporalConnection` spec, the controller would error out earlier in the reconciliation loop. 

Specifically, this would happen when it would try to create/fetch a client since having a `nil` hostport would force the controller to try creating a new SDK client trying to connect to `nil` temporal hostport. This shall fail!

internal/controller/worker_controller.go
Shivs11: thanks for responding to this; 

yes, I did think about the `storing a hash of the name + contents` idea but I decided against it because in one of my earlier PR's, we discussed that reading the contents of a secret on every reconciliation loop could be intensive. 

having said that, I like the `APIs that suggests "wrong credentials", we could try reloading them once just in case it's changed` idea but don't think it should come in this PR. I can make a separate issue for this!

config/rbac/role.yaml
Shivs11: ah yes; just verified that running `make manifests` is what created this file - removed it

PR #107, opened 2025-08-06
Release artifacts

robholland:
Part of https://github.com/temporalio/temporal-worker-controller/issues/11.

Future PRs will:

* Either add or switch to Dockerhub from ghcr.
* Switch the helm chart to using these images.
* Bump the helm chart version and publish an OCI image for it when a new app version is released.

### DIFF
--- a/.github/workflows/publish-latest-image.yml
+++ b/.github/workflows/publish-latest-image.yml
@@ -43,10 +43,10 @@ jobs:
 
       - name: Push snapshot images
         run: |
-          docker push ghcr.io/${GITHUB_REPOSITORY}:latest-amd64
-          docker push ghcr.io/${GITHUB_REPOSITORY}:latest-arm64
+          docker push "ghcr.io/${GITHUB_REPOSITORY}:latest-amd64"
+          docker push "ghcr.io/${GITHUB_REPOSITORY}:latest-arm64"
       
       - name: Create and push manifest for :latest tag
         run: |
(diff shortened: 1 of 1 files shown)


robholland:
You can see how the releases/images look at https://github.com/robholland/temporal-worker-controller/releases and https://github.com/robholland/temporal-worker-controller/pkgs/container/temporal-worker-controller.

Note that this PR does not do any automation of creating releases, it just creates artifacts when a release is created. I suggest we create releases manually for now.

carlydf (approved):
Approved (assuming you'll replace GHCR with temporalio dockerhub as a follow up).

If I understand correctly, I think that publishing to https://hub.docker.com/u/temporalio directly from this repo would require maintaining some secrets to push the image.

In case it's helpful, the server and admin tools images are published to temporaliotest dockerhub by [temporalio/docker-builds docker.yml workflow](https://github.com/temporalio/docker-builds/blob/main/.github/workflows/docker.yml) and then "released" by moving them from temporaliotest to temporalio with [this workflow](https://github.com/temporalio/docker-builds/blob/main/.github/workflows/release-temporal.yml). I don't think it has to be as complicated as the docker-builds repo is; that's just to manage the dependency between server and admintools.

FWIW, the temporal repo workflow that triggers this pipeline is [trigger-publish.yml](https://github.com/temporalio/temporal/blob/main/.github/workflows/trigger-publish.yml)

PR #112, opened 2025-08-12
Migration demo

carlydf:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


internal/demo/helm/templates/worker.yaml
```diff
@@ -1,15 +1,30 @@
...
+    steps:
+      - rampPercentage: 1
+        pauseDuration: 30s
+      - rampPercentage: 50
+        pauseDuration: 1m
+    gate:
+      workflowType: "HelloWorld"
+  sunset:
+    scaledownDelay: 30s
+    deleteDelay: 1m
```
carlydf: these are optional

PR #132, opened 2025-08-29
Check versioned and unversioned pollers in certain cases

carlydf:
## What was changed
1. Check unversioned pollers for the target version if the current version is nil and strategy is progressive
2. Check versioned pollers for Drained versions if the controller is not running a k8s Deployment for that version

## Why?
1. So that we can still respect Progressive rollout steps if Current Version is nil and we confirm unversioned pollers for all Target Version task queues
2. So that we can only block rollout of a new version if we have a pathological # of ineligible for delete versions, meaning in most cases, users can rely on the server to delete their old versions.

## Checklist

1. Closes 

3. How was this tested:
Tests coming asap :)

4. Any docs updates needed?

### DIFF
--- a/api/v1alpha1/worker_types.go
+++ b/api/v1alpha1/worker_types.go
@@ -61,8 +61,19 @@ type TemporalWorkerDeploymentSpec struct {
 	// TODO(jlegrone): add godoc
 	WorkerOptions WorkerOptions `json:"workerOptions"`
 
-	// MaxVersions defines the maximum number of worker deployment versions allowed.
-	// This helps prevent hitting Temporal's default limit of 100 versions per deployment.
+	// MaxVersionsIneligibleForDeletion defines the maximum number of worker deployment versions
+	// allowed that are not eligible for deletion allowed.
+	// This helps prevent hitting Temporal's default limit of 100 versions per deployment and
+	// being unable to create new version.
+	//
+	// When a Worker Deployment has the maximum number of versions (100 by default), it will
+	// delete the oldest eligible version when a worker with the 101st version arrives.
+	// If no versions are eligible for deletion, that worker's poll will fail, which is dangerous.
+	// To protect against this, when a Worker Deployment has hits the maximum versions that are
+	// ineligible for deletion, the controller will stop deploying new workers in order to give the
+	// user the opportunity to adjust their sunset policy to avoid this situation before it actually
+	// blocks deployment of a new worker version on the server side.
+	//
 	// Defaults to 75. Users can override this by explicitly setting a higher value in
 	// the CRD, but should exercise caution: once the server's version limit is reached,
 	// Temporal attempts to delete an eligible version. If no version is eligible for deletion,
@@ -71,7 +82,7 @@ type TemporalWorkerDeploymentSpec struct {
 	// in dynamicconfig.
 	// +optional
 	// +kubebuilder:validation:Minimum=1
-	MaxVersions *int32 `json:"maxVersions,omitempty"`
+	MaxVersionsIneligibleForDeletion *int32 `json:"maxVersionsIneligibleForDeletion,omitempty"`
 }
 
 // VersionStatus indicates the status of a version.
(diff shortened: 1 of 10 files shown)


api/v1alpha1/worker_types.go
```diff
@@ -71,7 +82,7 @@ type TemporalWorkerDeploymentSpec struct {
 	// in dynamicconfig.
 	// +optional
 	// +kubebuilder:validation:Minimum=1
-	MaxVersions *int32 `json:"maxVersions,omitempty"`
+	MaxVersionsIneligibleForDeletion *int32 `json:"maxVersionsIneligibleForDeletion,omitempty"`
```
carlydf: global on the controller side (via helm chart)?

carlydf: connection CRD is per-temporal-environment

carlydf: remove from public API

api/v1alpha1/worker_types.go
```diff
@@ -141,6 +152,12 @@ type TemporalWorkerDeploymentStatus struct {
 	// This includes current, target, ramping, and deprecated versions.
 	// +optional
 	VersionCount int32 `json:"versionCount,omitempty"`
+
+	// VersionCountIneligibleForDeletion is the total number of versions currently known by the worker
+	// deployment that are ineligible for deletion.
+	// A version is eligible for deletion if it is Drained and has no pollers.
+	// +optional
+	VersionCountIneligibleForDeletion int32 `json:"versionCountIneligibleForDeletion,omitempty"`
```
carlydf: put this on a deprecated version

api/v1alpha1/worker_types.go
```diff
@@ -267,7 +284,8 @@ type RolloutStrategy struct {
 	// - "AllAtOnce": start 100% of new workflow executions on the new worker deployment version as soon as it's healthy;
 	// - "Progressive": ramp up the percentage of new workflow executions targeting the new worker deployment version over time;
 	//
-	// Note: If the Current Version of a Worker Deployment is nil, the controller will ignore any Progressive Rollout
+	// Note: If the Current Version of a Worker Deployment is nil and the controller cannot confirm that all Task Queues
+	// in the Target Version have at least one unversioned poller, the controller will ignore any Progressive Rollout
```
carlydf:
```suggestion
	// in the Target Version have at least one unversioned poller, the controller will ignore any Progressive Rollout
```
"ignore is weirdly passive"

nit: move docs up to sepcific policies

internal/temporal/worker_deployment.go
```diff
@@ -26,6 +29,20 @@ type VersionInfo struct {
...
+	// Used to decide whether to fast track the rollout; rollout will be AllAtOnce if:
+	//   - Current Version is nil
+	//   - Strategy is Progressive, and
+	//   - Presence of unversioned pollers in all task queues of target version cannot be confirmed.
+	AllTaskQueuesHaveUnversionedPoller bool
+	// True if all task queues in this version have no versioned pollers.
+	// False could just mean unknown / not checked / not checked successfully.
+	// Only checked for Drained versions that don't have controller-managed Deployments.
+	// Used to compute status.VersionCountIneligibleForDeletion.
+	NoTaskQueuesHaveVersionedPoller bool
```
carlydf, line 50: this is the meat of the new info that we're getting from server, what it's for, and when we get it

internal/testhelpers/workers.go
```diff
@@ -109,9 +109,6 @@ func newClient(ctx context.Context, hostPort, namespace string) (client.Client,
 // RunHelloWorldWorker runs one worker per replica in the pod spec. callback is a function that can be called multiple times.
 func RunHelloWorldWorker(ctx context.Context, podTemplateSpec corev1.PodTemplateSpec, callback func(stopFunc func(), err error)) {
 	w, stopFunc, err := newVersionedWorker(ctx, podTemplateSpec)
-	defer func() {
-		callback(stopFunc, err)
-	}()
```
carlydf, line 114: realized during testing that this was wrong

internal/controller/genplan.go
```diff
@@ -94,6 +94,7 @@ func (r *TemporalWorkerDeploymentReconciler) generatePlan(
 		connection,
 		plannerConfig,
 		workerDeploymentName,
+		getControllerMaxDeploymentVersionsIneligibleForDeletion(),
```
jlegrone: Can we call this once at startup and set it as a field on `TemporalWorkerDeploymentReconciler`? That way we can handle the Atoi() parsing error and setting a bad value via the env var will surface right away when updating the controller.

internal/testhelpers/workers.go
jlegrone: Maybe we could use [T.Cleanup](https://pkg.go.dev/testing#T.Cleanup) for stuff like this (if the helper accepts a `testing.TB` argument).

PR #133, opened 2025-09-03
Keep dots in Build IDs and format Worker Deployment names `<k8s-namespace>/<twd-name>`

carlydf:
## What was changed
Keep dots in Build IDs and format Worker Deployment names `<k8s-namespace>/<twd-name>`

## Why?
Dots are allowed in pod names and its nice to maintain the semantic version appearance of an image.
Namespace first means the same alphabetization as `kubectl list --all-namespaces`.
Namespace first is more expected for a slash delimiter that implies hierarchy

## Checklist

1. Closes 

2. How was this tested:
Unit tests and functional tests

3. Any docs updates needed?

### DIFF
--- a/internal/k8s/deployments.go
+++ b/internal/k8s/deployments.go
@@ -162,8 +162,8 @@ func CleanAndTruncateString(s string, n int) string {
 	if len(s) > n && n > 0 {
 		s = s[:n]
 	}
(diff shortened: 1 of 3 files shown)


jlegrone (approved):
🎉

PR #137, opened 2025-09-09
Refactor README for Public Preview

### LINKED ISSUE #87: General structure of README's
- A general structure about how the different README's in the repo would look like

carlydf:
## What was changed
Refactor README

## Why?
Update information and make it more user friendly

## Checklist

1. Closes #87 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,15 +9,14 @@
 
 **The Temporal Worker Controller makes it simple and safe to deploy Temporal workers on Kubernetes.**
 
-Instead of worrying about breaking running workflows when you deploy new code, the controller automates 
-[rainbow deployments](https://docs.temporal.io/production-deployment/worker-deployments/worker-versioning#deployment-systems), 
-managing multiple [versions](https://docs.temporal.io/production-deployment/worker-deployments/worker-versioning) of your
-workers so that existing Pinned workflows continue running uninterrupted while new workflows and existing AutoUpgrade workflows use the latest version.
+Temporal workflows require deterministic execution, which means updating worker code can break running workflows if the changes aren't backward compatible. Traditional deployment strategies force you to either risk breaking existing workflows or use Temporal's [Patching API](https://docs.temporal.io/patching) to maintain compatibility across versions.
+
+Temporal's [Worker Versioning](https://docs.temporal.io/production-deployment/worker-deployments/worker-versioning) feature solves this dilemma by providing programmatic control over worker versions and traffic routing. The Temporal Worker Controller automates a deployment system that uses Worker Versioning on Kubernetes. When you deploy new code, the controller automatically creates a new worker version while keeping the old version running. Existing workflows continue on the old version while new workflows use the new version. This approach eliminates the need for patches in many cases and ensures running workflows are never disrupted.
 
 ## What does it do?
 
-🔒 **Protected Pinned workflows** - Workflows pinned to a version stay on that version and won't break  
-🎚️ **Controlled rollout for AutoUpgrade workflows** - AutoUpgrade workflows shifted to new versions with configurable safety controls  
+🔒 **Protected [Pinned](https://docs.temporal.io/worker-versioning#pinned) workflows** - Workflows pinned to a version stay on that version and won't break  
+🎚️ **Controlled rollout for [AutoUpgrade](https://docs.temporal.io/worker-versioning#auto-upgrade) workflows** - AutoUpgrade workflows shifted to new versions with configurable safety controls  
 📦 **Automatic version management** - Registers versions with Temporal, manages routing rules, and tracks version lifecycle  
 🎯 **Smart traffic routing** - New workflows automatically get routed to your target worker version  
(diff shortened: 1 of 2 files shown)


docs/architecture.md
```diff
@@ -0,0 +1,119 @@
+# Architecture
+
+This document provides a detailed technical overview of how the Temporal Worker Controller works, including architecture diagrams, worker lifecycle management, and the relationship between Kubernetes resources and Temporal Worker Deployments.
+
+## How It Works
+
+Every `TemporalWorkerDeployment` resource manages one or more standard `Deployment` resources. Each Deployment manages pods which in turn poll Temporal for tasks routed to their respective worker versions.
+
+```mermaid
```
Shivs11, line 11: Image

Over here, I am not sure what "Deployment ...." directly means - Did you wanna do this intentionally? Moreover, I think it could also be a little neater if this box could be placed outside the two versions to show that this is the deployment both these versions belong to

docs/architecture.md
```diff
@@ -0,0 +1,119 @@
+# Architecture
+
+This document provides a detailed technical overview of how the Temporal Worker Controller works, including architecture diagrams, worker lifecycle management, and the relationship between Kubernetes resources and Temporal Worker Deployments.
+
+## How It Works
+
+Every `TemporalWorkerDeployment` resource manages one or more standard `Deployment` resources. Each Deployment manages pods which in turn poll Temporal for tasks routed to their respective worker versions.
```
Shivs11, line 9: Maybe we should also state that a single TWD manages a single Temporal Worker Deployment.

README.md
```diff
@@ -1,168 +1,171 @@
...
+[![License](https://img.shields.io/github/license/temporalio/temporal-worker-controller)](LICENSE)
+[![Go Report Card](https://goreportcard.com/badge/github.com/temporalio/temporal-worker-controller)](https://goreportcard.com/report/github.com/temporalio/temporal-worker-controller)
 
-The goal of the Temporal Worker Controller is to make it easy to run workers on Kubernetes while leveraging
-[Worker Deployments](https://docs.temporal.io/production-deployment/worker-deployments).
+> 🚀 **Public Preview**: This project is in [Public Preview](https://docs.temporal.io/evaluate/development-production-features/release-stages) and ready for production use cases*. Core functionality is complete with stable APIs.
+> 
+> *Dynamic auto-scaling based on workflow load is not yet implemented. Use cases must work with fixed worker replica counts.
 
-## Why
```
Shivs11, line 8: Is there a specific reason you chose to remove the "why" section over here? As someone who has not stared at the docs for over a month now, I went over the two versions (one with the Why section (as is currently present on main) and the Why section present at the bottom of your PR)

The existing "Why" section felt like a nice way for me to fully understand why this project came into being while giving me some context on why patching is annoying  - I really appreciated having an introduction to Worker Versioning as well as a quick introduction to rainbow deploys and how the controller uses these

You, and other folks, may not agree to this but I still thought it was a good idea to place this out here given that I am almost looking at this with fresh eyes!

carlydf: I removed it because people should look at https://docs.temporal.io/production-deployment/worker-deployments/worker-versioning for the why of worker versioning -- I don't want to repeat that information here.

docs/architecture.md
carlydf: this is a kubernetes Deployment, so it's 1 per version

docs/architecture.md
carlydf: hm you're right. We already explain that in docs/concepts.md though, so I think what I should do is link out to that explanation from here

docs/architecture.md
carlydf: both "Deployment ..." and the Deployment to the left of it are in the Deprecated Versions box, because they represent kubernetes Deployments for deprecated versions. "Deployment ..." is meant to represent that there are multiple other k8s Deployments hosting workers for the deprecated versions

README.md
```diff
@@ -1,168 +1,171 @@
...
+TEMPORAL_DEPLOYMENT_NAME=my-worker        # Unique worker deployment name
+TEMPORAL_WORKER_BUILD_ID=v1.2.3          # Version identifier
 ```
 
-## Contributing
+**Important**: Don't set the above environment variables manually - the controller manages these automatically.
+
+## 🤝 Contributing
+
+This project is in early stages. While external code contributions are not yet being solicited, we welcome:
```
drewhoskins-temporal: Let's think about this.

docs/architecture.md
Shivs11: oh, lol should have seen this - cool, makes sense

README.md
Shivs11: I guess the issue with the linked document is that it is not super clear as to what the why of worker-versioning is from there - moreover, I think having a small paragraph just here rather than making the user jump ships and look for the relevant information could be a little mental exercise we could avoid

wdyt?

carlydf: I think Worker Versioning is so complicated that people need to read the real docs before using the controller. I think it's unsafe to use the controller without having read the real docs

PR #141, opened 2025-09-11
pin artifact onto specific arch

Shivs11:
## What was changed
- Pin artifact onto the specific architecture that can support it.

## Why?
- So we don't get exec errors (we are getting them right now)

## Checklist

1. Closes 

2. How was this tested:
- Not tested yet. I am going to test this one out by first merging to main and then using the newly built image on the running controller in our staging environment.

3. Any docs updates needed?

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -40,7 +40,11 @@ checksum:
   algorithm: sha256
 
 dockers:
-  - image_templates:
+  - ids:
+      - nix
+    goos: linux
+    goarch: amd64
+    image_templates:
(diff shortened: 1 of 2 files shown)


.goreleaser.yml
```diff
@@ -58,7 +62,11 @@ dockers:
     extra_files:
       - LICENSE
 
-  - image_templates:
+  - ids:
+      - nix
```
carlydf, line 66: do you know what this is? anyways, I'm down to try this. Ideally we should also have an image that's just called `main` or `version tag` and is a multi-architecture image so that users don't need to choose what version to use. but first lets try this

PR #147, opened 2025-09-24
Throw error and retry Reconcile if Deployment is NotFound

carlydf:
## What was changed?
Return NotFound error as an error instead of catching it and returning an empty TemporalState.

## Why?
Returning an empty TemporalState on NotFound error means that the controller believes versions to be NotRegistered when in fact they are Registered and the NotFound error was temporary. 

The controller will delete the Deployment of any non-target version that is NotRegistered, so incorrectly perceiving a version as NotRegistered is very dangerous!

## Checklist

1. Closes 

2. How was this tested:
It is very clear what the bug is (based on timestamps and server logs) so we want to get this out without testing, but will follow up with a test tomorrow that ensures Deployment Not Found does NOT result in any Deprecated Versions being scaled down.

3. Any docs updates needed?


Shivs11 (approved):
yay

PR #148, opened 2025-09-26
Fix: Initial deployment without current version does not get promoted 

Shivs11:
## What was changed
- Right now, if someone were to deploy the first version of a fresh new deployment, the version keeps on being Inactive and does not get promoted to Current immediately. 
- Note: This only happens if there is a gated workflow specified on a TWD. This bug does not take place when there isn't a gated workflow.

## Why?
- We have the following invariant:  _Note that when there is no current version (as in an initial versioned worker deployment), the progressive steps are skipped and v1 becomes the current version immediately. All new workflow executions will now start on v1._ 

## Checklist

1. Closes 

2. How was this tested:
- Added two new unit tests.
- Also tested this out locally.
- Integration tests.

3. Any docs updates needed?

### DIFF
--- a/internal/tests/internal/integration_test.go
+++ b/internal/tests/internal/integration_test.go
@@ -76,18 +76,6 @@ func TestIntegration(t *testing.T) {
 					WithTargetVersion("v1.0", temporaliov1alpha1.VersionStatusCurrent, -1, true, false).
 					WithCurrentVersion("v1.0", true, false),
 			),
-		"progressive-rollout-no-gate-no-current-expect-all-at-once": testhelpers.NewTestCase().
-			WithInput(
-				testhelpers.NewTemporalWorkerDeploymentBuilder().
-					WithProgressiveStrategy(testhelpers.ProgressiveStep(10, time.Hour)).
-					WithReplicas(2).
-					WithTargetTemplate("v1.0"),
-			).
-			WithExpectedStatus(
-				testhelpers.NewStatusBuilder().
-					WithTargetVersion("v1.0", temporaliov1alpha1.VersionStatusCurrent, -1, true, false).
-					WithCurrentVersion("v1.0", true, false),
-			),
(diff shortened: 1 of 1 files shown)


internal/planner/planner.go
```diff
@@ -339,8 +339,7 @@ func getTestWorkflows(
 	// Skip if there's no gate workflow defined, if the target version is already the current, or if the target
 	// version is not yet registered in temporal
 	if config.RolloutStrategy.Gate == nil ||
-		status.CurrentVersion == nil ||
-		status.CurrentVersion.BuildID == status.TargetVersion.BuildID ||
+		(status.CurrentVersion != nil && status.CurrentVersion.BuildID == status.TargetVersion.BuildID) ||
```
Shivs11, line 342: If the target version is nil, as will be for an initial version deployed, the task queue should have a test workflow on it (also adhering to the comment right above this function)

internal/tests/internal/integration_test.go
```diff
@@ -63,6 +63,31 @@ func TestIntegration(t *testing.T) {
...
+					WithReplicas(2).
+					WithGate(true).
+					WithTargetTemplate("v1.0"),
+			).
+			WithExpectedStatus(
+				testhelpers.NewStatusBuilder().
+					WithTargetVersion("v1.0", temporaliov1alpha1.VersionStatusCurrent, -1, true, false).
+					WithCurrentVersion("v1.0", true, false),
+			),
+		"progressive-rollout-no-gate-no-current-expect-all-at-once": testhelpers.NewTestCase().
```
carlydf: this is a duplicate of `"progressive-rollout-no-unversioned-pollers-expect-all-at-once"`, so I think you can omit it

carlydf (approved):
approved once you remove the duplicate test and change the title of the PR to better describe the problem that was fixed

PR #149, opened 2025-09-30
Add API key support for the worker-controller

Shivs11:
## What was changed
- WISOTT

## Why?
- Extend the controller to users who use API's to connect to Temporal cloud namespaces

## Checklist

1. Closes 

2. How was this tested:
- Added unit tests 
- Ran the controller locally with both API and mTLS auth
- Existing suite of integration tests

3. Any docs updates needed?
- Updated demo README and relevant scripts as they were out of date.


internal/planner/planner.go
```diff
@@ -339,8 +339,7 @@ func getTestWorkflows(
 	// Skip if there's no gate workflow defined, if the target version is already the current, or if the target
 	// version is not yet registered in temporal
 	if config.RolloutStrategy.Gate == nil ||
-		status.CurrentVersion == nil ||
-		status.CurrentVersion.BuildID == status.TargetVersion.BuildID ||
+		(status.CurrentVersion != nil && status.CurrentVersion.BuildID == status.TargetVersion.BuildID) ||
```
Shivs11, line 342: This is required but is part of a different open PR: https://github.com/temporalio/temporal-worker-controller/pull/148

internal/demo/util/client.go
```diff
@@ -38,9 +38,9 @@ func newClient(buildID string, metricsPort int) (c client.Client, stopFunc func(
 		panic(err)
 	}
 
-	if _, err := c.CheckHealth(context.Background(), &client.CheckHealthRequest{}); err != nil {
-		panic(err)
-	}
+	// if _, err := c.CheckHealth(context.Background(), &client.CheckHealthRequest{}); err != nil {
```
Shivs11: I found out that when paired with API keys, this API call does not work - the only other call that could work is GetSystemInfo but I am not totally sure if we require this over here since we would be notified from client.Dial if anything has gone wrong between connecting our client to the server accepting our client connection

carlydf: hmm, the [CheckHealth API](https://pkg.go.dev/go.temporal.io/sdk@v1.36.0/internal#CheckHealthRequest) doesn't say anything that makes me think it's expected to not work with API keys -- could there be something wrong here?

Could you ask the SDK team to confirm that the error you got was expected?

internal/controller/clientpool/clientpool.go
```diff
@@ -21,16 +21,35 @@ import (
 	runtimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+type AuthMode string
+
+const (
+	AuthModeTLS      AuthMode = "TLS"
+	AuthModeAPIKey   AuthMode = "API_KEY"
+	AuthModeInMemory AuthMode = "IN_MEMORY"
```
carlydf: What does `IN_MEMORY` mean?

internal/controller/clientpool/clientpool.go
```diff
@@ -21,16 +21,35 @@ import (
...
-	MutualTLSSecret string // Include secret name in key to invalidate cache when the secret name changes
+	HostPort   string
+	Namespace  string
+	SecretName string   // Include secret name in key to invalidate cache when the secret name changes
+	AuthMode   AuthMode // Include auth mode in key to invalidate cache when the auth mode changes for the secret
+}
+
+type MTLSAuth struct {
+	tlsConfig  *tls.Config
+	expiryTime time.Time // cert NotAfter - buffer
```
carlydf: This comment is unclear to me -- would you mind elaborating?

api/v1alpha1/temporalconnection_types.go
```diff
@@ -32,6 +32,11 @@ type TemporalConnectionSpec struct {
 	// https://kubernetes.io/docs/concepts/configuration/secret/#tls-secrets
 	// +optional
 	MutualTLSSecretRef *SecretReference `json:"mutualTLSSecretRef,omitempty"`
+
+	// APIKeyRef is the name of the Secret that contains the API key. The secret must be `type: kubernetes.io/opaque` and exist
+	// in the same Kubernetes namespace as the TemporalConnection resource.
+	// +optional
+	APIKeyRef *SecretReference `json:"apiKeyRef,omitempty"`
```
carlydf: Is there anything else we need to communicate about the expected format of the secret? Like that the API Key should be stored under the key `"api-key"`. 

```
	clientOpts.Credentials = sdkclient.NewAPIKeyDynamicCredentials(func(ctx context.Context) (string, error) {
		return string(secret.Data["api-key"]), nil
	})
```

internal/controller/clientpool/clientpool.go
```diff
@@ -142,24 +146,129 @@ func (cp *ClientPool) UpsertClient(ctx context.Context, opts NewClientOptions) (
...
+	clientOpts := sdkclient.Options{
+		Logger:    cp.logger,
+		HostPort:  opts.Spec.HostPort,
+		Namespace: opts.TemporalNamespace,
+		ConnectionOptions: sdkclient.ConnectionOptions{
+			TLS: &tls.Config{},
+		},
+	}
+
+	clientOpts.Credentials = sdkclient.NewAPIKeyDynamicCredentials(func(ctx context.Context) (string, error) {
```
carlydf, line 177: how does the `secret` object stay in memory? what is the benefit of using `DynamicCredentials` instead of just a string

internal/controller/clientpool/clientpool.go
```diff
@@ -142,24 +146,129 @@ func (cp *ClientPool) UpsertClient(ctx context.Context, opts NewClientOptions) (
...
+		auth: ClientAuth{
+			mode: AuthModeAPIKey,
+			mTLS: nil,
+		},
 	}
+
+	return c, nil
+}
+
+func (cp *ClientPool) fetchClientUsingInMemoryConnection(opts NewClientOptions) (sdkclient.Client, error) {
```
carlydf: thoughts on calling this "Unsafe" or "NoCredentials" instead of "InMemory"?
"InMemory" feels like the wrong term to me because all of these clients and connections live "in memory"

And using no mTLS or API Key auth is not exclusive to the in-memory server, it works for any server that is running without credential-based auth

internal/controller/worker_controller.go
```diff
@@ -36,12 +37,21 @@ const (
 	buildIDLabel   = "temporal.io/build-id"
 )
 
-// getMutualTLSSecretName extracts the mutual TLS secret name from a secret reference
-func getMutualTLSSecretName(secretRef *temporaliov1alpha1.SecretReference) (string, bool) {
```
carlydf, line 40: did you change the import name on purpose? I thought `make fmt-imports` wanted the package name to be `temporaliov1alpha1` but I could be wrong

internal/demo/README.md
```diff
@@ -29,45 +29,73 @@ This guide will help you set up and run the Temporal Worker Controller locally u
      ```bash
      make create-cloud-mtls-secret
      ```
-   - Create a `skaffold.env` file:
+   - In `skaffold.example.env`, set:
```
carlydf: I think the purpose of `skaffold.example.env` is for people to copy it and edit their own copy, not to edit it directly. Not to be pedantic, but could you keep the demo instructions in line with that user flow?

internal/demo/README.md
```diff
@@ -29,45 +29,73 @@ This guide will help you set up and run the Temporal Worker Controller locally u
...
+   - Create the Kubernetes Secret (trims any accidental newlines):
+     ```bash
+     make create-api-key-secret
+     ```
+   - In `skaffold.example.env`, set:
+     ```env
+     TEMPORAL_API_KEY_SECRET_NAME=temporal-api-key
+     TEMPORAL_MTLS_SECRET_NAME=""
+     ```
+   - Note: Do not set both mTLS and API key for the same connection. If both present, the controller
```
carlydf:
```suggestion
   - Note: Do not set both mTLS and API key for the same connection. If both are present, the controller 
```

carlydf: The behavior of the controller when two secrets are present is a good thing to think about. IMO, we should add a TemporalConnection validator that makes it so that TemporalConnections with both the API SecretRef and mTLS SecretRef field populated are rejected as invalid. Then the controller won't have to make the call about which to use. WDYT?

internal/demo/util/client.go
```diff
@@ -38,10 +38,6 @@ func newClient(buildID string, metricsPort int) (c client.Client, stopFunc func(
 		panic(err)
 	}
 
-	if _, err := c.CheckHealth(context.Background(), &client.CheckHealthRequest{}); err != nil {
```
carlydf, line 41: I think this should work. What is the error? If it doesn't work, we should understand why and replace it with a better health check.

internal/k8s/deployments.go
```diff
@@ -306,6 +322,7 @@ func NewDeploymentWithOwnerRef(
 	}
 }
 
+// TODO (Shivam): Change hash when secret name is updated as well.
```
carlydf, line 321: are you going to do that in this PR?

internal/planner/planner.go
carlydf: I think if you rebase onto main, this part of the diff will go away (since #148 was merged to main)

internal/planner/planner_test.go
```diff
@@ -1204,6 +1204,42 @@ func TestGetVersionConfigDiff(t *testing.T) {
 			expectSetCurrent:  true,
 			expectRampPercent: func() *int32 { f := int32(0); return &f }(),
 		},
+		{
+			name: "gate configured with no current version should set current immediately",
```
carlydf, line 1208: this part of the diff should also go away after you rebase onto main I thinnk

internal/demo/util/client.go
jacobstr: I encountered this as well in my own fork:

```
 {"level":"error","ts":1759797658.093413,"msg":"Observed a panic","controller":"temporalworkerdeployment","controllerGroup":"temporal.io","controllerKind":"TemporalWorkerDeployment","TemporalWorkerDeployment":{"name":"weather-basic-worker-controller","namespace":"temporal-demo-python-worker"},"namespace":"temporal-demo-python-worker","name":"weather-basic-worker-controller","reconcileID":"f51bdb58-c64e-4783-88e2-d23680b9dcdf","panic":"health check error: Request unauthorized.","panicGoValue":"&fmt.wrapError{msg:\"health check error: Request unauthorized.\", err:(*serviceerror.PermissionDenied)(0xc000b9c090)}","stacktrace":"goroutine 211 [running]:\nk8s.io/apimachinery/pkg/util/runtime.logPanic({0x2324430, 0xc000a3cde0}, {0x1cb5700, 0xc000e9c000})\n\t/go/pkg/mod/k8s.io/apimachinery@v0.34.0/pkg/util/runtime/runtime.go:132
 ```

---

Is it possible the API Key is restricted to a temporal namespace and CheckHealth is a system-wide endpoint? Speculative - but if I'm using namespaced-scoped `TemporalConnection` CRDs in kubernetes, it's likely they're using namespace-scoped temporal API keys.

That does bring to mind that the docs don't indicate what scopes the key should be provisioned with, if used.

I switched to using a worker listing request and it succeeded FWIW. I'm not sure how necessary this might be though - I was getting 401's earlier in the process when my creds were in fact borked... If I'm not mistaken!

---

Repro with:

```
	fmt.Println("\n=== Test 2: Client with API key authentication ===")
	clientOpts2 := sdkclient.Options{
		HostPort:    serverAddr,
		Namespace:   temporalNamespace,
		Credentials: sdkclient.NewAPIKeyStaticCredentials(apiKey),
		ConnectionOptions: sdkclient.ConnectionOptions{
			TLS: &tls.Config{}, // Basic TLS + API key
		},
	}

	c2, err2 := sdkclient.Dial(clientOpts2)
	if err2 != nil {
		fmt.Printf("❌ Failed to create client: %v\n", err2)
	} else {
		defer c2.Close()
		ctx2, cancel2 := context.WithTimeout(context.Background(), 10*time.Second)
		defer cancel2()

		_, err2 := c2.CheckHealth(ctx2, &sdkclient.CheckHealthRequest{})
		if err2 != nil {
			fmt.Printf("❌ Health check failed: %v\n", err2)
		} else {
			fmt.Printf("✅ Health check succeeded\n")
		}
	}
```

File: api/v1alpha1/temporalconnection_types.go (hunk 5 above)
Comment (jacobstr) [reply api/v1alpha1/temporalconnection_types.go]: IMO 

```
corev1 "k8s.io/api/core/v1"

APIKeySecretRef *corev1.SecretKeySelector `json:"apiKeySecretRef,omitempty"`
```

Is a more familiar way to do this as in:

```
valueFrom:
    name: <secret name>
    key: <secret key>
```

File: internal/controller/clientpool/clientpool.go (hunk 3 above)
Comment (Shivs11) [reply internal/controller/clientpool/clientpool.go]: In-memory in this context just means “no external auth; local, ephemeral backend.” The client talks to a local/in-process service using plaintext (no TLS), and no API key or certificates are required.

This is required right now for our integration tests that test the client against the in-memory ephemeral DB that it spins up

Comment (Shivs11) [reply internal/controller/clientpool/clientpool.go]: added a comment where I have defined this ENUM

File: internal/k8s/deployments.go (hunk 12 above)
Comment (Shivs11) [reply internal/k8s/deployments.go:321]: Going to work on it in a different subsequent PR - have an issue opened for this already: https://github.com/temporalio/temporal-worker-controller/issues/113

File: internal/demo/util/client.go (hunk 11 above)
Comment (Shivs11) [reply internal/demo/util/client.go:41]: hi @jacobstr, thank you for digging in. Apologies, I should have placed the reasoning behind the deletion of these lines in a comment or something.

I spoke to the SDK team and the reasoning more or less aligns with @jacobstr's assumption of API Key being restricted to a temporal namespace and CheckHealth is a system-wide endpoint. Further more, the SDK has interceptors that add headers to a specific request if a specific condition is met. After peeking at the code, I realized that once such condition is the requirement of a namespace on the HealthCheck request which it does not have. Due to this reason, the namespace is not added in the request header by the interceptor and we get an unauth error

File: internal/demo/util/client.go (hunk 2 above)
Comment (Shivs11) [reply internal/demo/util/client.go]: answered below in the thread.

File: internal/demo/README.md (hunk 9 above)
Comment (Shivs11) [reply internal/demo/README.md]: oh no you are right - fixed it in my latest commit!

File: internal/demo/util/client.go (hunk 11 above)
Comment (Shivs11) [reply internal/demo/util/client.go:41]: I spoke with Chad from SDK and he confirmed a hunch that I had, which was: 

> client.Dial already calls get system info by default. Health check adds nothing.
(it actually used to do health check back in the day, but we changed to get system info because get system info is a superset of health check - not only does it check health, it also returns us some valuable information)

In other words, we don't need to do an additional health check since the .Dial call is doing a getSystemInfo call behind the scenes

File: internal/controller/clientpool/clientpool.go (hunk 6 above)
Comment (Shivs11) [reply internal/controller/clientpool/clientpool.go:177]: DynamicCredentials, from the API documentation: 

> NewAPIKeyDynamicCredentials creates credentials powered by a callback that is invoked on each request. The callback accepts the context that is given by
the calling user and can return a key or an error. When error is non-nil, the
client call is failed with that error. When string is non-empty, it is used
as the API key. When string is empty, nothing is set/overridden.
This is the equivalent of providing a headers provider that returns the
"Authorization" header with "Bearer " + the given function result. If the
resulting string is non-empty, it will overwrite any "Authorization" header
that may be on the context or from existing header provider.

In other words, a client created using this option shall load the latest API key that it can fetch on every single request that it makes. (Which is what we want since it reduces code as we don't need to write more logic for client certs refreshers, which is something we have to do for TLS based auth)

Comment (Shivs11) [reply internal/controller/clientpool/clientpool.go:177]: ah, another thing to note (and i am just realizing this): we technically have a cache for our API key secrets. So, every-time someone were to change their API secret name and content, we would generate a new SDK client for them. What I mean by this is we could, in theory, use NewAPIKeyStaticCredentials and have this work. 

However, I think I am going to revamp this logic a bit when the controller would support client refreshes on secret content change only. When that happens, we would need to use NewAPIKeyDynamicCredentials.

File: internal/controller/clientpool/clientpool.go (hunk 7 above)
Comment (Shivs11) [reply internal/controller/clientpool/clientpool.go]: calling it "Unsafe" feels a bit eh to me - I would not mind calling it "NoCredentials" because it does represent that tbf

File: api/v1alpha1/temporalconnection_types.go (hunk 5 above)
Comment (carlydf) [reply api/v1alpha1/temporalconnection_types.go]: +1 to @jacobstr 
Kian's Gate Input PR has an example as well https://github.com/temporalio/temporal-worker-controller/pull/152/files#diff-a3557816ba113bd59b7f3ede479f75ecef4965a17763f079ecb9b9f794f4239eR293

File: internal/controller/clientpool/clientpool.go (hunk 7 above)
Comment (carlydf) [reply internal/controller/clientpool/clientpool.go]: ok cool. I'm happy with a variety of things, just not `InMemory`

File: internal/demo/README.md (hunk 10 above)
Comment (Shivs11) [reply internal/demo/README.md]: Agreed. Shall add it in this PR itself!

File: config/rbac/role.yaml [hunk 14]
```diff
@@ -0,0 +1,66 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
```
carlydf: I think this was accidentally committed (sidenote, we should fix the `make manifests` target to put these files in the right place after generating them, but for this PR just remove the extra files)

config/webhook/manifests.yaml
```diff
@@ -0,0 +1,26 @@
+---
+apiVersion: admissionregistration.k8s.io/v1
```
carlydf: same as above

internal/controller/clientpool/clientpool.go
```diff
@@ -21,16 +22,35 @@ import (
...
+	// Add more auth modes here as they are supported
+)
+
 type ClientPoolKey struct {
-	HostPort        string
-	Namespace       string
-	MutualTLSSecret string // Include secret name in key to invalidate cache when the secret name changes
+	HostPort   string
+	Namespace  string
+	SecretName string   // Include secret name in key to invalidate cache when the secret name changes
```
carlydf: unlikely but what if the secret name stays the same, but the key name changes in the `SecretKeyConfigRef`?

internal/controller/worker_controller.go
```diff
@@ -36,12 +38,44 @@ const (
...
+	}
+}
+
+func getAuthMode(temporalConnection *temporaliov1alpha1.TemporalConnection) clientpool.AuthMode {
+	if temporalConnection.Spec.MutualTLSSecretRef != nil {
+		return clientpool.AuthModeTLS
+	} else if temporalConnection.Spec.APIKeySecretRef != nil {
+		return clientpool.AuthModeAPIKey
+	}
+	return clientpool.AuthModeNoCredentials
```
carlydf: nit: I only see `getAuthMode` used in this one place, so it's more concise to just do:
```
func resolveAuthSecretName(tc *temporaliov1alpha1.TemporalConnection) (clientpool.AuthMode, string, error) {
    if temporalConnection.Spec.MutualTLSSecretRef != nil {
		name, err := getTLSSecretName(tc.Spec.MutualTLSSecretRef)
		return clientpool.AuthModeTLS, name, err
    } else if temporalConnection.Spec.APIKeySecretRef != nil {
		name, err := getAPIKeySecretName(tc.Spec.APIKeySecretRef)
		return clientpool.AuthModeAPIKey, name, err
    }
	return clientpool.AuthModeNoCredentials, "", nil
}
```

internal/demo/helloworld/helm/helloworld/values.yaml
```diff
@@ -9,6 +9,9 @@ temporal:
   # Connection details (required if connectionName is empty)
```
carlydf:
```suggestion
  # Connection details (either mtlsSecretName or apiKey is required if connectionName is empty)
```
(Or feel free to say that only one or the other is required in another way)

internal/demo/helloworld/helm/helloworld/values.schema.json
```diff
@@ -35,6 +35,20 @@
...
           "type": "string",
           "description": "Name of the secret containing mTLS certificates (required if connectionName is empty)"
+        },
+        "apiKey": {
+          "type": "object",
+          "additionalProperties": false,
+          "properties": {
+            "name": {
+              "type": "string",
+              "description": "Name of the secret containing API key (required if connectionName is empty)"
```
carlydf, line 45: nit: this says that both apiKey and mtlsSecretName are required but only one or the other is required -- right?

Shivs11: These are just the descriptions of individual fields - the real enforcement in the schema is being done in the AnyOf block (present right below the comment of yours)
```

{
          "required": ["address", "mtlsSecretName"]
        },
        {
          "properties": {
            "apiKey": {
              "type": "object",
              "required": ["name", "key"]
            }
          },
          "required": ["address", "apiKey"]
        }
      ]
```

PR #152, opened 2025-10-07
feat: add support for gating rollouts behind successful workflow executions

kianjones9:
This PR adds support for custom inputs to the "gate" workflow, which was originally part of the docs but I discovered was not implemented when I tried to use it 😅. 

## What was changed

This PR includes a few changes:
1. Gated workflow input specification support both in-line and as a ConfigMap
2. Moved CRDs to top level chart directory, which ensures they are installed first per [helm docs](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-1-let-helm-do-it-for-you)
3. Because of 2, TemporalConnection can now be added to the helm templates/values definitions
4. Relevant docs updates

## Why?

Gated workflow success for our use case really depends on being able to specify exactly the workflow inputs.

### Closes

Here is the [link to temporal community slack thread](https://temporalio.slack.com/archives/C07MDJ6S3HP/p1759190414718889) where I discussed with @Shivs11 and @carlydf .

### How was this tested:

Already deployed in our staging cluster, using configMap with a huge workflow input. This workflow must complete successfully or the new workers will not roll out.

### Any docs updates needed?

Updated docs accordingly.


docs/migration-guide.md
```diff
@@ -541,6 +541,17 @@ spec:
...
+      # Inline JSON object:
+      # input:
+      #   thresholds:
+      #     errorRate: 0.01
+      #     p95LatencyMs: 250
+      # Or load from a ConfigMap/Secret key containing JSON:
+      # inputFrom:
+      #   configMapKeyRef:
+      #     name: order-processor-gate-input
+      #     key: payload.json
```
carlydf, line 558: nit: I think you might as well put the SecretKeyRef example here too, as above in the configuration.md

helm/temporal-worker-controller/crds/temporal.io_temporalworkerdeployments.yaml
```diff
@@ -63,6 +63,36 @@ spec:
                 properties:
```
carlydf, line 63: ah I see now that the crds directory was supposed to be in the chart directory, not under `templates/`. I'm going to move these in a separate PR so that the change is more clearly logged in the commit history.

EDIT: here's the PR, will merge soon and then you can pull it in https://github.com/temporalio/temporal-worker-controller/pull/153

helm/temporal-worker-controller/templates/temporal_connection.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: temporal.io/v1alpha1
+kind: TemporalConnection
```
carlydf: The intention of the `temporal-worker-controller` helm chart is to install the controller + necessary CRDs, RBAC stuff, and services to run the controller. One controller can manage `TemporalWorkerDeployment`s that connect to multiple different Temporal namespaces, often with different `TemporalConnection` resources for each namespace (or even for different workers in the same namespace, if they use different credentials).

So, the `TemporalConnection` should be deployed with whatever pipeline you use to deploy your `TemporalWorkerDeployment`. This also helps ensure that the name of the `TemporalConnection` can be synchronized with any `TemporalWorkerDeployment` references to that name.

helm/temporal-worker-controller/values.yaml
```diff
@@ -113,3 +113,9 @@ certmanager:
 # Not yet supported
 prometheus:
   enabled: false
+
+# change to match your environment
+temporal:
```
carlydf: for the reasons above, we don't want to include a TemporalConnection in the temporal-worker-controller chart. TemporalConnections should be deployed separately

internal/controller/execplan.go
```diff
@@ -68,7 +69,28 @@ func (r *TemporalWorkerDeploymentReconciler) executePlan(ctx context.Context, l
 	deploymentHandler := temporalClient.WorkerDeploymentClient().GetHandle(p.WorkerDeploymentName)
 
 	for _, wf := range p.startTestWorkflows {
-		if _, err := temporalClient.ExecuteWorkflow(ctx, sdkclient.StartWorkflowOptions{
+		// Log workflow start details and a safe preview of input
```
carlydf: by "safe" do you just mean <= 512 characters?
Thinking about this, I definitely could see people wanting to use the SecretRef method for specifying input, and not wanting to log the input anywhere (even shortened) because it contains sensitive data. WDYT? Could we allow people to opt out of the logging in a graceful way? Maybe there could be a field on the gate workflow config that says "redact input in logs" or "display input in logs" or something.

internal/controller/genplan.go
```diff
@@ -138,6 +149,56 @@ func (r *TemporalWorkerDeploymentReconciler) generatePlan(
 	return plan, nil
 }
 
+// resolveGateInput resolves the gate input from inline JSON or from a referenced ConfigMap/Secret
+func (r *TemporalWorkerDeploymentReconciler) resolveGateInput(
```
carlydf: I think it would be cool to see a unit test for this function. It looks like it doesn't really need to reference anything in `r *TemporalWorkerDeploymentReconciler`, so it could live in planner/planner.go and get a unit test in planner_test.go. You could add `gateInput` to https://github.com/temporalio/temporal-worker-controller/blob/main/internal/planner/planner.go#L49-L54 and then populate it like the rest of the fields are populated.

helm/temporal-worker-controller/crds/temporal.io_temporalworkerdeployments.yaml
carlydf: merged!

helm/temporal-worker-controller/templates/temporal_connection.yaml
kianjones9: okay gotcha. I honestly just assumed it was bc it might fail to install if the CRDs weren't installed first. I'll update this to remove.

helm/temporal-worker-controller/crds/temporal.io_temporalworkerdeployments.yaml
kianjones9: awesome thanks!

internal/controller/execplan.go
kianjones9: yeah sorry I just meant bc at least in our use case, we have pretty huge workflow inputs, and I'd hate to log the entire thing on each deployment I could see that getting out of hand. Now that I've let it simmer, what do you think about just logging something like `input.keys()`, to make it safe in that regard. 

I think opting out of logging inputs would probably be common enough request that I propose we log some workflow input details as described above, **UNLESS** input is coming from a secretRef object, in which case we don't log at any input. Does that sound reasonable? I think in some ways this may be expected behaviour.

carlydf: > I propose we log some workflow input details as described above, UNLESS input is coming from a secretRef object, in which case we don't log at any input. 

I think this sounds good!

kianjones9: okay I think I sufficiently addressed this now/

helm/temporal-worker-controller/templates/temporal_connection.yaml
carlydf: reminder to remove this file from the helm chart

api/v1alpha1/worker_types.go
```diff
@@ -273,6 +274,23 @@ const (
 
 type GateWorkflowConfig struct {
 	WorkflowType string `json:"workflowType"`
+	// Input is an arbitrary JSON object passed as the first parameter to the gate workflow.
```
carlydf, line 277: can you add to this comment that the input will be logged, but that InputFrom will not be logged?

carlydf (commented):
Please try to fix the "cyclomatic complexity" linter error also.

Once these three comments are addressed, I feel good about merging it with the unit tests that you added + existing integration tests passing + knowing that you've run this locally.

PR #156, opened 2025-10-16
Use an intermediate environment variable in GHA

jackdawm:
## What was changed
Use an intermediate environment variable instead of variable interpolation of untrusted user input

## Why?
Best practice is not to trust user input

## Checklist

1. Closes VLN-399

2. How was this tested:

### DIFF
--- a/.github/workflows/helm.yml
+++ b/.github/workflows/helm.yml
@@ -43,6 +43,8 @@ jobs:
 
       - name: Bump Chart Version
         id: bump_version
+        env:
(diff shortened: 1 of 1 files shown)


robholland (approved):

Shivs11 (approved):
maybe a bad question: This PR only proxies the input into an env var, right? I see the PR description highlighting the fact that we should not be trusting user input as it could be malicious, but is the intention of the PR to also prevent that from happening?

PR #163, opened 2025-11-10
Bug Fix: Ignore LastModifierIdentity if server deleted a version for garbage collection

carlydf:
## What was changed
Bug Fix: Ignore LastModifierIdentity if server deleted a version for garbage collection

Also, fix issue in describe version that would sometimes cause a deprecated version to be Drained with nil DrainedSince time.

## Why?
The server sets the `LastModifierIdentity` to `"try-delete-for-add-version"` when it deletes old versions to do garbage collection upon addition of the a worker version beyond the maximum. We are changing this on the server and will patch v1.29 of the OSS server to handle this more gracefully in v1.29.2 of the server, however, for users who cannot or do not want to use the new patch of the server immediately, this will fix the problem for them on the controller side so that they won't encounter this if they're running against Temporal Server v1.29.1 or v1.29.0.

The issue of nil DrainedSince time caused flakiness in tests that involve Deprecated Drained versions. That flakiness appeared a lot in the new test I added to test the bug fix, so I also fixed that.

## Checklist

1. Closes 

2. How was this tested:
New functional test

3. Any docs updates needed?

### DIFF
--- a/internal/tests/internal/integration_test.go
+++ b/internal/tests/internal/integration_test.go
@@ -184,9 +184,9 @@ func TestIntegration(t *testing.T) {
 				WithWaitTime(5*time.Second).
 				WithExpectedStatus(
 					testhelpers.NewStatusBuilder(). // controller won't deploy v5, so it's not registered
-						WithTargetVersion("v5", temporaliov1alpha1.VersionStatusNotRegistered, -1, false, false).
-						WithCurrentVersion("v4", true, false).
-						WithDeprecatedVersions( // drained but has pollers, so ineligible for deletion
+									WithTargetVersion("v5", temporaliov1alpha1.VersionStatusNotRegistered, -1, false, false).
+									WithCurrentVersion("v4", true, false).
+									WithDeprecatedVersions( // drained but has pollers, so ineligible for deletion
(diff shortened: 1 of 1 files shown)


internal/temporal/worker_deployment.go
```diff
@@ -151,17 +151,37 @@ func GetWorkerDeploymentState(
...
+					return fmt.Errorf("drainage info nil for build %s", version.Version.BuildId)
+				}
+				if desc.Info.DrainageInfo.DrainageStatus != temporalClient.WorkerDeploymentVersionDrainageStatusDrained {
+					return fmt.Errorf("version info does not say that build %s is drained", version.Version.BuildId)
+				}
+				return err
+			}
+			// At first, version is found in DeploymentInfo.VersionSummaries but may not have the full drainage info in
+			// describe version, so we describe with backoff.
+			// If the version was just deleted by the server, we may never succeed at describing it, and it should
```
Shivs11, line 172: nit: This comment was a bit difficult to make sense of - I think what you were trying to say is "The drainage information present in DeploymentInfo.VersionSummaries may not be up-to-date always, hence, we do a DescribeVersion with backoff"

Is that right?

carlydf: I'm going to merge and then fix in the follow up PR with the test matrix

Issue #165, opened 2025-11-13
temporal-operator: Build ID regeneration for non-workflow code changes causes excessive deployments with PINNED versioning


tshtark:
* Overview

We are using the temporal worker controller with PINNED versioning strategy for long-running workflows (days-weeks).

However, we're experiencing deployment proliferation because the build_id is
regenerated for ANY pod spec change, including non-workflow code changes.

 

* Use Case

  - Service: Multiple workflow workers
  - Workflow characteristics:
    - New workflow starts every ~30 minutes
    - Workflows run for days
    - Long-running workflows cannot be patched mid-execution (business requirements)
  - Deployment frequency:
    - Workflow code changes: Infrequent (detected by CI - files ending in .workflow.ts)
    - Non-workflow code changes (main.ts, configuration, dependencies): Every ~2 hours
  - Versioning strategy: PINNED (required due to long-running workflows and desire to make
  breaking changes)

 

* The Problem

  Current Behavior:
  1. ComputeBuildID() hashes the entire pod template spec
  2. ANY change (image tag, env vars, resources, etc.) generates a new build_id
  3. New build_id → new Kubernetes Deployment → new worker version registered with Temporal
  4. With PINNED versioning, old deployments stay alive until their workflows complete

 

* Result:
  - Deploying non-workflow code changes (bug fixes in main.ts, dependency updates, etc.)
  creates new deployments
  - Combined with our timing constraints (new workflow every 30min, workflows run for days), we end up with 10-15 active deployments running simultaneously
  - This causes resource waste and operational complexity

 

* Expected Behavior

  Ideally, we'd like to:
  1. Workflow code changes → new build_id → new deployment/version (current behavior - this is correct)
  2. Non-workflow code changes → same build_id → rolling update of existing deployment pods (new behavior)

  This would allow us to:
  - Deploy bug fixes and non-workflow improvements frequently
  - Keep only 1-2 active deployments at any time
  - Maintain workflow execution integrity with PINNED versioning

carlydf:
Hi there, our reasoning for including the entire pod spec in the build ID was in case things such as secrets and environment variables (esp. feature flags) would cause branching in the Workflow definition that is nondeterministic (even with the same workflow code / image, a feature flag could be a breaking change).

[...]

PR #177, opened 2025-12-11
feat(api): add stable build ID override via spec.workerOptions.customBuildID

tshtark:
## Summary

Adds support for user-controlled build IDs via `spec.workerOptions.customBuildID`, enabling rolling updates for non-workflow code changes while preserving new deployment creation for workflow code changes.

## Problem

With PINNED versioning strategy and long-running workflows, **any** pod spec change (image tag, env vars, resources) generates a new build ID, causing deployment proliferation.

* Related issue - https://github.com/temporalio/temporal-worker-controller/issues/165

**Result:** 10-15 active deployments running simultaneously, causing resource waste and operational complexity.

## Solution

Allow users to set a stable build ID via `spec.workerOptions.customBuildID`. When the build ID is stable but pod spec changes, trigger a rolling update instead of creating a new deployment.

## Key Changes

- Add `CustomBuildID` field to `WorkerOptions` struct in API types
- Update `ComputeBuildID` to use custom field when set
- Implement hash-based drift detection using SHA256 of user-provided pod template spec
- Store hash in `temporal.io/pod-template-spec-hash` annotation on deployments
- Extract `ApplyControllerPodSpecModifications` as shared helper for code reuse
- Only check for drift when `customBuildID` is explicitly set by user

## Drift Detection

When `spec.workerOptions.customBuildID` is set, the controller detects spec drift by comparing a SHA256 hash of the user-provided pod template spec against the hash stored in the deployment annotation.

**How it works:**
1. When a deployment is created, the controller computes a hash of the user-provided pod template spec (before controller modifications) and stores it in an annotation
2. On each reconciliation, the controller computes the hash of the current spec and compares it to the stored hash
3. If hashes differ, a rolling update is triggered

**This approach:**
- Detects **ALL** changes to the pod template spec (images, env vars, commands, volumes, resources, etc.)
- Avoids issues with cluster-provided values and timestamps
- Maintains backwards compatibility (legacy deployments without the hash annotation are not affected)

## Usage

```yaml
metadata:
  name: my-worker
spec:
  workerOptions:
    connectionRef:
      name: my-connection
    temporalNamespace: default
    # Set this to your workflow code hash (e.g., from CI/CD)
    customBuildID: "wf-a1b2c3d4"
  template:
    spec:
      containers:
        - name: worker
          image: my-worker:v1.2.3  # Can change without new deployment
```

## Behavior Matrix

| Scenario | Build ID | Result |
|----------|----------|--------|
| No customBuildID field | Auto-generated from image + hash | Existing behavior |
| customBuildID set, first deploy | Uses field value | New deployment |
| customBuildID unchanged, spec changed | Same build ID | Rolling update |
| customBuildID changed | New build ID | New deployment |
| Empty/invalid customBuildID | Falls back to auto-generated | Existing behavior |

## Backwards Compatibility

- Empty or invalid `customBuildID` values fall back to existing hash-based generation
- No changes required for users who don't use this feature
- Drift detection only runs when `spec.workerOptions.customBuildID` is explicitly set
- Legacy deployments without the pod template spec hash annotation are not affected

## Test plan

- [x] CustomBuildID spec field override (6 test cases)
- [x] Hash computation tests (determinism, different images/env vars/commands/volumes)
- [x] Drift detection with hash comparison (7 test cases)
- [x] Backwards compatibility for legacy deployments without hash annotation
- [x] Edge cases: empty values, invalid chars, long values
- [x] `go test ./...` passes

### DIFF
--- a/internal/k8s/deployments.go
+++ b/internal/k8s/deployments.go
@@ -32,9 +32,9 @@ const (
 	twdNameLabel                  = "temporal.io/deployment-name"
 	WorkerDeploymentNameSeparator = "/"
 	ResourceNameSeparator         = "-"
-	MaxBuildIdLen                    = 63
-	ConnectionSpecHashAnnotation     = "temporal.io/connection-spec-hash"
-	PodTemplateSpecHashAnnotation    = "temporal.io/pod-template-spec-hash"
+	MaxBuildIdLen                 = 63
+	ConnectionSpecHashAnnotation  = "temporal.io/connection-spec-hash"
+	PodTemplateSpecHashAnnotation = "temporal.io/pod-template-spec-hash"
 )
 
 // DeploymentState represents the Kubernetes state of all deployments for a temporal worker deployment
--- a/internal/planner/planner_test.go
+++ b/internal/planner/planner_test.go
@@ -2195,8 +2195,8 @@ func TestCheckAndUpdateDeploymentPodTemplateSpec(t *testing.T) {
 			expectUpdate:       false,
 		},
 		{
-			name:    "no update when buildID is not explicitly set (auto-generated buildID)",
-			buildID: "v1",
+			name:               "no update when buildID is not explicitly set (auto-generated buildID)",
+			buildID:            "v1",
 			existingDeployment: createDeploymentForDriftTest(1, "v1", "old-image:v1"),
 			newSpec: &temporaliov1alpha1.TemporalWorkerDeploymentSpec{
 				Replicas: int32Ptr(1),
@@ -2216,8 +2216,8 @@ func TestCheckAndUpdateDeploymentPodTemplateSpec(t *testing.T) {
 			expectUpdate: false, // No update because BuildID is not explicitly set
 		},
 		{
-			name:    "same image does not trigger update when buildID is set",
-			buildID: "stable-build-id",
+			name:               "same image does not trigger update when buildID is set",
+			buildID:            "stable-build-id",
 			existingDeployment: createDeploymentForDriftTest(1, "stable-build-id", "my-image:v1"),
 			newSpec: &temporaliov1alpha1.TemporalWorkerDeploymentSpec{
 				Replicas: int32Ptr(1),
@@ -2237,8 +2237,8 @@ func TestCheckAndUpdateDeploymentPodTemplateSpec(t *testing.T) {
 			expectUpdate: false, // No update because image is the same
 		},
 		{
-			name:    "different image triggers update when buildID is set",
-			buildID: "stable-build-id",
+			name:               "different image triggers update when buildID is set",
+			buildID:            "stable-build-id",
 			existingDeployment: createDeploymentForDriftTest(1, "stable-build-id", "old-image:v1"),
 			newSpec: &temporaliov1alpha1.TemporalWorkerDeploymentSpec{
 				Replicas: int32Ptr(1),
@@ -2259,8 +2259,8 @@ func TestCheckAndUpdateDeploymentPodTemplateSpec(t *testing.T) {
 			expectImage:  "new-image:v2",
 		},
 		{
-			name:    "replicas-only change does not trigger update (handled by scaling logic)",
-			buildID: "stable-build-id",
+			name:               "replicas-only change does not trigger update (handled by scaling logic)",
+			buildID:            "stable-build-id",
 			existingDeployment: createDeploymentForDriftTest(1, "stable-build-id", "my-image:v1"),
 			newSpec: &temporaliov1alpha1.TemporalWorkerDeploymentSpec{
 				Replicas: int32Ptr(3), // Changed from 1 to 3
@@ -2310,8 +2310,8 @@ func TestCheckAndUpdateDeploymentPodTemplateSpec(t *testing.T) {
 			expectImage:  "my-image:v1",
 		},
 		{
-			name:    "backwards compat: no hash annotation means no update",
-			buildID: "stable-build-id",
+			name:               "backwards compat: no hash annotation means no update",
+			buildID:            "stable-build-id",
 			existingDeployment: createDeploymentWithoutHashAnnotation(1, "stable-build-id", "old-image:v1"),
 			newSpec: &temporaliov1alpha1.TemporalWorkerDeploymentSpec{
 				Replicas: int32Ptr(1),


api/v1alpha1/worker_types.go
```diff
@@ -27,6 +27,25 @@ type WorkerOptions struct {
...
+	// WARNING: Using a custom build ID requires careful management. If workflow code changes
+	// but BuildID stays the same, pinned workflows may execute on workers running incompatible
+	// code. Only use this when you have a reliable way to compute a hash of your workflow
+	// definitions (e.g., hashing workflow source files in CI/CD).
+	//
+	// When the BuildID is stable but pod template spec changes, the controller triggers
+	// a rolling update instead of creating a new deployment version. Currently detected
+	// changes: replicas, minReadySeconds, container images, container resources (limits/requests),
+	// and init container images. Other fields (env vars, volumes, commands) are not currently
+	// monitored for drift.
```
carlydf: I think we should still apply the rolling update if the env vars or commands change. What was the reasoning to not apply those changes?

api/v1alpha1/worker_types.go
```diff
@@ -27,6 +27,25 @@ type WorkerOptions struct {
...
 	TemporalNamespace string `json:"temporalNamespace"`
+	// BuildID optionally overrides the auto-generated build ID for this worker deployment.
+	// When set, the controller uses this value instead of computing a build ID from the
+	// pod template hash. This enables rolling updates for non-workflow code changes
+	// (bug fixes, config changes) while preserving the same build ID.
+	//
+	// WARNING: Using a custom build ID requires careful management. If workflow code changes
+	// but BuildID stays the same, pinned workflows may execute on workers running incompatible
+	// code. Only use this when you have a reliable way to compute a hash of your workflow
+	// definitions (e.g., hashing workflow source files in CI/CD).
```
carlydf:
```suggestion
	// code. Only use this when you have a reliable way to detect changes in your workflow
	// definitions (e.g., hashing workflow source files in CI/CD).
```
nit: "reliable way to compute a hash" followed by the "e.g." is redundant. I think the core of the issue is workflow code changes, so we can just focus on that.

internal/planner/planner.go
```diff
@@ -168,31 +168,294 @@ func updateDeploymentWithConnection(deployment *appsv1.Deployment, connection te
...
+	// This applies all controller modifications (env vars, TLS mounts, etc.)
+	updateDeploymentWithPodTemplateSpec(existingDeployment, spec, connection)
+
+	return existingDeployment
+}
+
+// hasPodTemplateSpecDrift compares the existing deployment with the desired TWD spec
+// to determine if a rolling update is needed. It compares user-controllable fields
+// that would indicate the pod spec has changed.
+func hasPodTemplateSpecDrift(deployment *appsv1.Deployment, spec *temporaliov1alpha1.TemporalWorkerDeploymentSpec) bool {
```
carlydf: I think you proposed initially to store the hash of the submitted pod spec in an annotation on the Deployment, and then, on each reconciliation loop, if the Build ID override is set, take a hash of the new submitted pod spec and compare it to that annotation to detect drift. I actually think that would be the best drift detection option after all, since it allows us to compare the full user-provided spec without dealing with cluster-provided values and timestamps, and without cherry-picking only a subset of fields.

Could you change this to do drift detection in via a hash of the user-submitted pod spec (before the controller creates the Deployment) instead of this per-field method? Sorry for waffling!

carlydf (commented):
Hi there, sorry for the late review, just getting back to this after the holidays. I'm committed to turning this PR around quickly though and I think it's really close! I'll be keeping an eye on this tomorrow so we can iterate quickly.

api/v1alpha1/worker_types.go
tshtark: Fixed! Switched to hash-based drift detection that computes a SHA256 hash of the entire user-provided pod template spec. This now catches ALL changes including env vars, commands, volumes, and any other fields.

Added test `env_var_change_triggers_update_when_buildID_is_set` to verify this behavior.

api/v1alpha1/worker_types.go
tshtark: Applied! Changed to "reliable way to detect changes in your workflow" as suggested.

internal/planner/planner.go
tshtark: Implemented! Now using hash-based drift detection as you suggested:

1. **Hash storage**: `ComputePodTemplateSpecHash()` computes SHA256 of the user-provided pod template spec (before controller modifications)
2. **Annotation**: Hash stored in `temporal.io/pod-template-spec-hash` annotation on the deployment
3. **Comparison**: On each reconciliation, compare stored hash vs. current spec hash
4. **Backwards compat**: Legacy deployments without the hash annotation won't trigger updates

This approach avoids dealing with cluster-provided values/timestamps and captures ALL user-controllable fields without cherry-picking.

Used `github.com/davecgh/go-spew` (already a dependency) with deterministic config for hashing.

internal/planner/planner.go
```diff
@@ -168,31 +168,225 @@ func updateDeploymentWithConnection(deployment *appsv1.Deployment, connection te
...
+	// Pod template has changed - rebuild the pod spec from the TWD spec
+	// This applies all controller modifications (env vars, TLS mounts, etc.)
+	updateDeploymentWithPodTemplateSpec(existingDeployment, spec, connection)
+
+	return existingDeployment
+}
+
+// updateDeploymentWithPodTemplateSpec updates an existing deployment with a new pod template spec
+// from the TWD spec. This applies all the controller modifications that NewDeploymentWithOwnerRef does.
+func updateDeploymentWithPodTemplateSpec(
```
carlydf, line 222: can you re-use the pod-spec-modification code from `NewDeploymentWithOwnerRef` (if the whole thing can't be reused, maybe a shared helper function) so that we don't have to keep two separate functions in sync?

api/v1alpha1/worker_types.go
```diff
@@ -27,6 +27,24 @@ type WorkerOptions struct {
...
+	//
+	// When the BuildID is stable but pod template spec changes, the controller triggers
+	// a rolling update instead of creating a new deployment version. The controller uses
+	// a hash of the user-provided pod template spec to detect ANY changes, including
+	// container images, env vars, commands, volumes, resources, and all other fields.
+	// +optional
+	// +kubebuilder:validation:MaxLength=63
+	// +kubebuilder:validation:Pattern=`^[a-zA-Z0-9]([a-zA-Z0-9._-]*[a-zA-Z0-9])?$`
+	BuildID string `json:"buildID,omitempty"`
```
carlydf: I'd prefer to name this in a way that makes it clear that providing your own Build ID is not required, since as we say in the warning, it's not safe unless you do work on your own side to ensure that Build ID is updated when incompatible changes happen.

I worry that people could get confused, not read the doc-string description of the field, and copy-pasta themselves into a situation where they're not updating the Build ID and get NDEs in their workflows despite "using versioning."

Would you be opposed to calling it something like `CustomBuildID`?

tshtark: Good call! Renamed to `customBuildID` to make it clear this is optional and requires careful management. The name now signals that users are taking on responsibility for build ID management.

internal/planner/planner.go
tshtark: Done! Extracted `ApplyControllerPodSpecModifications()` as a shared helper in `internal/k8s/deployments.go`. It's now used by both `NewDeploymentWithOwnerRef` and `updateDeploymentWithPodTemplateSpec` to keep them in sync.

carlydf (approved):
approved! you'll need to run `make fmt-imports` to get the linter to pass. 
I also changed the repo settings so the CI jobs should all run when you next push instead of requiring admin approval.

PR #179, opened 2025-12-22
Helm chart improvements

### LINKED ISSUE #178: [Feature Request] Helm chart improvements

* Cannot set `replicas` for manager deployment
* Cannot set `tolerations` for manager deployment
* Cannot set `nodeSelector` for manager deployment
* Affinity for manager deployment do not work correctly

nikolaitsarenkov:
## What was changed
Added optional value for `nodeSelector`, `tolerations` and `replicas` for manager deployment.
Fixed `affinity` for manager deployment/

## Checklist

1. Closes 
#178 

2. How was this tested:
Deployed to the local environment

3. Any docs updates needed?
No

### DIFF
--- a/helm/temporal-worker-controller/values.yaml
+++ b/helm/temporal-worker-controller/values.yaml
@@ -50,20 +50,18 @@ securityContext:
   seccompProfile:
     enabled: false
 
-# Configure the nodeAffinity expression according to the platforms which are 
-# supported by your solution. It is considered best practice to support multiple
-# architectures. You can build your manager image using the makefile target docker-buildx.
-affinity: {}
-  # podAntiAffinity:
-  #   requiredDuringSchedulingIgnoredDuringExecution:
-  #     - labelSelector:
-  #         matchExpressions:
-  #           - key: app.kubernetes.io/component
-  #             operator: In
-  #             values:
-  #               - manager
-  #       topologyKey: kubernetes.io/hostname
-  # nodeAffinity: {}
+affinity:
+# Use when running multiple replicas to spread manager pods across nodes.
+# For strict HA, switch to requiredDuringSchedulingIgnoredDuringExecution.
+  podAntiAffinity:
+    preferredDuringSchedulingIgnoredDuringExecution:
+      - weight: 100
+        podAffinityTerm:
+          labelSelector:
+            matchLabels:
+              app.kubernetes.io/component: manager
+              app.kubernetes.io/instance: {{ .Release.Name }}
+          topologyKey: kubernetes.io/hostname
 
 # Opt out of these resources if you want to disable the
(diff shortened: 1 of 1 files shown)


helm/temporal-worker-controller/values.yaml
```diff
@@ -53,34 +53,31 @@ securityContext:
...
+  #           operator: In
+  #           values:
+  #             - amd64
+  #             - arm64
+  #             - ppc64le
+  #             - s390x
+  #         - key: kubernetes.io/os
+  #           operator: In
+  #           values:
+  #             - linux
```
carlydf: Actually looking at this, I think the `nodeAffinity` example is not that helpful for most people, because most nodes meet these architectural environments. If someone has a k8s cluster with some Windows nodes, I expect they probably have already encountered this issue for other deployments in their cluster, and can produce whatever `nodeAffinity` section they want themselves.

Would you be open to just deleting that example?
```suggestion
```

helm/temporal-worker-controller/values.yaml
```diff
@@ -53,34 +53,31 @@ securityContext:
...
+affinity: {}
+  # podAntiAffinity:
+  #   requiredDuringSchedulingIgnoredDuringExecution:
+  #     - labelSelector:
+  #         matchExpressions:
+  #           - key: app.kubernetes.io/component
+  #             operator: In
+  #             values:
+  #               - manager
+  #       topologyKey: kubernetes.io/hostname
```
carlydf: Instead of just providing this as an example, I think we should make the default # of replicas 2, and the default affinity what I propose below, because Temporal workers are often critical infrastructure, so in a production environment, the safest default is to have >1 controller, wdyt?
```
affinity:
# Use when running multiple replicas to spread manager pods across nodes.
# For strict HA, switch to requiredDuringSchedulingIgnoredDuringExecution.
  podAntiAffinity:
    preferredDuringSchedulingIgnoredDuringExecution:
      - weight: 100
        podAffinityTerm:
          labelSelector:
            matchLabels:
              app.kubernetes.io/component: manager
              app.kubernetes.io/instance: {{ .Release.Name }}
          topologyKey: kubernetes.io/hostname
```

carlydf (commented):
Thanks so much for this contribution! Sorry about the untested affinity section. I think we should be able to merge this soon; lmk your thoughts on my suggestions when you get a change

helm/temporal-worker-controller/values.yaml
nikolaitsarenkov: No problem

helm/temporal-worker-controller/values.yaml
nikolaitsarenkov: ok

carlydf (approved):

PR #189, opened 2026-01-14
Rename `CustomBuildID` -> `UnsafeCustomBuildID` and integration test it

carlydf:
## What was changed
See title.

## Why?
- To prevent human users or agents from accidentally filling in this field without considering the full implications.
- To test and prevent regressions

## Checklist

1. Closes 

2. How was this tested:
New test

3. Any docs updates needed?

### DIFF
--- a/internal/testhelpers/make.go
+++ b/internal/testhelpers/make.go
@@ -112,7 +112,10 @@ func MakeTWDWithImage(name, namespace, imageName string) *temporaliov1alpha1.Tem
 // MakeBuildId computes a build id based on the image and
 // If no podSpec is provided, defaults to HelloWorldPodSpec with the given image name.
 // If you provide your own podSpec, make sure to give the first container your desired image name if success is expected.
-func MakeBuildId(twdName, imageName string, podSpec *corev1.PodTemplateSpec) string {
+func MakeBuildId(twdName, imageName, unsafeCustomBuildID string, podSpec *corev1.PodTemplateSpec) string {
+	if unsafeCustomBuildID != "" {
+		return unsafeCustomBuildID
+	}
 	return k8s.ComputeBuildID(
 		ModifyObj(
 			MakeTWDWithName(twdName, ""),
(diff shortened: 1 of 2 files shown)


internal/testhelpers/make.go
```diff
@@ -183,9 +183,14 @@ func MakeDeprecatedVersion(namespace, twdName, imageName string, status temporal
 	return ret
 }
 
-func MakeBaseVersion(namespace, twdName, imageName string, status temporaliov1alpha1.VersionStatus, createDeployment, hasDeployment bool) temporaliov1alpha1.BaseWorkerDeploymentVersion {
+func MakeBaseVersion(namespace, twdName, imageName, unsafeCustomBuildID string, status temporaliov1alpha1.VersionStatus, createDeployment, hasDeployment bool) temporaliov1alpha1.BaseWorkerDeploymentVersion {
+	buildID := unsafeCustomBuildID
```
eniko-dif: nit: I think this logic should be handled inside the MakeBuildId function similarly how ComputeBuildID works to keep build ID handling logic at the same place.

internal/testhelpers/test_builder.go
```diff
@@ -418,7 +444,10 @@ func (tcb *TestCaseBuilder) Build() TestCase {
 		expectedDeploymentReplicas: make(map[string]int32),
 	}
 	for _, info := range tcb.existingDeploymentInfos {
-		buildId := MakeBuildId(tcb.name, info.image, nil)
+		buildId := info.unsafeCustomBuildID
```
eniko-dif: and then you also don't need to duplicate the logic.

internal/testhelpers/make.go
carlydf: done! good idea

Shivs11 (approved):
lgtm - don't think this change is risky as well since it's mostly involving a rename and integ tests to bullet proof the rename

PR #190, opened 2026-01-21
Fix helm chart: Remove templating from values file

carlydf:
## What was changed
The previous helm chart improvements PR templated some labels inside the values file, which breaks the chart.

## Why?
Fix chart build

## Checklist

1. Closes 

2. How was this tested:
Tested by installing to a 2-node minikube cluster with default affinity values and with an override custom affinity.

3. Any docs updates needed?

### DIFF
--- a/internal/demo/README.md
+++ b/internal/demo/README.md
@@ -17,7 +17,7 @@ This guide will help you set up and run the Temporal Worker Controller locally u
 
 1. Start a local Minikube cluster:
    ```bash
-   minikube start --nodes 2
+   minikube start
    ```
 
 2. Create the `skaffold.env` file:


helm/temporal-worker-controller/templates/manager.yaml
```diff
@@ -57,7 +67,7 @@ spec:
         args:
         - --leader-elect
         {{- if .Values.metrics.enabled }}
-        - --metrics-bind-address=127.0.0.1:{{ .Values.metrics.port }}
+        - "--metrics-bind-address=127.0.0.1:{{ .Values.metrics.port }}"
```
Shivs11, line 70: was this (the lack of quotations) an issue?

helm/temporal-worker-controller/values.yaml
```diff
@@ -50,18 +50,12 @@ securityContext:
...
-# Use when running multiple replicas to spread manager pods across nodes.
-# For strict HA, switch to requiredDuringSchedulingIgnoredDuringExecution.
-  podAntiAffinity:
-    preferredDuringSchedulingIgnoredDuringExecution:
-      - weight: 100
-        podAffinityTerm:
-          labelSelector:
-            matchLabels:
-              app.kubernetes.io/component: manager
-              app.kubernetes.io/instance: {{ .Release.Name }}
```
Shivs11, line 63: oh sorry, i guess this was the error (you can't have helm templating inside values.yaml files i think?)

internal/demo/README.md
```diff
@@ -17,7 +17,7 @@ This guide will help you set up and run the Temporal Worker Controller locally u
 
 1. Start a local Minikube cluster:
    ```bash
-   minikube start
+   minikube start --nodes 2
```
Shivs11: not sure how I feel about this - things, right now, will still work on a single node right (since the affinity requirement is a soft one by default). If a user wants to run a demo, they probably don't care about things like HA. Maybe we should have that as an optional section in this document, but keep things as simple as we can for the demo by starting only a single node cluster.

Shivs11 (approved):

helm/temporal-worker-controller/templates/manager.yaml
carlydf: no,  the lack of quotations just made this line red in my IDE, but it still compiled correctly

internal/demo/README.md
carlydf: Hm I guess that makes sense. I didn't think about the preference vs not requirement thing. I'll change this back

PR #194, opened 2026-01-28
Add task queue guidance documentation

bchav:
## Summary
- Adds `docs/task-queues.md` explaining the relationship between TemporalWorkerDeployment and Task Queues
- Documents the recommended pattern of one TWD per Task Queue
- Includes FAQ addressing common questions

## Why
Users are asking whether a single TWD can handle multiple Task Queues, or if each TWD should map to a single worker instance. This doc clarifies:
- Task Queue is defined in application code, not the TWD spec
- One TWD per Task Queue is recommended for independent scaling and rollouts
- When multiple queues per TWD might be acceptable

## Checklist
- [x] Follows existing docs structure and style
- [x] Includes practical YAML examples
- [x] Links to related concepts (TemporalConnection)

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/docs/task-queues.md
+++ b/docs/task-queues.md
@@ -1,0 +1,158 @@
+# Task Queues and TemporalWorkerDeployment
+
+This document explains how Task Queues relate to TemporalWorkerDeployment resources and provides guidance on structuring your deployments.
+
+## Key Concept: Task Queue is Defined in Your Code
+
+The Task Queue is **not** configured in the TemporalWorkerDeployment spec. Instead:
+
+1. The controller injects environment variables into your pods:
+   - `TEMPORAL_ADDRESS`
+   - `TEMPORAL_NAMESPACE`
+   - `TEMPORAL_DEPLOYMENT_NAME`
+   - `TEMPORAL_WORKER_BUILD_ID`
+
+2. Your worker code reads these variables and specifies which Task Queue to poll.
+
+```
+┌──────────────────────────────────────────────────────────────────┐
+│              TemporalWorkerDeployment                            │
+│                                                                  │
+│   Manages:                      Does NOT manage:                 │
+│   - Replicas                    - Task Queue name(s)             │
+│   - Rollout strategy            - Workflows/Activities           │
+│   - Version lifecycle           - Worker business logic          │
+│   - K8s Deployments                                              │
+│   - Env var injection                                            │
+│                                                                  │
+└──────────────────────────────────────────────────────────────────┘
+                         │
+                         │ Creates pods with env vars
+                         ▼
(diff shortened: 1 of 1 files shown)


docs/task-queues.md
```diff
@@ -0,0 +1,158 @@
...
+│  │  Polls: "orders"   │  │  Polls: "payments" │                  │
+│  └────────────────────┘  └────────────────────┘                  │
+│                                                                  │
+│  Each container runs a worker process polling ONE Task Queue     │
+│  All containers share the same TEMPORAL_WORKER_BUILD_ID          │
+│                                                                  │
+└──────────────────────────────────────────────────────────────────┘
+```
+
+> **Note:** Each worker process polls exactly one Task Queue. To handle multiple queues, run multiple containers.
```
carlydf, line 48: From the controller perspective, it's totally fine to handle multiple queues within the same container though. If there are other reasons someone wants to separate pollers of different task queues into different containers, they're free to do that, but I can't think of a compelling benefit.

If different workers have different resource requirements or scaling requirements, they need to be put into different Deployments anyway.

docs/task-queues.md
```diff
@@ -0,0 +1,158 @@
...
+
+- Task Queues are part of the same logical service
+- You want to deploy and version workers for all queues together
+- The queues have similar resource and scaling requirements
+
+## When to Split Task Queues into Separate TWDs
+
+Consider creating separate TemporalWorkerDeployment resources when Task Queues have:
+
+**Different scaling requirements** - One queue may need 10 replicas while another needs 2:
```
carlydf, line 111: This guidance makes sense to me given the current state of the controller.

Someday I'd like to see the controller support creating multiple Deployments with the same Version, so that people can have multiple task queues in the same version, with workers scaled separately. I don't think we should block GA or AutoScaling on that though, so it will just have to come later I suppose

docs/task-queues.md
```diff
@@ -0,0 +1,158 @@
...
+    spec:
+      containers:
+      - name: worker
+        image: order-service:v1.0
+        env:
+        - name: TASK_QUEUE
+          value: "notifications"
+```
+
+**Different deployment cadences** - You want to roll out changes to one queue without affecting others, or test changes on a low-risk queue before rolling to critical queues.
```
carlydf, line 147: This is the best reason to put your worker and task queue in a separate TWD. I'd love to live in a world some day where the only reason to separate your workers is because you want to deploy and version them separately (not just because you want to scale them separately)

jaypipes:
I'm sitting next to @bchav and he gave me instructions to close this.

Issue #192, opened 2026-01-21
[Bug] Helm chart deployment fails on v1.2.0: "container has runAsNonRoot and image has non-numeric user (nonroot)"


gabriel-yahav:
## What are you really trying to do?

Deploy the temporal-worker-controller Helm chart on Kubernetes with default security settings.

## Describe the bug

The v1.2.0 Docker image fails to start when deployed via the Helm chart with the default `runAsNonRoot: true` security context.

**Error:**
```
Error: container has runAsNonRoot and image has non-numeric user (nonroot), cannot verify user is non-root
```

This was introduced in v1.2.0 when `USER nonroot` was added to `Dockerfile.goreleaser`. Kubernetes cannot verify non-root status when the user is specified as a string rather than a numeric UID.

**The change in question:**

```dockerfile
# Before (worked)
FROM gcr.io/distroless/static-debian12:nonroot
COPY temporal-worker-controller /usr/local/bin/temporal-worker-controller
ENTRYPOINT ["temporal-worker-controller"]

# After (broken)
FROM gcr.io/distroless/static-debian12:nonroot
WORKDIR /app
COPY --chown=nonroot:nonroot temporal-worker-controller /usr/local/bin/temporal-worker-controller
USER nonroot  # <-- This line breaks runAsNonRoot
ENTRYPOINT ["temporal-worker-controller"]
```

The `USER nonroot` line is redundant since `gcr.io/distroless/static-debian12:nonroot` already runs as UID 65532, but it causes Kubernetes to see a non-numeric user in the image metadata.

## Minimal Reproduction

```bash
helm install temporal-worker-controller \
  oci://docker.io/temporalio/temporal-worker-controller \
  --namespace devops
```

Then observe the pod failing with `CreateContainerConfigError`.

## Environment/Versions

- **OS and processor:** Linux x86_64 (EKS)
- **Temporal Version:** temporal-worker-controller v1.2.0
- **Are you using Docker or Kubernetes:** Kubernetes (EKS)

## Additional context

**Suggested fix:** Change the Dockerfile to use a numeric UID:

```dockerfile
USER 65532:65532
```

Or remove the `USER` line entirely since the base image already sets it.

Issue #193, opened 2026-01-21
Helm chart: Add runAsUser support to pod securityContext


tshtark:
### Problem

When deploying the Helm chart in environments with strict Pod Security Standards (e.g., ArgoCD, OpenShift), the deployment fails with:

```
Error: container has runAsNonRoot and image has non-numeric user (nonroot), cannot verify user is non-root
```

### Root Cause

The `manager.yaml` template hardcodes `runAsNonRoot: true` at the pod securityContext level ([line 52-53](https://github.com/temporalio/temporal-worker-controller/blob/main/helm/temporal-worker-controller/templates/manager.yaml#L52-L53)):

```yaml
securityContext:
  runAsNonRoot: true
```

However, the container image uses a non-numeric user (`nonroot`). Kubernetes requires a **numeric UID** to verify that the user is not root. Without `runAsUser` specified, Kubernetes cannot perform this verification and rejects the pod.

### Proposed Solution

Add support for `runAsUser` in the Helm values. 

**In `values.yaml`:**
```yaml
securityContext:
  runAsUser: 65532  # Optional: numeric UID for the nonroot user
  seccompProfile:
    enabled: false
```

**In `manager.yaml`:**
```yaml
securityContext:
  runAsNonRoot: true
  {{- with .Values.securityContext.runAsUser }}
  runAsUser: {{ . }}
  {{- end }}
```

### Workaround

Currently requires Kustomize post-rendering to patch the deployment, which is cumbersome.

### Environment
- Chart version: 0.12.0
- Kubernetes: 1.28+
- Deployment method: ArgoCD

PR #195, opened 2026-01-28
Helm chart bug Fix: Make sure image has nonRoot   

Shivs11:
## What was changed
- Make sure that the image, during create time, has the right image config so that kubelet can rightly validate against it.
- kubelet could not prove that the string "USER nonroot" is not the root. The following changes do that.

## Why?
- Correctness.

## Checklist

1. Closes 
#192 
#193 

3. How was this tested:

4. Any docs updates needed?

### DIFF
--- a/Dockerfile.goreleaser
+++ b/Dockerfile.goreleaser
@@ -2,8 +2,6 @@ FROM gcr.io/distroless/static-debian12:nonroot
 
 WORKDIR /app
 
-COPY --chown=nonroot:nonroot temporal-worker-controller /usr/local/bin/temporal-worker-controller
-
-USER nonroot
+COPY --chown=65532:65532 temporal-worker-controller /usr/local/bin/temporal-worker-controller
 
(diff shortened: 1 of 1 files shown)


Dockerfile.goreleaser
```diff
@@ -2,8 +2,6 @@ FROM gcr.io/distroless/static-debian12:nonroot
 
 WORKDIR /app
 
-COPY --chown=nonroot:nonroot temporal-worker-controller /usr/local/bin/temporal-worker-controller
-
-USER nonroot
+COPY --chown=65532:65532 temporal-worker-controller /usr/local/bin/temporal-worker-controller
```
Shivs11, line 5: also changed this to just have 65532 rather than nonroot to be more reliable: I think --chown=nonroot:nonroot depends on the build environment being able to resolve the names nonroot and nonroot to IDs during the COPY.

no need to maybe take this risk and just be explicit in nature.

PR #197, opened 2026-02-10
feat: add KEDA/HPA managed scaling support

m1rl0k:
Added opt-in KEDA/HPA scaling support via `temporal.io/keda-managed` annotation on TemporalWorkerDeployment resources.
When this annotation is present:
- Non-zero scale operations are filtered out (only scale-to-zero is preserved for sunset behavior)
- Existing replicas are preserved during deployment updates to avoid fighting KEDA
This is a ~17 line change in `internal/controller/genplan.go` that is fully backwards compatible - existing users without the annotation see zero behavior change.

 Why?
Users running Temporal workers on Kubernetes often want to use KEDA for autoscaling based on Temporal task queue depth metrics. Currently, the operator's replica management conflicts with KEDA's scaling decisions, causing a "fight" between the two controllers.

This change allows the operator to defer scaling responsibility to KEDA while still handling:
- Version management and rollouts
- Scale-to-zero for deprecated/sunset versions
- All other non-scaling lifecycle operations

 Checklist
1. Closes #N/A (new feature request)
2. How was this tested:
   - Code review of change paths
   - Verified annotation check is opt-in (no impact to existing users)
   - Logic review: only affects `ScaleDeployments` and `UpdateDeployments` replica handling when annotation present
   
3. Any docs updates needed?
   - README update recommended to document `temporal.io/keda-managed` annotation
   - Example KEDA ScaledObject configuration for Temporal task queues would be helpful

### DIFF
--- a/internal/controller/genplan.go
+++ b/internal/controller/genplan.go
@@ -150,6 +150,23 @@ func (r *TemporalWorkerDeploymentReconciler) generatePlan(
 	plan.ScaleDeployments = planResult.ScaleDeployments
 	plan.UpdateDeployments = planResult.UpdateDeployments
 
+	// If KEDA/HPA managed, only keep scale-to-zero operations (for sunset behavior).
+	if _, ok := w.GetAnnotations()["temporal.io/keda-managed"]; ok {
+		for ref, replicas := range plan.ScaleDeployments {
+			if replicas > 0 {
+				delete(plan.ScaleDeployments, ref)
+			}
+		}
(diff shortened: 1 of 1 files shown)


vpo-cdc:
I have an interrogation about who's responsible for creating ScaledObject. Isn't Worker Controller supposed to also do it because it creates the deployment with names based on the worker image version, so it is the only one knowing what the scaledobject targetref will be. What do you think ?

Issue #191, opened 2026-01-21
Automate validation of PR's changing the Helm chart.


Shivs11:
### Is your feature request related to a problem? Please describe.

- Automate the validation of helm charts on a PR. These validations can include basic syntax checking as well as actually rendering the helm chart itself.
- If there any errors in the automation, the PR author can be aware that their changes are buggy.

### Describe the solution you'd like
- Adding those validation pipelines as a Github CI action.

PR #199, opened 2026-02-17
Adding a CI action to validate helm chart renderings

Shivs11:
## What was changed
- WISOTT
- I have added some very basic "helm lint and helm template" validation checks in this PR. One could very well build on this by adding something like "is the image that is being pulled a valid docker image". I decided to keep it small and just get this in, for now, so that we have a pipeline to test things against.

## Why?
- to detect failures quite early

## Checklist

1. Closes 
#191 

3. How was this tested:

4. Any docs updates needed?

### DIFF
--- a/.github/workflows/helm-validate.yml
+++ b/.github/workflows/helm-validate.yml
@@ -17,8 +17,6 @@ jobs:
 
       - name: Install Helm
         uses: azure/setup-helm@v3
-        with:
-          version: 'v3.14.3'
 
       - name: Lint chart
         run: helm lint --strict helm/temporal-worker-controller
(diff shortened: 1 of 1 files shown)


.github/workflows/helm-validate.yml
```diff
@@ -0,0 +1,61 @@
...
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install Helm
+        uses: azure/setup-helm@v3
+
+      - name: Lint chart
+        run: helm lint --strict helm/temporal-worker-controller
+
+  helm-template:
```
Shivs11, line 24: the main intention here was to render our helm templates with different combinations so that we can catch changes to those manifests that are not included by default.

PR #203, opened 2026-02-21
Add event recording and status conditions for worker deployments

### LINKED ISSUE #28: Add events to the TemporalWorkerDeployment CRD when there is a problem

thearcticwatch:
## What changed: Added Kubernetes events and status conditions          
  (TemporalConnectionHealthy, RolloutReady) to the worker controller
  reconciliation loop.           
                                                                       
##Why: Reconciliation failures were only visible in controller logs —  
  events and conditions let users diagnose issues directly via kubectl.

1. Closes #28 

2. How was this tested:
added unit tests and functional tests

3. Any docs updates needed?
N/A

4. Is this risky? Explain

Making a change to the CRD (adding conditions) opens up the risk that users could upgrade the controller but fail to upgrade their CRD. In this case, it is ok if new features are silently ignored, but we don't want the controller to panic or fail to successfully do the actions that were available in the previous CRD version. I believe that this change is safe even if someone forgets to upgrade their CRD, because when this new controller runs against a v1.2.0 CRD:
  - No panic. The controller calls r.Status().Update(ctx, twd) with conditions populated in memory. The API server validates against the CRD schema and prunes unknown fields (standard behavior for structural schemas without x-kubernetes-preserve-unknown-fields). The status write succeeds with a 200 and the conditions are silently dropped before storage.
  - Kubernetes Events work fine. Events are written as separate events.k8s.io/v1 resources, completely independent of the TWD CRD schema. All r.Recorder.Eventf(...) calls will succeed normally.
  - Conditions simply don't persist. kubectl get twd foo -o yaml will show no conditions field. The controller sets them in memory on every reconcile, tries to write, and the API server drops them. Functionally the controller does the right thing, it just can't communicate the health status via conditions until the CRD is upgraded.

### DIFF
--- a/internal/planner/planner.go
+++ b/internal/planner/planner.go
@@ -354,8 +354,9 @@ func getDeleteDeployments(
 			// Deleting a deployment is only possible when:
 			// 1. The deployment has been drained for deleteDelay + scaledownDelay.
 			// 2. The deployment is scaled to 0 replicas.
-			if (time.Since(version.DrainedSince.Time) > spec.SunsetStrategy.DeleteDelay.Duration+spec.SunsetStrategy.ScaledownDelay.Duration) &&
-				*d.Spec.Replicas == 0 {
+			if version.DrainedSince != nil &&
+				(time.Since(version.DrainedSince.Time) > spec.SunsetStrategy.DeleteDelay.Duration+spec.SunsetStrategy.ScaledownDelay.Duration) &&
+				d.Spec.Replicas != nil && *d.Spec.Replicas == 0 {
 				deleteDeployments = append(deleteDeployments, d)
 			}
 		case temporaliov1alpha1.VersionStatusNotRegistered:
@@ -428,7 +429,7 @@ func getScaleDeployments(
 				scaleDeployments[version.Deployment] = uint32(replicas)
 			}
 		case temporaliov1alpha1.VersionStatusDrained:
-			if time.Since(version.DrainedSince.Time) > spec.SunsetStrategy.ScaledownDelay.Duration {
+			if version.DrainedSince != nil && time.Since(version.DrainedSince.Time) > spec.SunsetStrategy.ScaledownDelay.Duration {
 				// TODO(jlegrone): Compute scale based on load? Or percentage of replicas?
 				// Scale down drained deployments after delay
 				if d.Spec.Replicas != nil && *d.Spec.Replicas != 0 {


internal/controller/worker_controller.go
```diff
@@ -235,14 +265,24 @@ func (r *TemporalWorkerDeploymentReconciler) Reconcile(ctx context.Context, req
...
 	// Execute the plan, handling any errors
-	if err := r.executePlan(ctx, l, temporalClient, plan); err != nil {
+	if err := r.executePlan(ctx, l, &workerDeploy, temporalClient, plan); err != nil {
+		r.Recorder.Eventf(&workerDeploy, corev1.EventTypeWarning, "PlanExecutionFailed",
+			"Unable to execute reconciliation plan: %v", err)
 		return ctrl.Result{}, err
 	}
 
+	// Mark as Ready on successful reconciliation
+	if workerDeploy.Status.TargetVersion.BuildID == workerDeploy.Status.CurrentVersion.BuildID {
```
carlydf: I think this is causing a nil pointer error in the integration test. Because Current Version is initially nil.

internal/controller/execplan.go
```diff
@@ -129,51 +141,83 @@ func (r *TemporalWorkerDeploymentReconciler) executePlan(ctx context.Context, l
 			_, err = temporalClient.ExecuteWorkflow(ctx, opts, wf.workflowType)
 		}
 		if err != nil {
+			r.Recorder.Eventf(workerDeploy, corev1.EventTypeWarning, "TestWorkflowStartFailed",
+				"Failed to start gate workflow %q (buildID %s): %v", wf.workflowType, wf.buildID, err)
```
carlydf: can you put the task queue name in this event?

internal/controller/execplan.go
```diff
@@ -129,51 +141,83 @@ func (r *TemporalWorkerDeploymentReconciler) executePlan(ctx context.Context, l
...
+		return err
+	}
+
+	deploymentHandler := temporalClient.WorkerDeploymentClient().GetHandle(p.WorkerDeploymentName)
+
+	if err := r.startTestWorkflows(ctx, l, workerDeploy, temporalClient, p); err != nil {
+		return err
+	}
+
+	if err := r.updateVersionConfig(ctx, l, workerDeploy, deploymentHandler, p); err != nil {
```
carlydf, line 225: was this refactor just to make the code easier to understand, or was it necessary for the events change? (I'm all for it, just curious)

internal/controller/worker_controller.go
```diff
@@ -235,14 +265,25 @@ func (r *TemporalWorkerDeploymentReconciler) Reconcile(ctx context.Context, req
...
+		r.Recorder.Eventf(&workerDeploy, corev1.EventTypeWarning, "PlanExecutionFailed",
+			"Unable to execute reconciliation plan: %v", err)
 		return ctrl.Result{}, err
 	}
 
+	// Mark as Ready on successful reconciliation
+	if workerDeploy.Status.CurrentVersion != nil &&
+		workerDeploy.Status.TargetVersion.BuildID == workerDeploy.Status.CurrentVersion.BuildID {
+		r.setCondition(&workerDeploy, temporaliov1alpha1.ConditionRolloutReady, metav1.ConditionTrue,
+			"RolloutSucceeded", "Target version rollout complete "+workerDeploy.Status.TargetVersion.BuildID)
```
carlydf: It feels inconsistent to me to be saying "rollout ready", "rollout succeeded", and "rollout complete" all together. Maybe just "rollout complete" for all of them?

Also, I just realized, the condition as written will cause us to emit a `ConditionRolloutReady` event at the end of the first reconcile loop that meets this condition (which is correct), but also at the end of every subsequent reconcile loop, which mean we will emit the same event every 30s.

I think if we instead emit this event in the actual `updateVersionConfig` function where we successfully set the Current Version, that would ensure that we only emit it once per completed rollout

internal/controller/worker_controller.go
```diff
@@ -96,6 +99,7 @@ type TemporalWorkerDeploymentReconciler struct {
 //+kubebuilder:rbac:groups=core,resources=secrets,verbs=get;list;watch
 //+kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=apps,resources=deployments/scale,verbs=update
+// +kubebuilder:rbac:groups="",resources=events,verbs=create;patch
```
carlydf: https://book.kubebuilder.io/reference/raising-events#granting-the-required-permissions says that this would be enough:
```
// +kubebuilder:rbac:groups=events.k8s.io,resources=events,verbs=create;patch
```
so I think `kubebuilder:rbac:groups=""` may be overly permissive?

carlydf (commented):
looking good! just did initial review, we should still add a functional test once these comments are addressed.

I found https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#events and https://book.kubebuilder.io/reference/raising-events#creating-events helpful while reviewing.

internal/controller/worker_controller.go
```diff
@@ -146,16 +150,30 @@ func (r *TemporalWorkerDeploymentReconciler) Reconcile(ctx context.Context, req
 		Namespace: workerDeploy.Namespace,
 	}, &temporalConnection); err != nil {
 		l.Error(err, "unable to fetch TemporalConnection")
+		r.Recorder.Eventf(&workerDeploy, corev1.EventTypeWarning, "TemporalConnectionNotFound",
+			"Unable to fetch TemporalConnection %q: %v", workerDeploy.Spec.WorkerOptions.TemporalConnectionRef.Name, err)
+		r.setCondition(&workerDeploy, temporaliov1alpha1.ConditionTemporalConnectionHealthy, metav1.ConditionFalse,
+			"TemporalConnectionNotFound", fmt.Sprintf("TemporalConnection %q not found: %v", workerDeploy.Spec.WorkerOptions.TemporalConnectionRef.Name, err))
+		_ = r.Status().Update(ctx, &workerDeploy)
```
Shivs11: seems like this pattern of setting and recording events is repeated throughout the codebase; think we should clean this up by having a nice helper which could also make it look neater

wdyt

internal/controller/execplan.go
```diff
@@ -11,22 +11,26 @@ import (
...
 	"go.temporal.io/sdk/worker"
 	appsv1 "k8s.io/api/apps/v1"
 	autoscalingv1 "k8s.io/api/autoscaling/v1"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func (r *TemporalWorkerDeploymentReconciler) executePlan(ctx context.Context, l logr.Logger, temporalClient sdkclient.Client, p *plan) error {
+func (r *TemporalWorkerDeploymentReconciler) executeK8sOperations(ctx context.Context, l logr.Logger, workerDeploy *temporaliov1alpha1.TemporalWorkerDeployment, p *plan) error {
```
Shivs11, line 26: any specific reason to rename this? i quite liked the approach of having these three files/steps where we first generate a status, make a plan and then act on it.

internal/controller/execplan.go
```diff
@@ -129,51 +141,85 @@ func (r *TemporalWorkerDeploymentReconciler) executePlan(ctx context.Context, l
...
+	}
+
+	if vcfg.SetCurrent {
+		l.Info("registering new current version", "buildID", vcfg.BuildID)
+		if _, err := deploymentHandler.SetCurrentVersion(ctx, sdkclient.WorkerDeploymentSetCurrentVersionOptions{
+			BuildID:       vcfg.BuildID,
+			ConflictToken: vcfg.ConflictToken,
+			Identity:      getControllerIdentity(),
+		}); err != nil {
+			r.Recorder.Eventf(workerDeploy, corev1.EventTypeWarning, "VersionRegistrationFailed",
```
Shivs11: we should also log the error here, in addition to recording the event here imo. We have followed this elsewhere in the same file and in the code, and I think it could be good practice given that if we do have a persistence failure, the logs would still be an option for an operator.

internal/controller/worker_controller.go
```diff
@@ -235,11 +265,15 @@ func (r *TemporalWorkerDeploymentReconciler) Reconcile(ctx context.Context, req
...
+		r.Recorder.Eventf(&workerDeploy, corev1.EventTypeWarning, "PlanGenerationFailed",
+			"Unable to generate reconciliation plan: %v", err)
 		return ctrl.Result{}, err
 	}
 
 	// Execute the plan, handling any errors
-	if err := r.executePlan(ctx, l, temporalClient, plan); err != nil {
+	if err := r.executePlan(ctx, l, &workerDeploy, temporalClient, plan); err != nil {
+		r.Recorder.Eventf(&workerDeploy, corev1.EventTypeWarning, "PlanExecutionFailed",
+			"Unable to execute reconciliation plan: %v", err)
```
Shivs11: any specific reason you chose to not persist failures for plan generation and execution phases but persisted when we do generate the status from the server?

internal/controller/execplan.go
thearcticwatch: Yes, the tets were failing for Cognitive Complexity and the refactor fixed them. Also just a bonus refactor.

internal/controller/worker_controller.go
carlydf: yes, but I wouldn't block this PR on that personally

internal/controller/execplan.go
carlydf: `func (r *TemporalWorkerDeploymentReconciler) executePlan(ctx context.Context, l logr.Logger, workerDeploy *temporaliov1alpha1.TemporalWorkerDeployment, temporalClient sdkclient.Client, p *plan) error {` still exists (see below), it now just calls multiple helper functions in sequence, starting with `executeK8sOperations`. Apparently this refactor gets rid of the cognitive complexity linter error.

internal/controller/execplan.go
carlydf: We are logging the error on line 166, no?

internal/controller/execplan.go
```diff
@@ -129,51 +141,89 @@ func (r *TemporalWorkerDeploymentReconciler) executePlan(ctx context.Context, l
...
-				return fmt.Errorf("unable to set ramping deployment version: %w", err)
-			}
+		if _, err := deploymentHandler.SetRampingVersion(ctx, sdkclient.WorkerDeploymentSetRampingVersionOptions{
+			BuildID:       vcfg.BuildID,
+			Percentage:    float32(vcfg.RampPercentage),
+			ConflictToken: vcfg.ConflictToken,
+			Identity:      getControllerIdentity(),
+		}); err != nil {
+			l.Error(err, "unable to set ramping deployment version", "buildID", vcfg.BuildID, "percentage", vcfg.RampPercentage)
+			r.Recorder.Eventf(workerDeploy, corev1.EventTypeWarning, "VersionRegistrationFailed",
```
carlydf: sorry to nitpick, but "version registration" sounds to me like what happens when a versioned poller first arrives, polling the server and creating a version record. Then, you could say the version has been "registered" because it exists. Anyway, "VersionRoutingConfigUpdateFailed" or "VersionPromotionFailed" or "RoutingConfigUpdateFailed" makes more sense to me. But open to hearing reasons for "VersionRegistrationFailed" being better! And I also don't want to "bikeshed" naming, but at the same time want to choose names that reinforce existing terminology that we use elsewhere in the versioning world (ie. "routing config" or "version promotion").

The other note I have is, we should consider making these event strings:
- `"VersionRegistrationFailed"`
- `"MetadataUpdateFailed"`
- `"TestWorkflowStartFailed"`
- `"DeploymentXXXFailed"`
- `"PlanXXXFailed"`
constants so that it's easy to see a list of all the possible event types in one place, and so that in the cases where we emit the same event type in multiple places, there is no risk of spelling errors.

carlydf: tests can also use the constants to verify expected content, instead of also re-typing the strings there

internal/controller/worker_controller.go
```diff
@@ -146,16 +150,28 @@ func (r *TemporalWorkerDeploymentReconciler) Reconcile(ctx context.Context, req
 		Namespace: workerDeploy.Namespace,
 	}, &temporalConnection); err != nil {
 		l.Error(err, "unable to fetch TemporalConnection")
+		r.recordWarningAndSetCondition(ctx, &workerDeploy, temporaliov1alpha1.ConditionTemporalConnectionHealthy,
+			"TemporalConnectionNotFound",
```
carlydf: similar note as above. I'd recommend using constants for these conditions: `"TemporalConnectionNotFound"`, `"AuthSecretInvalid"`, `"TemporalConnectionHealthy"`, `"TemporalClientCreationFailed"`, `"TemporalStateFetchFailed"`.

What I see a lot in the Go SDK which I look to as a well-designed code base is if a string value is actually a strict subset of valid strings, make a special type for it, and then use that type to define the acceptable constants as well as the input values of functions. I can check out this branch and push a commit to show what I mean, if that's cool with you!

internal/controller/worker_controller.go
carlydf: I standardized this all to be "RolloutComplete" in the changes I pushed. Leaving comment open for visibility

internal/controller/worker_controller.go
carlydf: do you mean, why just record an event and not also set a condition?
The only conditions we have right now are `ConditionTemporalConnectionHealthy` and `ConditionRolloutComplete`. `ConditionRolloutComplete` is set in execplan right after the SetCurrent call succeeds (so that we don't do it on every iteration).

Ah actually, I'm not sure if doing 
```
	// Mark TemporalConnection as valid since we fetched it and resolved auth
	r.setCondition(&workerDeploy, temporaliov1alpha1.ConditionTemporalConnectionHealthy, metav1.ConditionTrue,
		temporaliov1alpha1.ReasonTemporalConnectionHealthy, "TemporalConnection is healthy and auth secret is resolved")
```
unconditionally on every iteration is good or bad. Will ask Claude

internal/controller/worker_controller.go
```diff
@@ -275,6 +275,9 @@ func (r *TemporalWorkerDeploymentReconciler) Reconcile(ctx context.Context, req
 		return ctrl.Result{}, err
 	}
 
+	// Persist any conditions set during plan execution (e.g. RolloutComplete).
+	_ = r.Status().Update(ctx, &workerDeploy)
```
carlydf: this was added because without it, the `RolloutComplete` condition was never being written back to the status, but it's wasteful, so I'm fixing that now

internal/controller/worker_controller.go
carlydf: > Also, I just realized, the condition as written will cause us to emit a ConditionRolloutReady event at the end of the first reconcile loop that meets this condition (which is correct), but also at the end of every subsequent reconcile loop, which mean we will emit the same event every 30s.

> I think if we instead emit this event in the actual updateVersionConfig function where we successfully set the Current Version, that would ensure that we only emit it once per completed rollout

@thearcticwatch When I said the above, I was wrong about where the "write" happens. I thought moving `setCondition` within `executePlan` was meaningful for reducing API writes, but `setCondition` only modifies an in-memory Go struct and makes no network call at all. The actual write to the Kubernetes API server happens exclusively when `r.Status().Update()` is called. So what matters for write frequency isn't where `setCondition` is placed, but how many times `Update()` is called per reconcile. The real fix was to consolidate from two `Status().Update()` calls per successful reconcile down to one, by removing the intermediate write after `generateStatus` and doing a single write at the very end of the loop after `executePlan` completes.

internal/controller/worker_controller.go
carlydf: turns out, `setCondition` doesn't do the write, `r.Status().Update()` does, so my concern about setting `ConditionTemporalConnectionHealthy` on every iteration being potentially wasteful was wrong

internal/controller/worker_controller.go
carlydf: addressed in [standardize event/condition reason constants and split by API stability](https://github.com/temporalio/temporal-worker-controller/pull/203/commits/6e92e3a56bfc95f8dab14c369184d709a1f52420)

internal/controller/execplan.go
carlydf: addressed in [standardize event/condition reason constants and split by API stability](https://github.com/temporalio/temporal-worker-controller/pull/203/commits/6e92e3a56bfc95f8dab14c369184d709a1f52420)

internal/controller/worker_controller.go
carlydf: addressed in [only write to status once per reconcile loop](https://github.com/temporalio/temporal-worker-controller/pull/203/commits/371a02fcec7df92507f0efc69d7b9eb0a7c082d8)

internal/controller/worker_controller.go
```diff
@@ -146,16 +150,30 @@ func (r *TemporalWorkerDeploymentReconciler) Reconcile(ctx context.Context, req
...
+			fmt.Sprintf("Unable to fetch TemporalConnection %q: %v", workerDeploy.Spec.WorkerOptions.TemporalConnectionRef.Name, err),
+			fmt.Sprintf("TemporalConnection %q not found: %v", workerDeploy.Spec.WorkerOptions.TemporalConnectionRef.Name, err))
 		return ctrl.Result{}, err
 	}
 
 	// Get the Auth Mode and Secret Name
 	authMode, secretName, err := resolveAuthSecretName(&temporalConnection)
 	if err != nil {
+		// Note: as things are now, this will never happen, because getAPIKeySecretName only errors when secretRef == nil,
+		// but resolveAuthSecretName only calls it inside the tc.Spec.APIKeySecretRef != nil branch
```
carlydf: we should decide what to do here. We can refactor `resolveAuthSecretName` to not return an error, and then completely get rid of `ReasonAuthSecretInvalid`, or we could just leave it "just in case". 

One issue is that because `temporaliov1alpha1.ReasonAuthSecretInvalid` is in the public API, even if it never happens, it could be weird or feel risky to remove later. Because of that, refactoring this stack now is probably best.

One idea for the refactor: instead of refactoring to never error, we could require people to opt-in to "No Auth Mode." Then, if people don't put any auth mode in, there would be an error that we would surface here. Tbh, that's probably the safest solution.

carlydf: changed to return an error when secretRef is non nil but has an empty secret name. Probably won't happen, but if it does we should error.

carlydf (approved):
Approved, but @Shivs11 if you could take a peek at my refactor of ClientPool which I did so we could emit a separate event type for invalid secret vs failed dial to Temporal server, that would be great!

PR #204, opened 2026-02-21
Shorten deployment names when over 47 characters

carlydf:
## What was changed
Shorten deployment names (while keeping them unique) if over 47 characters.

## Why?
Deployment names in k8s should be <=47 characters so that Pod names can be short enough for DNS limits.

## Checklist

1. Closes 

2. How was this tested:
New unit tests, and existing functional tests. 

3. Any docs updates needed?

### DIFF
--- a/docs/limits.md
+++ b/docs/limits.md
@@ -11,5 +11,6 @@ Kubernetes operational best practice is to limit Deployment names to 47 characte
 The controller will first generate the Deployment name for each Version as follows: `<TemporalWorkerDeployment.Name>-<BuildID>`.
 If that name exceeds 47 characters, the controller will generate a unique less-than-47-character name for the Deployment.
 You can find the names of all the Deployment resources owned by a certain `TemporalWorkerDeployment` in the `Status` field of the `TemporalWorkerDeployment` resource.
+Note that the original untruncated deployment name and build ID can always be found in the `temporal.io/deployment-name` and `temporal.io/build-id` labels.
 
 For transparency, know that the controller will generate these unique short names as follows, but don't depend on this name format because it may change: `trunc(<TemporalWorkerDeployment.Name>)-trunc(<BuildID>)-hash(<TemporalWorkerDeployment.Name>-<BuildID>)`


internal/k8s/deployments.go
```diff
@@ -142,8 +143,20 @@ func ComputeWorkerDeploymentName(w *temporaliov1alpha1.TemporalWorkerDeployment)
 }
 
 // ComputeVersionedDeploymentName generates a name for a versioned deployment
+// Name will be <=47 characters and unique for that Worker Deployment Version within the namespace.
 func ComputeVersionedDeploymentName(baseName, buildID string) string {
```
carlydf: see kubebuilder library for shortening k8s resource names

internal/tests/internal/env_helpers.go
```diff
@@ -175,6 +177,7 @@ func setupTestEnvironment(t *testing.T) (*rest.Config, client.Client, manager.Ma
 	// Return cleanup function
 	cleanup := func() {
 		cancel()
+		<-managerStopped
```
carlydf, line 180: this fixes a race during test environment cleanup that would cause the test to time out even after all subtests passed

docs/limits.md
```diff
@@ -1,4 +1,15 @@
...
+- Max `BuildID` length: 63 characters, since Kubernetes label values are limited to 63 characters
+
+### Note on naming of controller-generated resources:
+The controller generates a Kubernetes Deployment resource for each Version in your Worker Deployment.
+
+Kubernetes operational best practice is to limit Deployment names to 47 characters to ensure that full Pod names are always less than 63 characters.
+
+The controller will first generate the Deployment name for each Version as follows: `<TemporalWorkerDeployment.Name>-<BuildID>`.
+If that name exceeds 47 characters, the controller will generate a unique less-than-47-character name for the Deployment.
+You can find the names of all the Deployment resources owned by a certain `TemporalWorkerDeployment` in the `Status` field of the `TemporalWorkerDeployment` resource.
```
jaypipes, line 13: Might want to add the following:

> Note that the original untruncated deployment name and build ID can always be found in the `temporal.io/deployment-name` and `temporal.io/build-id` labels.

PR #205, opened 2026-02-23
Add support to watchNamespace and namespace-scoped rbac for Temporal controller

R1-lav:
# **Add support to watchNamespace**

## Summary

Adds the support to namespace-scoped deployment for the controller manager for environments where cluster wide permissions are not allowed.## Summary

This PR adds support for watching multiple namespaces in the controller via the `--watch-namespace` flag (or `WATCH_NAMESPACE` environment variable).

The flag now accepts a comma-separated list of namespaces, allowing the controller to operate in multi-namespace scoped mode.

## How it works

- If `--watch-namespace` (or `WATCH_NAMESPACE`) is empty, the controller watches all namespaces (default behavior).
- If a single namespace is provided, the controller runs in single-namespace scoped mode.
- If multiple namespaces are provided as a comma-separated list, the controller configures the manager cache to watch only those namespaces.

Example:

```bash
--watch-namespace=ns1,ns2,ns3
```

## Other changes

Support of ownNamespace in helm values to support creating roles and rolebinding instead of clusterrole and clusterrolebinding for namespace-scoped deployments


carlydf (commented):
Thanks for this, I think namespace-scoped operation is a useful feature. Before we discuss implementation details, it would help to know which use case you're actually targeting:

  1. Release namespace only: the controller always watches the same namespace it's deployed in (rbac.ownNamespace + watchNamespace both implicitly
  refer to the release namespace)
  2. Single arbitrary namespace: the controller is deployed in namespace A but watches namespace B
  3. Multiple arbitrary namespaces: watchNamespace: "ns1,ns2" with the controller in a third namespace

The reason this matters is that `rbac.ownNamespace` and `watchNamespace` are currently independent knobs, but their interaction has sharp edges depending on the answer:

  - If the watched namespace(s) differ from the release namespace and `rbac.ownNamespace: true`, the controller SA has no RBAC in the namespaces it's actually watching. It will silently fail to act there.
  - If `watchNamespace: "ns1,ns2"` with `rbac.ownNamespace: true`, a single Role in the release namespace can't cover ns1 or ns2 — you'd need a ClusterRole or separate Roles per namespace.

If the scenario you're trying to support is (1), the implementation could be simplified significantly and the two flags could be collapsed into one. If the answer is (2) or (3), the RBAC side needs more thought before we add tests or docs.

carlydf:
Hi @R1-lav! Circling back here, sorry for the delay.

I think my preference would be for a simpler helm chart, so I would prefer one knob to configure this + minimize how people can unexpectedly misconfigure. 

I think that could mean just leaving the `ownNamespace` flag in rbac options, and if that is true, pass the controller release namespace to the watchNamespaces flag in cmd/main.go. That is simple and solves your issue now, but it could be a little messy to expand to multi-namespace option later.

Another idea is to put a `restrictWatchNamespaces` list in the rbac section of the helm values. That will help make it clear that which namespaces are watched are tightly intertwined with RBAC. Then, we could have a slightly more complicated helm chart which creates watch Roles for each namespace in that list. (Or something along those lines, I'm not dictating the exact interface here, I just want there to be synchronization between watchNamespaces and the Roles).

I think I prefer the second option -- what do you think?

carlydf:
See how `allowedResources` is one helm value that is used both for code behavior and for permissions so that both are in sync https://github.com/temporalio/temporal-worker-controller/blob/main/helm/temporal-worker-controller/values.yaml#L98

jaypipes (approved):
Very nice, I like this approach quite a bit. thank you @R1-lav!

ndouchin:
Great to hear you like those changes @jaypipes.
Is everything all good on your side? Do you need something more from us on this PR? :)

jaypipes:
> Great to hear you like those changes @jaypipes. Is everything all good on your side? Do you need something more from us on this PR? :)

@ndouchin Nothing from me, no. I'd like @carlydf to give this another review though :)

carlydf (commented):
Agree this approach looks great! One note on testing: have you been able to test locally that this helm chart renders correctly with the new value set? Our CI pipelines check that the helm charts render correctly with the default values, but they don't test opt-in values. 

In the long run we should consider changing that (like maybe provide a few other values files with opt-in options to test with), but for this PR, a sanity check would be enough for me.

ndouchin:
Thanks @carlydf!

Yes, sanity-checked locally with `helm template` and the new value set: it renders cleanly with `rbac.restrictWatchNamespaces` set to both a single namespace and multiple namespaces.

The watch scope (`WATCH_NAMESPACES` on the manager) stays in sync with the per-namespace RoleBindings that get generated, and the two cluster-wide grants (`namespaces:get` + `subjectaccessreviews:create`) render as expected.

I can also add a render step for this value to the `helm-template` CI job — mirroring the existing `namespace.create` / `metrics.disableAuth` steps.

aldredb:
Thanks for the merge. May I know when is the next release so that we can use this feature?

PR #208, opened 2026-02-26
Separate CRDs Helm chart for upgradeable CRD lifecycle

carlydf:
## Summary

- Creates a new `temporal-worker-controller-crds` Helm chart with CRDs in `templates/` (not `crds/`) so they are upgraded on `helm upgrade`
- Removes `crds/` from the main chart; CRDs now live exclusively in the CRDs chart
- Updates `make manifests` to output to the new chart's `templates/`
- Updates `make install`, `make uninstall`, and `make deploy` to use the new paths
- Adds `helm-lint-crds` and `helm-template-crds` jobs to `helm-validate.yml` CI gate
- Updates `release.yml` and `helm.yml` to bump both chart versions and push both charts on release
- Adds `docs/crd-management.md` with compatibility commitment, install/upgrade/rollback instructions, and a migration guide for existing users
- Updates `README.md` installation section to document the two-chart install flow

## Upgrade order

```
# Install / upgrade CRDs first, then the controller
helm upgrade temporal-worker-controller-crds ...
helm upgrade temporal-worker-controller ...
```

## Rollback order

```
# Roll back controller first, CRDs optionally afterward
helm rollback temporal-worker-controller
helm rollback temporal-worker-controller-crds  # usually not needed
```

## Migration for existing users (Controller Helm Chart v0.12.0 and earlier)

Helm does not delete CRDs when upgrading to the new chart, so upgrading the main chart is safe. A one-time `helm install temporal-worker-controller-crds` adopts the existing CRDs into the new release.

## Test plan

- [x] `helm lint --strict helm/temporal-worker-controller-crds` passes
- [x] `helm template test-release helm/temporal-worker-controller-crds` emits two CRDs with `helm.sh/resource-policy: keep`
- [x] `helm template test-release helm/temporal-worker-controller` emits zero CRDs
- [x] CI `helm-validate.yml` passes all 4 jobs (lint, template, lint-crds, template-crds)

### DIFF
--- a/docs/crd-management.md
+++ b/docs/crd-management.md
@@ -10,7 +10,7 @@ Benefits:
 - CRDs and the controller can be upgraded and rolled back independently
 - Clear versioning: both charts always use the same version number
 - CRDs won't be accidentally uninstalled via `helm uninstall` of `temporal-worker-controller` chart
-- CRDs can be uninstalled separately with `helm uninstall temporal-worker-controller-crds`
+- CRDs can be uninstalled separately with `helm uninstall temporal-worker-controller-crds`. WARNING: Uninstalling CRDs will delete all Custom Resources in your cluster that use those CRDs.
 
 ## Compatibility Commitment
 


docs/crd-management.md
```diff
@@ -0,0 +1,147 @@
+# CRD Management
+
+## Why a Separate CRDs Chart?
+
+Helm's `crds/` directory installs CRDs on `helm install` but **silently ignores them on `helm upgrade`** and does not delete them on `helm uninstall`. This means users have no supported mechanism to upgrade CRDs when upgrading the controller chart via the standard Helm workflow.
+
+To provide an explicit, version-tracked CRD upgrade path, the temporal-worker-controller ships CRDs as a separate Helm chart: `temporal-worker-controller-crds`. This is the same pattern used by [Karpenter](https://karpenter.sh/docs/upgrading/upgrade-guide/) (`karpenter-crd`), [prometheus-operator-crds](https://github.com/prometheus-community/helm-charts/tree/main/charts/prometheus-operator-crds), and [VictoriaMetrics](https://docs.victoriametrics.com/helm/victoria-metrics-operator-crds/) (`victoria-metrics-operator-crds`).
```
Shivs11: nit: personally never heard of VictoriaMetrics, and not sure it is needed to have this bit of information here as Prometheus is a strong enough example imo

helm/temporal-worker-controller-crds/Chart.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: v2
+name: temporal-worker-controller-crds
+description: CRDs for the Temporal Worker Controller. Install this chart before the temporal-worker-controller chart.
```
Shivs11, line 3: if we do independently version our worker-controller and the corresponding crd's, it's quite easy for both of them to drift apart imho. There shouldn't be a world where we have the user doing math and looking at a combinatorial matrix to understand which crd version is compatible with which controller version.

wdyt about the idea of adding to the ci a validation step which shall check/enforce this?

docs/crd-management.md
```diff
@@ -0,0 +1,147 @@
+# CRD Management
```
Shivs11, line 1: actually, just while reading this I began to wonder if we really do need this (maybe we do, and would be happy to be corrected). It seems to me that the main problem is that on a helm upgrade, CRD's don't upgrade and this could be annoying to users.

However, what if we just moved the crds into the helm chart of the controller itself? in that way, the helm upgrade command would work and we don't need to create a new chart too

Shivs11: you could argue for us to actually then just version them independently, but I am struggling to think about why one would require independent versioning if there is a higher chance of these version mismatches happening down the line

carlydf: When someone rolls back the controller, in most cases, the safest thing is to not roll back the CRD. If the CRD is in the main helm chart itself, then rolling back would roll back the CRD definitions, and cause field loss.

docs/crd-management.md
carlydf: removed!

docs/crd-management.md
```diff
@@ -0,0 +1,147 @@
...
+
+## Why a Separate CRDs Chart?
+
+Helm's `crds/` directory installs CRDs on `helm install` but **silently ignores them on `helm upgrade`** and does not delete them on `helm uninstall`. This means users have no supported mechanism to upgrade CRDs when upgrading the controller chart via the standard Helm workflow.
+
+To provide an explicit, version-tracked CRD upgrade path, the temporal-worker-controller ships CRDs as a separate Helm chart: `temporal-worker-controller-crds`. This is the same pattern used by [Karpenter](https://karpenter.sh/docs/upgrading/upgrade-guide/) (`karpenter-crd`) and [prometheus-operator-crds](https://github.com/prometheus-community/helm-charts/tree/main/charts/prometheus-operator-crds).
+
+Benefits:
+- CRDs survive accidental `helm uninstall` of either chart (`helm.sh/resource-policy: keep`)
```
jaypipes: This seems at odds with the sentence on line 5 that says:

> and does not delete them on `helm uninstall`

I actually don't think the `helm.sh/resource-policy: keep` thing is a good idea. If someone runs `helm uninstall`, they are doing it on purpose and likely want to clean up their installation of something entirely. Leaving the (perhaps old-versioned) CRDs installed in the Kubernetes cluster is not the behaviour expected when an operator runs `helm uninstall`, IMHO...

docs/crd-management.md
```diff
@@ -0,0 +1,147 @@
...
+- CRDs and the controller can be upgraded and rolled back independently
+- Clear versioning: both charts always use the same version number
+
+## Compatibility Commitment
+
+> **CRD chart version N is forward-compatible with controller chart versions N and N−1.**
```
jaypipes, line 17: is the reason that TWC's Helm chart version is currently on a `v0` release series because the CRD API version is `v1alpha1`? Are we planning on coordinating a `v1` Helm Chart release series when moving the CRD API version to `v1`?

docs/crd-management.md
```diff
@@ -0,0 +1,147 @@
...
+
+| Scenario | CRDs | Controller | Expected outcome |
+|---|---|---|---|
+| Normal upgrade (recommended order) | N → N+1 first | N → N+1 after | No issues |
+| CRDs ahead, controller not yet upgraded | N+1 | N | Safe: new fields are optional, so the older controller ignores unknown fields in reads and doesn't include them in its writes (which is acceptable since they're optional) |
+| Wrong order: controller upgraded first | N | N+1 (CRDs not yet upgraded) | Silent feature degradation: new status fields the N+1 controller writes are pruned by the API server (schema still at N). All existing CRs continue reconciling normally. New features depending on those fields don't work until CRDs are upgraded. No crashes, no per-CR errors. |
+| Controller rollback (CRDs stay at N+1) | N+1 | N+1 → N | Safe for spec (the controller never writes TWD spec). Status fields added in N+1 will be absent after the first reconciliation cycle — this is acceptable because status is fully reconstituted from live state, not preserved across reconciles. |
+| CRD rollback | N+1 → N | N | See [CRD rollback and field pruning](#crd-rollback-and-field-pruning) — this is the dangerous case. |
```
jaypipes: LOL, I think this is AI improperly using a Markdown table for something that isn't really a tabular concept. I would recommend removing the above section entirely (I think it's more confusing to the average reader than anything else...) and just sticking with the `## Upgrading` instructions below.

helm/temporal-worker-controller-crds/Chart.yaml
```diff
@@ -0,0 +1,5 @@
...
+type: application
+version: 0.12.0
```
jlegrone, line 5: Maybe we could align this with release numbers from the repo?

helm/temporal-worker-controller-crds/Chart.yaml
jlegrone, line 1: Can we add a readme with a pointer to the crd-management doc, and also guidance for contributors to NEVER add helm values (CRDs shouldn't really be configurable).

jlegrone (approved):

helm/temporal-worker-controller-crds/Chart.yaml
jaypipes: I don't think you would want to do that @jlegrone because the app version releases are already in the v1 semver release series. If you move the APIVersion for the CRDs from v1alpha1 to v1 you would want to bump the major release series which would mean a CRD APIVersion of v1 and an appVersion release series of v2. At least the way it is now, the v0 chartVersion matches the v0 (v1alpha1...) APIVersion "series". If you want to move from v1alpha1 to v1 APIVersion, you'd bump the chartVersion from v0 to v1 series and continue to be "aligned" from an APIVersion and ChartVersion perspective.

docs/crd-management.md
carlydf: done!

docs/crd-management.md
carlydf: We haven't discussed when to move the CRD API version to `v1` or when to move the Helm Chart from `v0` to `v1`

docs/crd-management.md
carlydf: removed. I think in table form it helped me think through things, but agree it's a lot of text for a new reader, so removing is good. We explain the concept in other places already.

helm/temporal-worker-controller-crds/Chart.yaml
carlydf: done

PR #217, opened 2026-03-06
Enable Controller-managed versioned scaling resources with `WorkerResourceTemplate`

### LINKED ISSUE #207: Enable CRUD of controller-managed scaling objects (and other custom scalers)
more details to come

carlydf:
### New CRD: `WorkerResourceTemplate` (WRT)

A new `WorkerResourceTemplate` CRD that lets users attach arbitrary namespaced Kubernetes resources (HPAs, PDBs, custom scalers, etc.) to a `TemporalWorkerDeployment`. The controller creates one copy of the resource per active worker version, with auto-injection of `scaleTargetRef`, `selector.matchLabels`, and metric selector labels to point at the correct versioned Deployment.

#### Key behaviors:
  - One copy per active Build ID, named `{twdName}-{wrtName}-{buildID}` (uniquely truncated to 47 chars, DNS-safe)
  - Auto-injects `spec.scaleTargetRef` (when set to `{}`) to reference the versioned Deployment → enables per-version HPA autoscaling, and any other scaler that uses `scaleTargetRef`
  - Auto-injects `selector.matchLabels` (when set to `{}`) with the correct per-version labels → enables per-version PDB targeting, and arbitrary CRDs that use `selector.matchLabels` to target versioned Deployments
  - Auto-appends `worker_deployment_name`, `worker_deployment_build_id`, and `temporal_namespace` to `spec.metrics[*].external.metric.selector.matchLabels` whenever `matchLabels` is present (including `{}`). User labels like `task_type: "Activity"` coexist alongside the injected keys. Absent `matchLabels` = no injection for that metric entry.
  - Applied via Server-Side Apply with field manager `"temporal-worker-controller"`
  - Owner ref on each resource copy points to the `WorkerResourceTemplate` → k8s GC deletes all copies when the WRT is deleted
  - Apply status written back to `WorkerResourceTemplate.status.versions[*]` (Applied, Message, BuildID)
  - Resource spec lives in `spec.template` (raw JSON/YAML embedded object)
  - Target TWD referenced via `spec.temporalWorkerDeploymentRef.name`

#### Validating Webhook

A `WorkerResourceTemplateValidator` webhook enforces:
  - `apiVersion` and `kind` required; `metadata.name`/`metadata.namespace` forbidden (controller sets these)
  - Allowed resource kinds configurable via `ALLOWED_KINDS` env var (default: `HorizontalPodAutoscaler`)
  - `minReplicas` ≠ 0 (currently required for `approximate_task_queue_backlog` metric-based autoscaling to work when queue is idle, plan to relax this in future)
  - `scaleTargetRef` must be absent or `{}` (opt-in sentinel); non-empty value rejected (controller owns injection)
  - `selector.matchLabels` must be absent or `{}` (opt-in sentinel); non-empty value rejected (controller owns injection)
  - `metrics[*].external.metric.selector.matchLabels` must not contain the controller-owned keys `worker_deployment_name`, `worker_deployment_build_id`, or `temporal_namespace`; user labels (e.g. `task_type`) are allowed
  - SAR check: requesting user must be able to create/update the embedded resource type
  - SAR check: controller service account must be able to create/update the embedded resource type
  - `spec.temporalWorkerDeploymentRef.name` is immutable after creation

### Helm chart updates

  - `helm/temporal-worker-controller-crds/templates/temporal.io_workerresourcetemplates.yaml` (new CRD manifest)
  - `helm/temporal-worker-controller/templates/webhook.yaml` (always-on `WorkerResourceTemplate` `ValidatingWebhookConfiguration`; `TemporalWorkerDeployment` webhook now behind `webhook.enabled`)
  - `helm/temporal-worker-controller/templates/certmanager.yaml` (cert-manager `Issuer` + `Certificate` for TLS, default enabled)
  - `helm/temporal-worker-controller/Chart.yaml` (cert-manager added as optional subchart dependency; opt in via `certmanager.install: true`)
  - `helm/temporal-worker-controller/templates/manager.yaml` (cert volume/port always present; `ALLOWED_KINDS`, `POD_NAMESPACE`, `SERVICE_ACCOUNT_NAME` env vars)
  - `helm/temporal-worker-controller/templates/rbac.yaml` (`WorkerResourceTemplate` + SAR rules in manager `ClusterRole`; editor/viewer roles; configurable attached-resource RBAC)
  - `helm/temporal-worker-controller/values.yaml` (`workerResourceConfig.allowedResources` default: HPA, piped to `ALLOWED_KINDS` and to controller rbac)

### Integration tests

New integration test subtests added to the existing envtest suite, all running through the shared `testTemporalWorkerDeploymentCreation` table-test runner:
  - `WorkerResourceTemplate` (7 tests): Deployment owner ref, `matchLabels` injection, multiple `WorkerResourceTemplate`s on same `TemporalWorkerDeployment`, metric selector label injection, multiple active versions, apply failure → Applied:false, SSA idempotency
  - Rollout gaps (5 tests): Progressive ramp to Current, ConnectionSpecHash annotation repair, gate input from ConfigMap, gate input from Secret, multiple deprecated versions
  - Webhook admission (5 tests, separate Ginkgo suite): Spec rejection, SAR pass, SAR fail (user), SAR fail (controller SA), `temporalWorkerDeploymentRef.name` immutability

## Why?
HPA autoscaling for versioned Temporal workers requires a separate HPA per worker version, each targeting only that version's Deployment with the correct `scaleTargetRef` and label selectors. Without this CRD, users have no way to create per-version resources that the controller lifecycle-manages alongside the versioned Deployments.

## Checklist

1. Closes #207

2. How was this tested:
    - Full envtest integration test suite: new subtests covering WRT lifecycle, previously uncovered rollout scenarios, and webhook admission via the real HTTP admission path
    - Unit tests: webhook validator, SSA naming/injection helpers, planner integration
    - All tests pass: `KUBEBUILDER_ASSETS=.../bin/k8s/1.27.1-darwin-arm64 go test -tags
  test_dep ./...`

3. Any docs updates needed?
    - `docs/worker-resource-template.md` added: concept overview, HPA example with cert-manager setup, RBAC configuration guide

### DIFF
--- a/.github/workflows/test-integration.yml
+++ b/.github/workflows/test-integration.yml
@@ -125,7 +125,13 @@ jobs:
         
     - name: Download dependencies
       run: go mod download
-      
+
+    - name: Install Helm
+      uses: azure/setup-helm@v3
+
+    - name: Build Helm chart dependencies
+      run: helm repo add jetstack https://charts.jetstack.io && helm dependency build helm/temporal-worker-controller
+
     - name: Run unit tests
       run: make test-unit
       


docs/test-coverage-analysis.md
```diff
@@ -0,0 +1,372 @@
+# Test Coverage Analysis
```
carlydf: I will be deleting this!

Also, it is AI generated, and I have only read the top half carefully. Use it for your information during review ("The Controller's Contract" might be an especially helpful section for us all to understand and agree on), but don't spend too much time on the details here. Not intended to be a part of the main repo.

api/v1alpha1/temporalworkerownedresource_types.go
```diff
@@ -0,0 +1,104 @@
...
+// This product includes software developed at Datadog (https://www.datadoghq.com/). Copyright 2024 Datadog, Inc.
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// WorkerDeploymentReference references a TemporalWorkerDeployment in the same namespace.
```
jaypipes: Temporal namespace or Kubernetes namespace?

api/v1alpha1/temporalworkerownedresource_types.go
```diff
@@ -0,0 +1,104 @@
...
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`
+	Name string `json:"name"`
+}
+
+// TemporalWorkerOwnedResourceSpec defines the desired state of TemporalWorkerOwnedResource.
+type TemporalWorkerOwnedResourceSpec struct {
+	// WorkerRef references the TemporalWorkerDeployment to attach this resource to.
+	// +kubebuilder:validation:Required
+	WorkerRef WorkerDeploymentReference `json:"workerRef"`
```
jaypipes: Should be `WorkerDeploymentRef` since this refers to the deployment not the worker.

api/v1alpha1/temporalworkerownedresource_types.go
```diff
@@ -0,0 +1,104 @@
...
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// WorkerDeploymentReference references a TemporalWorkerDeployment in the same namespace.
+type WorkerDeploymentReference struct {
```
jaypipes: The naming of objects in these new files is inconsistent in a couple ways. Some of the resources are prefixed with `Temporal` while others are not. Other resources are prefixed with `TemporalWorkerOwned` but actually refer to things that are owned by the TemporalWorkerDeployment, not the Worker.

api/v1alpha1/workerresourcetemplate_types.go
```diff
@@ -0,0 +1,104 @@
...
+
+// TemporalWorkerOwnedResourceSpec defines the desired state of TemporalWorkerOwnedResource.
+type TemporalWorkerOwnedResourceSpec struct {
+	// WorkerRef references the TemporalWorkerDeployment to attach this resource to.
+	// +kubebuilder:validation:Required
+	WorkerRef WorkerDeploymentReference `json:"workerRef"`
+
+	// Object is the Kubernetes resource template to attach to each versioned Deployment.
+	// One copy of this resource is created per active Build ID, owned by the corresponding
+	// versioned Deployment (so it is garbage collected when the Deployment is deleted).
```
jaypipes: If it's a template, this field really should be called `Template` :)

api/v1alpha1/workerresourcetemplate_types.go
```diff
@@ -0,0 +1,104 @@
...
+	//
+	// The object must include apiVersion, kind, and spec. The metadata.name and
+	// metadata.namespace are generated by the controller and must not be set by the user.
+	//
+	// String values in the spec may contain Go template expressions:
+	//   {{ .DeploymentName }}    - the controller-generated versioned Deployment name
+	//   {{ .TemporalNamespace }} - the Temporal namespace the worker connects to
+	//   {{ .BuildID }}           - the Build ID for this version
```
jaypipes: Allowing template variable interpolation is a dangerous game to get into because you now need to guard on the controller side (as opposed to the API machinery/schema-side) against malformed or malicious content. I'd recommend not supporting template variable interpolation for this first PR.

docs/owned-resources.md
```diff
@@ -0,0 +1,158 @@
...
+Each per-Build-ID copy is given a unique, DNS-safe name derived from the `(twdName, tworName, buildID)` triple. Names are capped at 47 characters to be safe for all Kubernetes resource types, including Deployment (which has pod-naming constraints that effectively limit deployment names to ~47 characters). The name always ends with an 8-character hash of the full triple, so uniqueness is guaranteed even when the human-readable prefix is truncated.
+
+Use `kubectl get <kind>` after a reconcile to see the created resources and their names.
+
+## Banned resource kinds
+
+Certain resource kinds are blocked by default to prevent misuse (e.g., using `TemporalWorkerOwnedResource` to spin up arbitrary workloads). The default banned list is:
+
+```
+Deployment, StatefulSet, Job, Pod, CronJob
```
jaypipes: Should DaemonSet be excluded as well? What about Service? At some point, you might find that the exclusion list basically includes a bunch of things and that specifying particular *supported* resource kinds (plus, perhaps CRDs) is a better approach...

helm/temporal-worker-controller/templates/certmanager.yaml
```diff
@@ -1,35 +1,32 @@
-# The following manifests contain a self-signed issuer CR and a certificate CR.
-# More document can be found at https://docs.cert-manager.io
-# WARNING: Targets CertManager v1.0. Check https://cert-manager.io/docs/installation/upgrading/ for breaking changes.
 {{- if .Values.certmanager.enabled }}
-{{ fail "certmanager isn't supported yet" }}
+# Self-signed issuer and certificate for the webhook server TLS.
+# Requires cert-manager to be installed in the cluster.
+# See https://cert-manager.io/docs/installation/ for installation instructions.
```
jaypipes: Note that you can add cert-manager as a conditional Helm chart dependency.

https://helm.sh/docs/chart_best_practices/dependencies/#conditions-and-tags

internal/k8s/ownedresources.go
```diff
@@ -0,0 +1,314 @@
...
+//     the same namespace, their field managers must differ so each can own its own
+//     set of fields without conflicting with the other.
+//
+// Using "twc/{namespace}/{name}" satisfies both: it never changes for a given TWOR
+// and is globally unique within the cluster (namespace+name is a unique identifier).
+// The string is capped at 128 characters to stay within the Kubernetes API limit.
+func OwnedResourceFieldManager(twor *temporaliov1alpha1.TemporalWorkerOwnedResource) string {
+	fm := "twc/" + twor.Namespace + "/" + twor.Name
+	return TruncateString(fm, 128)
+}
```
jaypipes: Why not just have the field manager be a const `temporal-worker-controller`? The general pattern is for the field manager to be a constant string named after the controller itself.

Examples:

* [ResourceClaim controller](https://github.com/kubernetes/kubernetes/blob/f8b277b2b8aea984f56fb284bce50eea89f9fd06/pkg/controller/resourceclaim/controller.go#L74)
* [Sample controller](https://github.com/kubernetes/kubernetes/blob/f8b277b2b8aea984f56fb284bce50eea89f9fd06/staging/src/k8s.io/sample-controller/controller.go#L65)

internal/k8s/workerresourcetemplates.go
```diff
@@ -0,0 +1,314 @@
...
+
+	// Merge labels
+	existingLabels, ok := meta["labels"].(map[string]interface{})
+	if !ok || existingLabels == nil {
+		existingLabels = make(map[string]interface{})
+	}
+	for k, v := range selectorLabels {
+		existingLabels[k] = v
+	}
+	meta["labels"] = existingLabels
```
jaypipes: I recommend using the [`unstructured`](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured) library instead of manually manipulating `map[string]interface{}`.

For example, things like this:

```go
	if spec, ok := raw["spec"].(map[string]interface{}); ok {
		autoInjectFields(spec, data.DeploymentName, selectorLabels)
	}
```

can be this:

```go
spec, ok, _ := unstructured.NestedMap(raw, "spec")
if ok {
    autoInjectFields(spec, data.DeploymentName, selectorLabels)
}
```

things like this:

```go
raw["metadata"] = meta
```

can be this:

```go
unstructured.SetNestedMap(raw, "metadata", meta)
```

etc...

internal/k8s/workerresourcetemplates.go
```diff
@@ -0,0 +1,314 @@
...
+			"uid":                string(deployment.UID),
+			"blockOwnerDeletion": blockOwnerDeletion,
+			"controller":         isController,
+		},
+	}
+
+	raw["metadata"] = meta
+
+	obj := &unstructured.Unstructured{Object: raw}
+	return obj, nil
```
jaypipes: The `unstructured.Unstructured` struct already has methods for things like `SetNamespace()` or `SetName()`, `SetLabels()` along with things like `SetOwnerReferences()`. So you might consider creating `obj` at the top of this function and using those object methods instead of the manual construction of various fields.

api/v1alpha1/temporalworkerownedresource_types.go
carlydf: kubernetes, I'll clarify

docs/owned-resources.md
carlydf: changed to an AllowList instead! default is just HPA

api/v1alpha1/workerresourcetemplate_types.go
carlydf: I was about to remove this, but then realized that I think our basic HPA autoscaling will need it (to add the correct version tag to each `approximate_backlog_count` metric that we want to scale off of).

Could do in a separate PR though to make it easier to revert / review separately

api/v1alpha1/temporalworkerownedresource_types.go
carlydf: changed to `TemporalWorkerDeploymentRef`

api/v1alpha1/temporalworkerownedresource_types.go
carlydf: replaced `TemporalWorkerOwnedResource` with `WorkerResourceTemplate` to address this

api/v1alpha1/workerresourcetemplate_types.go
carlydf: done!

internal/k8s/workerresourcetemplates.go
carlydf: done

internal/k8s/workerresourcetemplates.go
carlydf: done!

internal/k8s/ownedresources.go
carlydf: done

helm/temporal-worker-controller/templates/certmanager.yaml
carlydf: done!

api/v1alpha1/workerresourcetemplate_types.go
```diff
@@ -0,0 +1,104 @@
...
+	//   matchLabels    - set to the versioned Deployment's selector labels (for PDB, WPA, etc.)
+	// +kubebuilder:validation:Required
+	// +kubebuilder:pruning:PreserveUnknownFields
+	Template runtime.RawExtension `json:"template"`
+}
+
+// WorkerResourceTemplateVersionStatus describes the status of a worker resource template for a single Build ID.
+type WorkerResourceTemplateVersionStatus struct {
+	// BuildID is the Build ID of the versioned Deployment this status entry refers to.
+	BuildID string `json:"buildID"`
```
carlydf: would this be better as a full version? (ie DeploymentName included)
DeploymentName would be the same for all version statuses in one `WorkerResourceTemplate`, because one WRT points immutably at one TWD

api/v1alpha1/workerresourcetemplate_types.go
eniko-dif: You could limit the template to only accept the few expressions that you want to support. You can parse the template, then walk the AST and reject any node that isn't a simple variable dereference (and not a variable that exists).

api/v1alpha1/workerresourcetemplate_types.go
eniko-dif: I think we could keep BuildID and add DeploymentName too for better visibility.

api/v1alpha1/workerresourcetemplate_types.go
```diff
@@ -0,0 +1,100 @@
...
+
+	// Applied is true if the resource was successfully applied for this Build ID.
+	Applied bool `json:"applied"`
```
eniko-dif: I'm not a k8s expert, but shouldn't applied be a bit more specific to track changes to the WRT spec? Or we just want to track that it was applied at least once? I discussed with claude a bit and he said that using observedGeneration would solved this.

api/v1alpha1/workerresourcetemplate_webhook.go
```diff
@@ -0,0 +1,459 @@
...
+//     pod runs as; used when performing SubjectAccessReview checks for the controller
+//     SA. Populated via the downward API (fieldRef: spec.serviceAccountName).
+//
+//   - ALLOWED_KINDS — comma-separated list of kind names that are permitted as
+//     WorkerResourceTemplate objects (e.g. "HorizontalPodAutoscaler,PodDisruptionBudget").
+//     Configurable via workerResourceTemplate.allowedResources[*].kinds in values.yaml.
+//     Must be set; when empty or unset, all WRT kind submissions are rejected.
+func NewWorkerResourceTemplateValidator(mgr ctrl.Manager) *WorkerResourceTemplateValidator {
+	var allowedKinds []string
+	if env := os.Getenv("ALLOWED_KINDS"); env != "" {
```
eniko-dif, line 66: I think it's not really a good idea to go through env vars. We could derive the allowed resources from the controller's ClusterRole at runtime. That way we're guaranteed to be in sync with RBAC and there's no separate configuration surface that can drift or be bypassed. (though it is a bit more code to do, I guess)

api/v1alpha1/workerresourcetemplate_webhook.go
```diff
@@ -0,0 +1,459 @@
...
+			wrt.GetName(),
+			errs,
+		)
+	}
+	return warnings, nil
+}
+
+// validate runs all validation checks (pure spec + API-dependent).
+// verb is the RBAC verb to check for the underlying resource ("create" on create, "update" on update).
+func (v *WorkerResourceTemplateValidator) validate(ctx context.Context, old, new *WorkerResourceTemplate, verb string) (admission.Warnings, error) {
```
eniko-dif: nit: i'd use a bit more verbose variable names to not shadow go built-ins. Maybe oldWRT and newWRT?

api/v1alpha1/workerresourcetemplate_webhook.go
```diff
@@ -0,0 +1,459 @@
...
+// validateWithAPI performs API-dependent validation: RESTMapper scope check and
+// SubjectAccessReview for both the requesting user and the controller service account.
+// verb is the RBAC verb to check ("create" on create/update, "delete" on delete).
+// It is a no-op when Client or RESTMapper is nil (e.g., in unit tests).
+func (v *WorkerResourceTemplateValidator) validateWithAPI(ctx context.Context, wrt *WorkerResourceTemplate, verb string) (admission.Warnings, field.ErrorList) {
+	var allErrs field.ErrorList
+	var warnings admission.Warnings
+
+	if v.Client == nil || v.RESTMapper == nil {
+		return warnings, allErrs
```
eniko-dif: Shouldn't we have a log here to say the validation is skipped because the client or the rest mappers is nil? I feel like all the code below is important.

api/v1alpha1/workerresourcetemplate_webhook.go
```diff
@@ -0,0 +1,459 @@
...
+func (v *WorkerResourceTemplateValidator) validate(ctx context.Context, old, new *WorkerResourceTemplate, verb string) (admission.Warnings, error) {
+	var allErrs field.ErrorList
+	var warnings admission.Warnings
+
+	// Pure spec validation (no API calls needed)
+	specWarnings, specErrs := validateWorkerResourceTemplateSpec(new.Spec, v.AllowedKinds)
+	warnings = append(warnings, specWarnings...)
+	allErrs = append(allErrs, specErrs...)
+
+	// Immutability: temporalWorkerDeploymentRef.name must not change on update
```
eniko-dif, line 150: Apart from immutability, shouldn't we check if the ref points to an existing TWD?

api/v1alpha1/workerresourcetemplate_types.go
eniko-dif: This could be added to the template validation too so typos and other errors would be quickly caught.

internal/k8s/workerresourcetemplates.go
```diff
@@ -0,0 +1,274 @@
...
+	}
+	var buf bytes.Buffer
+	if err := tmpl.Execute(&buf, data); err != nil {
+		return "", fmt.Errorf("template execution failed for %q: %w", s, err)
+	}
+	return buf.String(), nil
+}
+
+// containsTemplateMarker returns true if s contains "{{" indicating a Go template expression.
+func containsTemplateMarker(s string) bool {
```
eniko-dif: nit: maybe `strings.Contains(s, "{{")` is enough here?

internal/k8s/workerresourcetemplates.go
```diff
@@ -0,0 +1,274 @@
...
+		// numbers, booleans, nil — pass through unchanged
+		return v, nil
+	}
+}
+
+// renderString renders a single string as a Go template with the given data.
+// Strings without template expressions are returned unchanged.
+func renderString(s string, data TemplateData) (string, error) {
+	// Fast path: skip parsing if no template markers
+	if !containsTemplateMarker(s) {
```
eniko-dif: Is this really faster than calling the lib?

internal/k8s/workerresourcetemplates.go
```diff
@@ -0,0 +1,274 @@
...
+	})
+
+	return obj, nil
+}
+
+// autoInjectFields recursively traverses obj and injects scaleTargetRef and matchLabels
+// wherever the key is present with a null value. Users signal intent by writing
+// `scaleTargetRef: null` or `matchLabels: null` in their spec.template. This covers Layer 1
+// auto-injection without the risk of adding unknown fields to resource types that don't use them.
+func autoInjectFields(obj map[string]interface{}, deploymentName string, selectorLabels map[string]string) {
```
eniko-dif: nit: I prefer recursive functions that are written in one function, because it's easier to reason about (and visualize in my head).

internal/planner/planner.go
```diff
@@ -120,9 +153,88 @@ func GeneratePlan(
...
+	for i := range wrts {
+		wrt := &wrts[i]
+		if wrt.Spec.Template.Raw == nil {
+			l.Info("skipping WorkerResourceTemplate with empty spec.template", "name", wrt.Name)
+			continue
+		}
+		for buildID, deployment := range k8sState.Deployments {
+			rendered, err := k8s.RenderWorkerResourceTemplate(wrt, deployment, buildID, temporalNamespace)
+			if err != nil {
+				l.Error(err, "failed to render WorkerResourceTemplate",
```
eniko-dif: Coming back to the "Applied" field in the VersionStatus, what happens here? Should it stay true if a previous version was applied? I feel like the users might want something more flashy than this log if it fails.

internal/planner/planner.go
```diff
@@ -120,9 +153,88 @@ func GeneratePlan(
...
+	temporalNamespace string,
+) []WorkerResourceApply {
+	var applies []WorkerResourceApply
+	for i := range wrts {
+		wrt := &wrts[i]
+		if wrt.Spec.Template.Raw == nil {
+			l.Info("skipping WorkerResourceTemplate with empty spec.template", "name", wrt.Name)
+			continue
+		}
+		for buildID, deployment := range k8sState.Deployments {
```
eniko-dif, line 202: Shouldn't we check the state of the deployment to only update ones which are still active and not being deleted for example?

api/v1alpha1/workerresourcetemplate_types.go
carlydf: addressed with [86065fe](https://github.com/temporalio/temporal-worker-controller/pull/217/commits/86065fe76d61cd10b90f8b214f495ab9044f2b12)

internal/planner/planner.go
carlydf: addressed with [86065fe](https://github.com/temporalio/temporal-worker-controller/pull/217/commits/86065fe76d61cd10b90f8b214f495ab9044f2b12)

carlydf: oh actually! not totally addressed with that one. The above-referenced commit added a Ready condition that is easily consumable by CD tools, and Ready=False if any applies failed.

[049d800](https://github.com/temporalio/temporal-worker-controller/pull/217/commits/049d8004b9ee07ade2c130fc4af4411f40f0f595) commit ensures that a failed render also gets surfaced in that same way

carlydf: great catch! thank you <3

internal/planner/planner.go
carlydf: addressed with [84efdfe](https://github.com/temporalio/temporal-worker-controller/pull/217/commits/84efdfe62810c81fe02cb2af84c4bd766c0b4006)

internal/k8s/workerresourcetemplates.go
carlydf: yes that's equivalent and more readable, replaced it!

internal/k8s/workerresourcetemplates.go
carlydf: strings.Contains is equivalent, replaced it!
as for whether `renderString` is necessary vs just always called `template.Parse` + `tmpl.Execute`, `renderString` is more efficient because `Execute` writes into a `bytes.Buffer` even for a template with no actions, allocating and copying the string unnecessarily. so `renderString` is more efficient because it just returns if it doesnt see the template marker in the string

api/v1alpha1/workerresourcetemplate_types.go
carlydf: At the render step, we already reject anything that is not in the allowed template fields. But to push those errors left in case of typos / other misunderstanding by the user of what can be templated, I added the validation you talked about to the webook with [7b7aa07](https://github.com/temporalio/temporal-worker-controller/pull/217/commits/7b7aa073b719fc56c6bcc64d787f742f761a3955), and [a9bcb2e](https://github.com/temporalio/temporal-worker-controller/pull/217/commits/a9bcb2ee6f23b8d8950a7f4f462ec2d0f73b4b69) just adds a comment specifically explaining what the `default` case covers for people who arent super familiar with the Go templating lib

api/v1alpha1/workerresourcetemplate_webhook.go
carlydf: After reading your comment I thought hard about this, but the problem with taking allowed resources for worker templates directly from the controller's cluster role is that the controller's cluster role has permission to read and modify `Deployments` and other resources the controller needs for its own operation managing the versioned workers. We don't want that permission to also mean that the user wants `WorkerResourceTemplates` to be able to create `Deployments` too. 

If we had made a separate Controller for managing `WorkerResourceTemplates`, with a separate `ClusterRole` from the controller that is managing `TemporalWorkerDeployments`, then we could consider basing `AllowedResources` directly from the `ClusterRole`. But with the current design, I think it makes sense for `AllowedResources` to be a subset of what is in the `ClusterRole`, and exactly what the user configured in helm. 

Re: the drift concern, since both `ALLOWED_KINDS` and the `ClusterRole` rules are generated from the same helm values, drift can only happen if someone manually edits the `ClusterRole` outside of Helm, and even then, the SubjectAccessReview check at admission time would catch it. So I think this is the right way to go.

As an alternative to an env var, we could instead have helm pass `--allowed-kinds=HPA,PDB` as container args instead, but I think that is pretty equivalent to the env var, so not worth changing imo?

api/v1alpha1/workerresourcetemplate_webhook.go
carlydf: addressed here [ddb9658](https://github.com/temporalio/temporal-worker-controller/pull/217/commits/ddb96583bba88a3e2438b15f993f9819880b97a1)

api/v1alpha1/workerresourcetemplate_types.go
carlydf: I looked at `BaseWorkerDeploymentVersion` because I think whatever we do here should be symmetric to how we list versions in the `TemporalWorkerDeployment`, and there it's just `BuildID` for the same reason (deployment name would be the same for the whole list), so I think this should also just be build id

https://github.com/temporalio/temporal-worker-controller/blob/7a22c5a1ab853ba98f9c082ade13e7432559fa4c/api/v1alpha1/worker_types.go#L248-L250

api/v1alpha1/workerresourcetemplate_webhook.go
carlydf: yes! addressed here: [f32623f](https://github.com/temporalio/temporal-worker-controller/pull/217/commits/f32623fb9f592601c31eea0cf7683d5816f2e123)

api/v1alpha1/workerresourcetemplate_webhook.go
carlydf: if we check TWD existence at WorkerResourceTemplate creation time, the TWD could still be deleted one instant later. Also, if both the TWD and the WRT are being applied for the first time together, ideally we should wait for the TWD to be created. At the same time, users should clearly be able to see that a WRT is not "Ready" if the TWD that it refers to does not exist. HPA acts the same way with `scaleTargetRef`. So, I (and claude) addressed this issue by setting Ready=False with reason="TemporalWorkerDeploymentNotFound" if the referenced TWD is not found. Then if the TWD arrives an instant later, the WRT will heal. I looked into how HPA handles it and it works this way too, so I think this aligns with k8s convention (and it also makes sense to me). See [8af9020](https://github.com/temporalio/temporal-worker-controller/pull/217/commits/8af902095767d001143a30d2320b3b2b228b120d) for fix

internal/k8s/workerresourcetemplates.go
carlydf: agreed! addressed by [cbb3640](https://github.com/temporalio/temporal-worker-controller/pull/217/commits/cbb3640cd292c2265a50258a024074e086a30c89)

api/v1alpha1/workerresourcetemplate_webhook.go
eniko-dif: What I've meant is that people with right permissions (or more likely Claude on their behalf...) might end up directly changing the env var, meaning that there would be one set of allowed list in the helm chart and another in env vars. But I understand your argument that the ClusterRole contains too much, so I'm not sure what is the right way to go.

api/v1alpha1/workerresourcetemplate_types.go
```diff
@@ -0,0 +1,144 @@
...
+	Name string `json:"name"`
+}
+
+// WorkerResourceTemplateSpec defines the desired state of WorkerResourceTemplate.
+type WorkerResourceTemplateSpec struct {
+	// TemporalWorkerDeploymentRef references the TemporalWorkerDeployment to attach this resource to.
+	// +kubebuilder:validation:Required
+	TemporalWorkerDeploymentRef TemporalWorkerDeploymentReference `json:"temporalWorkerDeploymentRef"`
+
+	// Object is the Kubernetes resource template to attach to each versioned Deployment.
```
jaypipes:
```suggestion
	// Template is the Kubernetes resource template to attach to each versioned Deployment.
```

api/v1alpha1/workerresourcetemplate_types.go
```diff
@@ -0,0 +1,144 @@
...
+const (
+	// ConditionTypeWRTReady is the condition type that reports whether all active
+	// Build ID instances of this WorkerResourceTemplate have been successfully applied.
+	// Using the standard "Ready" type makes this condition visible to helm install --wait
+	// and kubectl apply --wait.
+	//
+	// True:    all active Build ID instances applied successfully at the current template generation.
+	// False:   one or more instances failed to apply (see Message for the error).
+	// Unknown: no apply has been attempted yet (e.g. the WRT was just created).
+	ConditionTypeWRTReady = "Ready"
```
jaypipes: Any reason not to re-use the ConditionReady being [added in the other PR](https://github.com/temporalio/temporal-worker-controller/pull/235)?

api/v1alpha1/workerresourcetemplate_types.go
```diff
@@ -0,0 +1,144 @@
...
+	//   - 0                                  → never applied
+	//   - 0 < lastAppliedGeneration < .metadata.generation → older spec applied; current spec pending or failing (see Message)
+	//   - lastAppliedGeneration == .metadata.generation    → healthy, up-to-date
+	//   - lastAppliedGeneration > 0 && message != ""       → previously applied at that generation; current apply failing
+	// +optional
+	LastAppliedGeneration int64 `json:"lastAppliedGeneration,omitempty"`
+
+	// Message is non-empty when the most recent apply attempt failed.
+	// +optional
+	Message string `json:"message,omitempty"`
```
jaypipes: If this only contains error messages, maybe call it `Error` or `ApplyError`?

internal/controller/execplan.go
```diff
@@ -260,5 +264,195 @@ func (r *TemporalWorkerDeploymentReconciler) executePlan(ctx context.Context, l
...
+		}
+		apimeta.SetStatusCondition(&wrt.Status.Conditions, metav1.Condition{
+			Type:               temporaliov1alpha1.ConditionTypeWRTReady,
+			Status:             condStatus,
+			Reason:             condReason,
+			Message:            condMessage,
+			ObservedGeneration: wrt.Generation,
+		})
+
+		wrt.Status.Versions = versions
```
jaypipes, line 495: Did we want to sort the versions slice to make updates deterministic?

internal/controller/worker_controller.go
```diff
@@ -114,16 +117,20 @@ func (r *TemporalWorkerDeploymentReconciler) Reconcile(ctx context.Context, req
...
 	// Fetch the worker deployment
 	var workerDeploy temporaliov1alpha1.TemporalWorkerDeployment
 	if err := r.Get(ctx, req.NamespacedName, &workerDeploy); err != nil {
-		l.Error(err, "unable to fetch TemporalWorker")
-		// we'll ignore not-found errors, since they can't be fixed by an immediate
-		// requeue (we'll need to wait for a new notification), and we can get them
-		// on deleted requests.
-		return ctrl.Result{}, client.IgnoreNotFound(err)
+		if !apierrors.IsNotFound(err) {
+			l.Error(err, "unable to fetch TemporalWorker")
```
```suggestion
			l.Error(err, "unable to fetch TemporalWorkerDeployment")
```

:)

internal/controller/worker_controller.go
```diff
@@ -348,18 +387,47 @@ func (r *TemporalWorkerDeploymentReconciler) SetupWithManager(mgr ctrl.Manager)
...
+	}
+	return []reconcile.Request{
+		{
+			NamespacedName: types.NamespacedName{
+				Name:      wrtObj.Spec.TemporalWorkerDeploymentRef.Name,
+				Namespace: wrt.GetNamespace(),
+			},
+		},
+	}
+}
```
jaypipes: The name of this method is quite confusing :) How about `reconcileRequestForWRT` and changing the docstring to something like:

```go
// reconcileRequestForWRT returns a reconcile.Request to reconcile the TWD associated with the
// supplied WRT.
```

internal/k8s/workerresourcetemplates.go
```diff
@@ -0,0 +1,293 @@
...
+
+// RenderedHashLen is the number of hex characters used for rendered-object hashes stored in
+// WorkerResourceTemplateVersionStatus.LastAppliedHash. 16 hex chars = 8 bytes = 64 bits,
+// giving negligible collision probability across thousands of (WRT × BuildID) pairs.
+const RenderedHashLen = 16
+
+// WorkerResourceTemplateFieldManager is the SSA field manager name used when applying
+// WorkerResourceTemplate-rendered resources. A single constant per controller is the
+// standard Kubernetes pattern (see e.g. the ResourceClaim controller).
+const WorkerResourceTemplateFieldManager = "temporal-worker-controller"
```
jaypipes: Could also just name this `FieldManager` and use it throughout the controller, not just for WRTs.

api/v1alpha1/workerresourcetemplate_types.go
carlydf: we should! I just haven't pulled the changes from the other PR in this branch yet -- good note

docs/crd-management.md
```diff
@@ -125,8 +125,25 @@ helm upgrade temporal-worker-controller \
   --version <new-version> \
   --namespace temporal-system
 
-# Step 2: Install the CRDs chart to take Helm ownership of the existing CRDs
-# kubectl apply reconciles any changes; same-version CRDs are a no-op on the cluster
+# Step 2: Stamp Helm ownership labels/annotations onto the existing CRDs.
```
carlydf, line 128: @Shivs11 this change has to go into v0.21.0 chart actually / v1.4.0 github tag / whatever chart version we first pull the CRDs out in. just commenting to remind both of us

api/v1alpha1/workerresourcetemplate_types.go
```diff
@@ -0,0 +1,137 @@
...
+	//   scaleTargetRef: {}
+	//     Injected to point at the versioned Deployment. Use this for HPAs, WPAs, and
+	//     other autoscaler CRDs that need a scale target.
+	//
+	//   spec.selector.matchLabels: {}
+	//     Injected with the versioned Deployment's pod selector labels. Use this for
+	//     PodDisruptionBudgets and other resources that select pods.
+	//
+	//   spec.metrics[*].external.metric.selector.matchLabels: {} (or with user labels)
+	//     The controller appends worker_deployment_name, worker_deployment_build_id, and
```
carlydf: change `worker_deployment_build_id` -> `temporal_worker_build_id`, and `worker_deployment_name` -> `temporal_worker_deployment_name`

api/v1alpha1/workerresourcetemplate_types.go
```diff
@@ -0,0 +1,137 @@
...
+	// +optional
+	ResourceName string `json:"resourceName,omitempty"`
+
+	// LastAppliedGeneration is the WorkerResourceTemplate metadata.generation at the time
+	// of the last **successful** apply for this Build ID. A value of 0 means no successful
+	// apply has occurred yet.
+	//
+	// Reading the status:
+	//   - 0                                  → never applied
+	//   - 0 < lastAppliedGeneration < .metadata.generation → older spec applied; current spec pending or failing (see Message)
```
carlydf:
```suggestion
	//   - 0 < lastAppliedGeneration < .metadata.generation → older spec applied; current spec pending or failing (see ApplyError)
```

api/v1alpha1/workerresourcetemplate_types.go
```diff
@@ -0,0 +1,137 @@
...
+//+kubebuilder:resource:shortName=wrt
+//+kubebuilder:printcolumn:name="Worker",type="string",JSONPath=".spec.temporalWorkerDeploymentRef.name",description="Referenced TemporalWorkerDeployment"
+//+kubebuilder:printcolumn:name="Kind",type="string",JSONPath=".spec.template.kind",description="Kind of resource that is being templated"
+//+kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp",description="Age"
+
+// WorkerResourceTemplate attaches an arbitrary namespaced Kubernetes resource
+// (HPA, PDB, WPA, custom CRDs, etc.) to each per-Build-ID versioned Deployment
+// managed by a TemporalWorkerDeployment. One copy of the resource is created per
+// active Build ID, owned by this WorkerResourceTemplate (so all copies are deleted
+// when the WorkerResourceTemplate itself is deleted).
```
carlydf:
```suggestion
// WorkerResourceTemplate creates an arbitrary namespaced Kubernetes resource
// (HPA, PDB, WPA, custom CRDs, etc.) for each per-Build-ID versioned Deployment
// managed by a TemporalWorkerDeployment. One instance of the resource is created
// per active Build ID, and is owned by this WorkerResourceTemplate (so all instances
// are deleted when the WorkerResourceTemplate itself is deleted).
```

api/v1alpha1/workerresourcetemplate_webhook.go
```diff
@@ -0,0 +1,533 @@
...
+
+// WorkerResourceTemplateValidator validates WorkerResourceTemplate objects.
+// It holds API-dependent dependencies (client, RESTMapper, controller SA identity).
+// +kubebuilder:object:generate=false
+type WorkerResourceTemplateValidator struct {
+	Client                client.Client
+	RESTMapper            meta.RESTMapper
+	ControllerSAName      string
+	ControllerSANamespace string
+	// AllowedKinds is the explicit list of resource kinds permitted as WRT objects.
```
carlydf:
```suggestion
	// AllowedKinds is the explicit list of resource kinds permitted as WorkerResourceTemplate objects.
```

api/v1alpha1/workerresourcetemplate_webhook.go
```diff
@@ -0,0 +1,533 @@
...
+//     controller SA. Populated via the downward API (fieldRef: metadata.namespace).
+//
+//   - SERVICE_ACCOUNT_NAME — name of the Kubernetes ServiceAccount the controller
+//     pod runs as; used when performing SubjectAccessReview checks for the controller
+//     SA. Populated via the downward API (fieldRef: spec.serviceAccountName).
+//
+//   - ALLOWED_KINDS — comma-separated list of kind names that are permitted as
+//     WorkerResourceTemplate objects (e.g. "HorizontalPodAutoscaler,PodDisruptionBudget").
+//     Configurable via workerResourceTemplate.allowedResources[*].kinds in values.yaml.
+//     Must be set; when empty or unset, all WRT kind submissions are rejected.
```
carlydf:
```suggestion
//     Must be set; when empty or unset, all WorkerResourceTemplate kind submissions are rejected.
```

api/v1alpha1/workerresourcetemplate_webhook.go
carlydf: ah yeah that makes sense. I'm going to keep it this way since I don't want to increase scope here by redesigning how this works.

api/v1alpha1/workerresourcetemplate_webhook.go
```diff
@@ -0,0 +1,533 @@
...
+// validateWorkerResourceTemplateSpec performs pure (no-API) validation of the spec fields.
+// It checks structural constraints that can be evaluated without talking to the API server.
+func validateWorkerResourceTemplateSpec(spec WorkerResourceTemplateSpec, allowedKinds []string) (admission.Warnings, field.ErrorList) {
+	var allErrs field.ErrorList
+	var warnings admission.Warnings
+
+	if spec.Template.Raw == nil {
+		allErrs = append(allErrs, field.Required(
+			field.NewPath("spec").Child("template"),
+			"object must be specified",
```
carlydf:
```suggestion
			"template must be specified",
```

api/v1alpha1/workerresourcetemplate_webhook.go
```diff
@@ -0,0 +1,533 @@
...
+		))
+		return warnings, allErrs
+	}
+
+	var obj map[string]interface{}
+	if err := json.Unmarshal(spec.Template.Raw, &obj); err != nil {
+		allErrs = append(allErrs, field.Invalid(
+			field.NewPath("spec").Child("template"),
+			"<raw>",
+			fmt.Sprintf("failed to parse object: %v", err),
```
carlydf:
```suggestion
			fmt.Sprintf("failed to parse template: %v", err),
```

api/v1alpha1/workerresourcetemplate_webhook.go
```diff
@@ -0,0 +1,533 @@
...
+	if innerSpec, ok := obj["spec"].(map[string]interface{}); ok {
+		innerSpecPath := field.NewPath("spec").Child("template").Child("spec")
+
+		// 4. minReplicas must not be 0.
+		// Scaling to zero is not safe with Temporal's approximate_backlog_count metric: that metric
+		// is only emitted while the task queue is loaded in memory (i.e. while at least one worker
+		// is polling). If all workers are scaled to zero and the task queue goes idle for ~5 minutes,
+		// the metric stops being emitted and resets to zero. If new tasks then arrive but no worker
+		// polls, the metric remains zero — making it impossible for a metric-based autoscaler to
+		// detect the backlog and scale back up. Until Temporal provides a reliable mechanism for
```
carlydf:
```suggestion
		// detect the backlog and scale back up. Until Temporal makes this a reliable metric for
```

api/v1alpha1/workerresourcetemplate_webhook.go
```diff
@@ -0,0 +1,533 @@
...
+						"from zero once all workers are gone",
+				))
+			}
+		}
+
+		// 5. scaleTargetRef: if absent or empty ({}), the controller injects it to point at
+		// the versioned Deployment. If non-empty, reject — the controller owns this field and any
+		// hardcoded value would point at the wrong Deployment.
+		// Note: {} is the required opt-in sentinel. null is stripped by the k8s API server before
+		// storage, so the controller would never see it and injection would not occur.
```
carlydf:
```suggestion
		// storage, so the controller would never see the field and injection would not occur.
```

api/v1alpha1/workerresourcetemplate_webhook.go
```diff
@@ -0,0 +1,533 @@
...
+	return warnings, allErrs
+}
+
+// checkScaleTargetRefNotSet recursively traverses obj looking for any scaleTargetRef that is
+// set to a non-empty value. If absent or empty ({}), the controller injects it to point
+// at the versioned Deployment. If non-empty, reject — the controller owns this field when present,
+// and a hardcoded value would point at the wrong (non-versioned) Deployment.
+//
+// Note: {} is the required opt-in sentinel. null is stripped by the Kubernetes API server before
+// storage, so the controller would never see it and injection would not occur.
```
carlydf:
```suggestion
// storage, so the controller would never see the field and injection would not occur.
```

api/v1alpha1/workerresourcetemplate_webhook.go
```diff
@@ -0,0 +1,533 @@
...
+			checkScaleTargetRefNotSet(nested, path.Child(k), allErrs)
+		}
+	}
+}
+
+// checkMetricSelectorLabelsNotSet validates that metrics[*].external.metric.selector.matchLabels
+// does not contain any of the controller-owned keys. The controller appends
+// worker_deployment_name, worker_deployment_build_id, and temporal_namespace to whatever
+// matchLabels the user provides; hardcoding these keys would either conflict or be stale.
+// User labels (e.g. task_type: "Activity") are permitted alongside the controller-owned keys.
```
carlydf:
```suggestion
// checkMetricSelectorLabelsNotSet validates that metrics[*].external.metric.selector.matchLabels
// does not contain any of the controller-owned keys. The controller appends
// worker_deployment_name, worker_deployment_build_id, and temporal_namespace to whatever
// matchLabels the user provides. User labels (e.g. task_type: "Activity") are permitted alongside
// the controller-owned keys.
```

docs/worker-resource-templates.md
```diff
@@ -0,0 +1,177 @@
+# WorkerResourceTemplate
+
+`WorkerResourceTemplate` lets you attach arbitrary Kubernetes resources — HPAs, PodDisruptionBudgets, custom CRDs — to each worker version that has running workers. The controller creates one copy of the resource per worker version with a running Deployment, automatically wired to the correct versioned Deployment.
+
+## Why you need this
+
+The Temporal Worker Controller creates one Kubernetes `Deployment` per worker version (Build ID). If you attach an HPA directly to a single Deployment, it breaks as versions roll over — the old HPA still targets the old Deployment, the new Deployment has no HPA, and you have to manage cleanup yourself.
+
+`WorkerResourceTemplate` solves this by treating the resource as a template. The controller renders one instance per worker version with running workers, injects the correct versioned Deployment name, and cleans up automatically when the versioned Deployment is deleted (e.g., during the sunset process after traffic has drained).
```
carlydf:
```suggestion
`WorkerResourceTemplate` solves this by treating the resource as a template. The controller renders one instance per worker version with running workers, injects the correct versioned Kubernetes Deployment name, injects per-version metrics tags where appropriate, and cleans up automatically when the versioned Deployment is deleted (e.g., during the sunset process after traffic has drained).
```

docs/worker-resource-templates.md
```diff
@@ -0,0 +1,177 @@
...
+
+This is also the recommended mechanism for metric-based or backlog-based autoscaling: attach a standard HPA with custom metrics to your workers and the controller keeps one per running worker version, each pointing at the right Deployment.
```
carlydf:
```suggestion
This is also the recommended mechanism for metric-based or backlog-based autoscaling: attach a standard HPA with custom metrics to your workers and the controller keeps one per running worker version, each pointing at the right Deployment and version-tagged metrics.
```

docs/worker-resource-templates.md
```diff
@@ -0,0 +1,177 @@
...
+
+## How it works
+
+1. You create a `WorkerResourceTemplate` that references a `TemporalWorkerDeployment` and contains the resource spec in `spec.template`.
+2. The validating webhook checks that you have permission to manage that resource type yourself (SubjectAccessReview), and that the resource kind is on the allowed list (see below).
+3. On each reconcile loop, the controller renders one copy of `spec.template` per worker version with a running Deployment, injects fields (see below), and applies it via Server-Side Apply.
+4. Each copy is owned by the corresponding versioned `Deployment`, so it is garbage-collected automatically when that Deployment is deleted.
```
carlydf:
```suggestion
4. Each per-version copy is deleted by the controller when its corresponding Kubernetes Deployment is deleted per the sunset policy.
```

helm/temporal-worker-controller/templates/webhook.yaml
```diff
@@ -1,85 +1,84 @@
...
+  selector:
+    {{- include "temporal-worker-controller.selectorLabels" . | nindent 4 }}
 ---
+# ValidatingWebhookConfiguration for WorkerResourceTemplate.
+# Always installed — this webhook is a required security control that validates
+# all WorkerResourceTemplate create/update/delete requests.
 apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingWebhookConfiguration
 metadata:
   name: {{ .Release.Name }}-validating-webhook-configuration
```
carlydf:
```suggestion
  name: {{ .Release.Name }}-wrt-validating-webhook-configuration
```

internal/controller/worker_controller.go
```diff
@@ -297,6 +304,38 @@ func (r *TemporalWorkerDeploymentReconciler) Reconcile(ctx context.Context, req
...
+// reference on the WRT (which would otherwise cause Kubernetes GC to delete the WRT).
+// Returns an error if the List or any status update fails so the caller can requeue.
+func (r *TemporalWorkerDeploymentReconciler) markWRTsTWDNotFound(ctx context.Context, twd types.NamespacedName) error {
+	l := log.FromContext(ctx)
+	var wrtList temporaliov1alpha1.WorkerResourceTemplateList
+	if err := r.List(ctx, &wrtList,
+		client.InNamespace(twd.Namespace),
+		client.MatchingFields{wrtWorkerRefKey: twd.Name},
+	); err != nil {
+		return fmt.Errorf("list WRTs referencing missing TWD %q: %w", twd.Name, err)
```
carlydf:
```suggestion
		return fmt.Errorf("list WorkerResourceTemplates referencing missing TemporalWorkerDeployment %q: %w", twd.Name, err)
```

internal/controller/worker_controller.go
```diff
@@ -297,6 +304,38 @@ func (r *TemporalWorkerDeploymentReconciler) Reconcile(ctx context.Context, req
...
+		meta.SetStatusCondition(&wrt.Status.Conditions, metav1.Condition{
+			Type:               temporaliov1alpha1.ConditionReady,
+			Status:             metav1.ConditionFalse,
+			Reason:             temporaliov1alpha1.ReasonWRTTWDNotFound,
+			Message:            fmt.Sprintf("TemporalWorkerDeployment %q not found", twd.Name),
+			ObservedGeneration: wrt.Generation,
+		})
+		if err := r.Status().Update(ctx, wrt); err != nil {
+			l.Error(err, "unable to update WRT status for missing TWD", "wrt", wrt.Name, "twd", twd.Name)
+			errs = append(errs, fmt.Errorf("update status for WRT %s/%s: %w", wrt.Namespace, wrt.Name, err))
```
carlydf:
```suggestion
			l.Error(err, "unable to update WorkerResourceTemplate status for missing TemporalWorkerDeployment", "wrt", wrt.Name, "twd", twd.Name)
			errs = append(errs, fmt.Errorf("update status for WorkerResourceTemplate %s/%s: %w", wrt.Namespace, wrt.Name, err))
```

internal/demo/helloworld/helm/helloworld/templates/deployment.yaml
```diff
@@ -18,20 +18,20 @@ spec:
...
 {{- end }}
     gate:
       workflowType: "RolloutGate"
   sunset:
     scaledownDelay: 30s
-    deleteDelay: 24h
+    deleteDelay: 5m
   # Desired number of worker replicas
-  replicas: 1
+  #replicas:
```
carlydf:
```suggestion
```

internal/demo/helloworld/worker.go
```diff
@@ -17,22 +16,29 @@ import (
 
 func HelloWorld(ctx workflow.Context) (string, error) {
 	ctx = util.SetActivityTimeout(ctx, 5*time.Minute)
+	print("hi")
```
carlydf:
```suggestion
```

PR #220, opened 2026-03-08
Use `ManagerIdentity` API instead of `LastModifierIdentity` + ignore-last-modifier metadata hack

### LINKED ISSUE #162: Use `ManagerIdentity` instead of `LastModifierIdentity` to handle changes made by external users
And update https://github.com/temporalio/temporal-worker-controller/blob/main/docs/ownership.md#solution

carlydf:
## What was changed
Replace the `LastModifierIdentity` + `temporal.io/ignore-last-modifier` metadata mechanism with the Temporal server's first-class `ManagerIdentity` API for Worker Deployment ownership.

**Controller behavior:**
  - When the controller is about to make its first routing change to a Worker Deployment (i.e. `ManagerIdentity` is empty), it calls `SetManagerIdentity(Self=true)` to claim ownership before applying the change. The conflict token returned is threaded into the subsequent SetCurrentVersion / SetRampingVersion call.
  - If `ManagerIdentity` is already set to a different identity, the routing change is submitted as-is and the Temporal server rejects it. The controller returns an error and retries on backoff -> no special manual-mode logic needed.
  - The current `ManagerIdentity` value is surfaced in `TemporalWorkerDeploymentStatus.managerIdentity`.

**Planner:** `ClaimManagerIdentity` bool is now a field on `planner.VersionConfig`, keeping the decision unit-testable alongside other plan decisions.

**Removed:** the `LastModifierIdentity`-based forced-manual-strategy logic in `genplan.go` and the `IgnoreLastModifier/RemoveIgnoreLastModifierBuilds` plumbing.

## Why?
The old mechanism was a controller-side workaround: it watched `LastModifierIdentity`, forced manual strategy when it saw a non-controller modifier, and required users to set a special metadata key to pass ownership back. This was fragile (race-prone, required reading the metadata contract) and blocked on version-level metadata support.

`ManagerIdentity` is a server-enforced, deployment-level field designed exactly for this use case. It gives a clear, atomic protocol for ownership transfer without any special metadata conventions.

## Checklist

1. Closes #162 

2. How was this tested:
    - New planner unit tests cover the `ClaimManagerIdentity` decision (claims when `ManagerIdentity` is empty, does not claim when already set by someone else).
    - Six integration tests updated: three `*-blocked-by-manager-identity` tests verify the controller cannot make routing changes when `ManagerIdentity` is held by an external identity; three `*-unblocked-after-manager-identity-cleared` tests verify the controller proceeds normally once `ManagerIdentity` is cleared.

3. Any docs updates needed?
    - Yes: `docs/ownership.md` has been rewritten to document the `ManagerIdentity` mechanism and the `temporal worker deployment manager-identity set/unset` CLI commands for taking and returning control.

### DIFF
--- a/internal/planner/planner_test.go
+++ b/internal/planner/planner_test.go
@@ -38,10 +38,10 @@ func TestGeneratePlan(t *testing.T) {
 		expectUpdate                     int
 		expectWorkflow                   int
 		expectConfig                     bool
-		expectConfigSetCurrent           *bool  // pointer so we can test nil
-		expectConfigRampPercent          *int32 // pointer so we can test nil, in percentage (0-100)
-		expectManagerIdentity             *string // pointer so nil means "don't assert"
-		maxVersionsIneligibleForDeletion *int32 // set by env if non-nil, else default 75
+		expectConfigSetCurrent           *bool   // pointer so we can test nil
+		expectConfigRampPercent          *int32  // pointer so we can test nil, in percentage (0-100)
+		expectManagerIdentity            *string // pointer so nil means "don't assert"
+		maxVersionsIneligibleForDeletion *int32  // set by env if non-nil, else default 75
 	}{
 		{
 			name: "empty state creates new deployment",
@@ -435,9 +435,9 @@ func TestGeneratePlan(t *testing.T) {
 					Strategy: temporaliov1alpha1.UpdateAllAtOnce,
 				},
 			},
-			expectConfig:               true,
-			expectConfigSetCurrent:     func() *bool { b := true; return &b }(),
-			expectManagerIdentity: func() *string { s := ""; return &s }(),
+			expectConfig:           true,
+			expectConfigSetCurrent: func() *bool { b := true; return &b }(),
+			expectManagerIdentity:  func() *string { s := ""; return &s }(),
 		},
 		{
 			// Same routing change scenario but ManagerIdentity is already set — no claim.
@@ -482,9 +482,9 @@ func TestGeneratePlan(t *testing.T) {
 					Strategy: temporaliov1alpha1.UpdateAllAtOnce,
 				},
 			},
-			expectConfig:               true,
-			expectConfigSetCurrent:     func() *bool { b := true; return &b }(),
-			expectManagerIdentity: func() *string { s := "some-other-client"; return &s }(),
+			expectConfig:           true,
+			expectConfigSetCurrent: func() *bool { b := true; return &b }(),
+			expectManagerIdentity:  func() *string { s := "some-other-client"; return &s }(),
 		},
 	}
 


internal/tests/internal/deployment_controller.go
```diff
@@ -132,8 +139,8 @@ func setHealthyDeploymentStatus(t *testing.T, ctx context.Context, k8sClient cli
 			},
 		},
 	}
-	t.Logf("started %d healthy workers, updating deployment status", *deployment.Spec.Replicas)
-	if err := k8sClient.Status().Update(ctx, &deployment); err != nil {
+	t.Logf("started %d healthy workers, updating deployment status", *fresh.Spec.Replicas)
+	if err := k8sClient.Status().Update(ctx, &fresh); err != nil {
```
carlydf, line 143: this is unrelated to the PR, fixes a flaky test. Commit message for more info:

> Refetch the Deployment immediately before calling Status().Update so the resourceVersion is always current. Previously, the deployment was read at the start of applyDeployment, workers were started (up to 30s), and then the stale object was used for the status update — racing with the controller performing a rolling update in the meantime.

internal/planner/planner.go
```diff
@@ -45,6 +46,10 @@ type VersionConfig struct {
 	SetCurrent bool
 	// Acceptable values [0,100]
 	RampPercentage int32
+
+	// ClaimManagerIdentity indicates the controller should call SetManagerIdentity(Self=true)
+	// before applying this version config, because ManagerIdentity is currently unset.
+	ClaimManagerIdentity bool
```
jaypipes: Instead of a bool field, how about a string field called `ManagerIdentity` that matches the definition of the [`WorkerDeployment.ManagerIdentity` message field.](https://github.com/temporalio/api/blob/937084678c5234295bca14bd48e314d2951fd6f0/temporal/api/deployment/v1/message.proto#L208). I find mixing imperative/instruction-style fields with declarative-style fields in a struct called `VersionConfig` to be confusing, personally. Better to have logic and conditionals defined in the code that *consumes* the `VersionConfig` struct instead of having those conditionals/logic be encoded in bool fields.

In `execplan.go`, you could have a separate `shouldClaimManagerIdentity()` function that examines the `VersionConfig.ManagerIdentity` field:

```go
func (r *TemporalWorkerDeploymentReconciler) shouldClaimManagerIdentity(
	vcfg *VersionConfig,
) bool {
	return vcfg.ManagerIdentity == ""
}
```
and a little helper function to perform the claim if necessary:

```go
func (r *TemporalWorkerDeploymentReconciler) claimManagerIdentity(
	ctx context.Context,
	l logr.Logger,
	workerDeploy *temporaliov1alpha1.TemporalWorkerDeployment,
	deploymentHandler sdkclient.WorkerDeploymentHandle,
	vcfg *VersionConfig,
) error {
	resp, err := deploymentHandler.SetManagerIdentity(ctx, sdkclient.WorkerDeploymentSetManagerIdentityOptions{
		Self:          true,
		ConflictToken: vcfg.ConflictToken,
	})
	if err != nil {
		l.Error(err, "unable to claim manager identity")
		r.Recorder.Eventf(workerDeploy, corev1.EventTypeWarning, ReasonManagerIdentityClaimFailed,
			"Failed to claim manager identity: %v", err)
		return err
	}
	l.Info("claimed manager identity", "identity", getControllerIdentity())
	// Use the updated conflict token for the subsequent routing config change.
	vcfg.ConflictToken = resp.ConflictToken
	return nil
}
```

Then `updateVersionConfig` would be a little easier to follow/read:

```go
func (r *TemporalWorkerDeploymentReconciler) updateVersionConfig(ctx context.Context, l logr.Logger, workerDeploy *temporaliov1alpha1.TemporalWorkerDeployment, deploymentHandler sdkclient.WorkerDeploymentHandle, p *plan) error {
	vcfg := p.UpdateVersionConfig
	if vcfg == nil {
		return nil
	}
	if r.shouldClaimManagerIdentity(vcfg) {
		if err := r.claimManagerIdentity(ctx, l, workerDeploy, vcfg); err != nil {
			return fmt.Errorf("unable to claim manager identity: %w", err)
		}
	}
...
```

carlydf: makes sense will do!

carlydf: done!

PR #221, opened 2026-03-08
Fix metrics-bind-address binding to localhost when auth proxy is disabled

### LINKED ISSUE #167: Metrics port binding to localhost

When deployed  worker-controller-manager using CRD with auth-proxy enabled as false and metrics enabled true and metrics port provided as 8080, pod gets deployed with a binding of metrics port to localhost.  

Current : --metrics-bind-address=127.0.0.1:8080

Instead can it be deployed without binding to localhost so the metrics are available to be scrapped

Expected:  --metrics-bind-address=:8080

carlydf:
## What was changed and why
When metrics.disableAuth=true, no kube-rbac-proxy sidecar is injected, so the manager must bind metrics to 0.0.0.0 for Prometheus to scrape them. Previously the address was hardcoded to 127.0.0.1 regardless of this setting.

Also improves values.yaml documentation to explain the two supported modes (auth-protected vs unauthenticated) and how authProxy.enabled and metrics.disableAuth should be set together.

## Checklist

1. Closes #167 

2. How was this tested:
Ran `helm template` against both configurations with metrics enabled and confirmed the rendered arg:

disableAuth=true should output --metrics-bind-address=:8080
``` 
helm template test ./helm/temporal-worker-controller \
    --set metrics.enabled=true \
    --set metrics.disableAuth=true \
    | grep metrics-bind-address
```

disableAuth=false should output --metrics-bind-address=127.0.0.1:8080
```  
helm template test ./helm/temporal-worker-controller \
    --set metrics.enabled=true \
    --set metrics.disableAuth=false \
    | grep metrics-bind-address
```

3. Any docs updates needed?
helm values file docs updated.

### DIFF
--- a/helm/temporal-worker-controller/values.yaml
+++ b/helm/temporal-worker-controller/values.yaml
@@ -57,19 +57,42 @@ affinity: {}
 # More than one replica is required for high availability.
 replicas: 2
 
-# Opt out of these resources if you want to disable the
-# auth proxy (https://github.com/brancz/kube-rbac-proxy)
-# which protects your /metrics endpoint.
+# authProxy and metrics.disableAuth together control how /metrics is exposed.
+# They should be set consistently — see the two supported modes below.
+#
+# Mode 1 — Auth-protected metrics (default, recommended for production):
+#   authProxy.enabled: true
+#   metrics.disableAuth: false
+#
+#   A kube-rbac-proxy sidecar (https://github.com/brancz/kube-rbac-proxy) is
+#   injected into the manager pod. It listens on HTTPS port 8443 and proxies to
+#   the manager's metrics endpoint on localhost, authorizing each request via
+#   Kubernetes SubjectAccessReviews. The manager binds metrics to 127.0.0.1 so
+#   it is only reachable through the proxy.
+#
+#   Use this when: running in production, using Prometheus Operator with
+#   ServiceMonitor + RBAC bearer tokens, or in multi-tenant clusters where
+#   metrics should not be freely readable.
+#
+# Mode 2 — Unauthenticated metrics:
+#   authProxy.enabled: false
+#   metrics.disableAuth: true
+#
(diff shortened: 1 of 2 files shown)


helm/temporal-worker-controller/values.yaml
```diff
@@ -57,19 +57,42 @@ affinity: {}
...
+# Mode 2 — Unauthenticated metrics:
+#   authProxy.enabled: false
+#   metrics.disableAuth: true
+#
+#   No proxy sidecar is injected. The manager binds metrics to 0.0.0.0:metrics.port
+#   so Prometheus can scrape it directly without credentials.
+#
+#   Use this when: network-level controls already restrict access (NetworkPolicy,
+#   service mesh, same-namespace Prometheus), your scraper cannot present a
+#   bearer token, or simplicity is preferred (e.g. dev/staging).
```
jaypipes, line 86: Technically, since there are two distinct boolean values fields, there are more modes:

Mode 3:

```
#   authProxy.enabled: false
#   metrics.disableAuth: false
```

Mode 4:

```
#   authProxy.enabled: true
#   metrics.disableAuth: true
```

But, you note above:

> They should be set consistently — see the two supported modes below.

If this is the case, I'd recommend just having a single values field instead of relying on the user to set the two fields consistently.

jaypipes (commented):
I don't actually think there should be two values fields... a single field  `auth.enableProxy` would be easier to reason about and explain.

helm/temporal-worker-controller/values.yaml
carlydf: I'd also prefer a single values field, but I'm concerned about making a "breaking change" to the chart, such that people's previous values files won't map to the new chart. Do you think it's fine in this case to require people to update their values files?

jaypipes: @carlydf I think marking one of the fields as deprecated would be best. I'd recommend making the `metrics.disableAuth` field the one that is deprecated since having a negative polarity and a default value of false seems awkward to me. :)

PR #228, opened 2026-03-12
fix: lowercase CleanStringForDNS output for RFC 1123 compliance

aarontsharp:
## Summary

`CleanStringForDNS` strips invalid characters but does not lowercase the result. When an image tag contains uppercase characters (e.g. `myimage:master--HEAD`), `ComputeVersionedDeploymentName` produces a deployment name that violates RFC 1123 DNS label rules, causing Kubernetes to reject the Deployment.

This adds `strings.ToLower()` to `CleanStringForDNS` so deployment names are always valid DNS labels.

### Why `CleanStringForDNS` and not `cleanBuildID`?

The build ID is also used as a Kubernetes label value and as the Temporal worker build ID. Labels allow uppercase (`[a-zA-Z0-9._-]`), and Temporal build IDs just need to be ASCII. Lowercasing only in `CleanStringForDNS` keeps the fix scoped to where DNS compliance is required (resource names) without affecting label values or Temporal version matching.

### DIFF
--- a/internal/k8s/deployments.go
+++ b/internal/k8s/deployments.go
@@ -187,7 +187,8 @@ func TruncateString(s string, n int) string {
 func CleanStringForDNS(s string) string {
 	// Keep only letters, numbers, and dashes.
 	re := regexp.MustCompile(`[^a-zA-Z0-9-]+`)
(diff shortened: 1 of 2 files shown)


jaypipes (approved):
👍

PR #229, opened 2026-03-18
omit DescribeVersion API calls for drained versions

Shivs11:
## Summary
- Switch `GetWorkerDeploymentState` from the SDK's `Describe()` to the gRPC `DescribeWorkerDeployment` API, which returns full `DrainageInfo` (including `DrainedSince` timestamps) in the version summaries
- Eliminate per-version `DescribeVersion` calls for obtaining drainage timestamps — read them directly from the gRPC summary
- Additionally, change/alter the constraint surrounding drained versions that are considered eligible to be deleted by the server from the worker-controller's pov. 

## Test plan
- [ ] Verify existing unit tests pass (`go test ./internal/temporal/...`)
- [ ] Verify integration tests pass (`go test ./internal/tests/...`)
- [ ] Verify `DrainedSince` is correctly populated from the gRPC summary
- [ ] Verify `NoTaskQueuesHaveVersionedPoller` still works for drained versions with k8s Deployments at 0 replicas
- [ ] Verify drained versions without k8s Deployments no longer trigger `DescribeVersion` calls


internal/temporal/worker_deployment.go
```diff
@@ -119,80 +126,71 @@ func GetWorkerDeploymentState(
...
 	// Set ramping since time if applicable
-	if routingConfig.RampingVersion != nil {
-		var (
-			rampingSinceTime   = metav1.NewTime(routingConfig.RampingVersionChangedTime)
-			lastRampUpdateTime = metav1.NewTime(routingConfig.RampingVersionPercentageChangedTime)
-		)
-		state.RampingSince = &rampingSinceTime
-		state.RampLastModifiedAt = &lastRampUpdateTime
+	if routingConfig.RampingDeploymentVersion != nil {
+		if routingConfig.RampingVersionChangedTime != nil {
```
Shivs11, line 131: I know the server would never set a ramping version to be a non-nil without a ramping changed time. However, now, we are directly using the gRPC layer to communicate with the server. The SDK, prior to this PR, was smart enough to convert nil timestamps (if ever present) to time.Time values (nil-safe)

we are now playing with protobuf pointers, and the thought process was to be safe than sorry.

internal/controller/reconciler_events_test.go
```diff
@@ -215,6 +218,20 @@ type stubWDClient struct {
 
 func (s *stubWDClient) GetHandle(_ string) sdkclient.WorkerDeploymentHandle { return s.handle }
 
+// stubWorkflowServiceClient implements workflowservice.WorkflowServiceClient, returning
+// a NotFound error for DescribeWorkerDeployment (matching the default stub behavior).
```
jaypipes: It doesn't look like the return from DescribeWorkerDeployment is a NotFound :) Looks like the return is a `workflowservice.DescribeWorkerDeploymentResponse` to me!

internal/temporal/worker_deployment.go
jaypipes: ++ @Shivs11 it's good to be safe, here.

internal/temporal/worker_deployment.go
```diff
@@ -119,80 +126,71 @@ func GetWorkerDeploymentState(
...
+		if routingConfig.CurrentDeploymentVersion != nil &&
+			version.DeploymentVersion.DeploymentName == routingConfig.CurrentDeploymentVersion.DeploymentName &&
+			version.DeploymentVersion.BuildId == routingConfig.CurrentDeploymentVersion.BuildId {
 			versionInfo.Status = temporaliov1alpha1.VersionStatusCurrent
-		} else if routingConfig.RampingVersion != nil &&
-			version.Version.DeploymentName == routingConfig.RampingVersion.DeploymentName &&
-			version.Version.BuildID == routingConfig.RampingVersion.BuildID {
+		} else if routingConfig.RampingDeploymentVersion != nil &&
+			version.DeploymentVersion.DeploymentName == routingConfig.RampingDeploymentVersion.DeploymentName &&
+			version.DeploymentVersion.BuildId == routingConfig.RampingDeploymentVersion.BuildId {
```
jaypipes, line 156: is `version.DeploymentVersion` guaranteed to be non-nil? If not, let's add a nil-guard somewhere to this code.

Shivs11: yes! The versions being read here are from the version summaries list, which are fetched from the server. Moreover, we only add non-nil DeploymentVersions to that list in the server.

internal/temporal/worker_deployment.go
```diff
@@ -119,80 +127,60 @@ func GetWorkerDeploymentState(
...
-				versionInfo.DrainedSince = &drainedSince
-				// If the deployment exists and has replicas, we assume there are versioned pollers, no need to check
-				deployment, ok := k8sDeployments[version.Version.BuildID]
-				if !ok || deployment.Status.Replicas == 0 { //revive:disable-line:max-control-nesting
-					versionInfo.NoTaskQueuesHaveVersionedPoller = noTaskQueuesHaveVersionedPollers(ctx, client, desc.Info.TaskQueuesInfos)
-				}
-			} else if errors.As(err, &notFound) { //revive:disable-line:max-control-nesting
+			// If there is no controller-managed k8s Deployment for this drained version,
+			// treat it as NotRegistered so the planner can immediately clean up any
+			// orphaned resources rather than waiting for sunset delays.
```
Shivs11: same behaviour in today's world (without my PR)

api/v1alpha1/worker_types.go
```diff
@@ -308,9 +308,9 @@ type DeprecatedWorkerDeploymentVersion struct {
 	// +optional
 	DrainedSince *metav1.Time `json:"drainedSince,omitempty"`
 
-	// A Version is eligible for deletion if it is drained and has no pollers on any task queue.
-	// After pollers stop polling, the server will still consider them present until `matching.PollerHistoryTTL`
-	// has passed.
+	// A Version is considered eligible for deletion if it is drained and has no
+	// controller-managed workers with active replicas polling on its task queues.
+	// The server automatically deletes eligible versions when it needs to make room for new ones.
```
carlydf, line 313: Maybe add:
```
// Note: If this version has pollers that are not managed by the controller, the server 
// will not be able to delete the version. `EligibleForDeletion=true` does not consider
// that scenario, because it is rare and it is too expensive to cover (requires describing
// every drained version without controller-managed pollers and all task queues in 
// that version).
```

carlydf: or a shorter version of that!

internal/controller/state_mapper.go
```diff
@@ -159,7 +159,17 @@ func (m *stateMapper) mapDeprecatedWorkerDeploymentVersionByBuildID(buildID stri
 
 	eligibleForDeletion := false
 	if vInfo, exists := m.temporalState.Versions[buildID]; exists {
-		eligibleForDeletion = vInfo.Status == v1alpha1.VersionStatusDrained && vInfo.NoTaskQueuesHaveVersionedPoller
+		// A drained version is eligible for server-side auto-deletion only if its pollers
+		// have stopped. We approximate this by checking whether a controller-managed k8s
+		// Deployment with active replicas exists; if so, pollers are likely still running
+		// and the server cannot delete the version yet, making it ineligible for deletion.
+		// Note: This only considers pollers managed by the worker-controller. External
+		// pollers (e.g., manually deployed workers) are not accounted for.
```
carlydf: this note sounds great, maybe just copy it to where the `EligibleForDeletion` field is defined

internal/temporal/worker_deployment.go
```diff
@@ -94,23 +95,30 @@ func GetWorkerDeploymentState(
...
 	}
 	state.RampPercentage = routingConfig.RampingVersionPercentage
 	state.LastModifierIdentity = workerDeploymentInfo.LastModifierIdentity
 	state.VersionConflictToken = resp.ConflictToken
 
 	// Decide whether to ignore LastModifierIdentity
 	if state.LastModifierIdentity != controllerIdentity && state.LastModifierIdentity != "" {
-		state.IgnoreLastModifier, err = DeploymentShouldIgnoreLastModifier(ctx, deploymentHandler, routingConfig)
+		sdkRoutingConfig := toSDKRoutingConfig(routingConfig)
+		state.IgnoreLastModifier, err = DeploymentShouldIgnoreLastModifier(ctx, deploymentHandler, sdkRoutingConfig)
```
carlydf, line 121: I think this is needed in v1.2 and v1.3, but TODO remove in v1.4+ because we've switched to ManagerIdentity. That will allow us to eliminate even more DescribeVersion calls, since we no longer have to check metadata

carlydf: not a big deal though because `state.LastModifierIdentity != controllerIdentity && state.LastModifierIdentity != ""` is rare

internal/temporal/worker_deployment.go
```diff
@@ -119,80 +127,60 @@ func GetWorkerDeploymentState(
...
-				// If the deployment exists and has replicas, we assume there are versioned pollers, no need to check
-				deployment, ok := k8sDeployments[version.Version.BuildID]
-				if !ok || deployment.Status.Replicas == 0 { //revive:disable-line:max-control-nesting
-					versionInfo.NoTaskQueuesHaveVersionedPoller = noTaskQueuesHaveVersionedPollers(ctx, client, desc.Info.TaskQueuesInfos)
-				}
-			} else if errors.As(err, &notFound) { //revive:disable-line:max-control-nesting
+			// If there is no controller-managed k8s Deployment for this drained version,
+			// treat it as NotRegistered so the planner can immediately clean up any
+			// orphaned resources rather than waiting for sunset delays.
+			if _, ok := k8sDeployments[version.DeploymentVersion.BuildId]; !ok {
```
carlydf: I believe "NotRegistered" is a controller-side status that means "this version is not in the server-side versions list". Is there any issue here if we just stick with that definition?

internal/temporal/worker_deployment.go
carlydf: I will revert my manager id pr temporarily :)

internal/temporal/worker_deployment.go
```diff
@@ -160,14 +160,9 @@ func GetWorkerDeploymentState(
...
-			// orphaned resources rather than waiting for sunset delays.
-			if _, ok := k8sDeployments[version.DeploymentVersion.BuildId]; !ok {
-				versionInfo.Status = temporaliov1alpha1.VersionStatusNotRegistered
-			} else if version.DrainageInfo != nil && version.DrainageInfo.LastChangedTime != nil { //revive:disable-line:max-control-nesting
-				// Extract DrainedSince directly from the version summary's drainage info,
-				// avoiding a per-version DescribeVersion call.
+			// Extract DrainedSince directly from the version summary's drainage info,
+			// avoiding a per-version DescribeVersion call.
+			if version.DrainageInfo != nil && version.DrainageInfo.LastChangedTime != nil {
 				drainedSince := version.DrainageInfo.LastChangedTime.AsTime()
```
Shivs11, line 166: have also double checked if the drainedSince timestamp on the versionInfo is nil safe in our entire codebase, and it is

PR #235, opened 2026-03-20
feat: replace domain conditions with standard Ready/Progressing conditions

### LINKED ISSUE #198: [Feature Request] Add Kubernetes status.conditions (Ready/Progressing/Degraded) for TemporalConnection and TemporalWorkerDeployment for Argo CD health

We deploy Temporal Worker Controller CRDs via Argo CD. Argo CD can’t reliably determine when these resources are “Healthy/Ready”, so Applications assume those are healthy and we can’t gate sync waves or use `kubectl wait`.

Current issues:
- `TemporalConnection` effectively has no readiness signal (no meaningful status), so Argo CD can’t tell if the connection is usable.
- `TemporalWorkerDeployment` has rollout info under `status.targetVersion/currentVersion`, but no standard `status.conditions[]` that GitOps tooling can consume.

Example `TemporalWorkerDeployment` status we see:
```yaml
status:
  targetVersion:
    buildID: tag-c5fb
    deployment:
      apiVersion: apps/v1
      kind: Deployment
      name: my-deploy-tag-c5fb
      namespace: namespace_here
      uid: UUID4
    healthySince: '2026-02-12T07:15:09Z'
    status: NotRegistered
````

From Argo CD perspective, this is not a canonical “ready” signal, so we either accept Unknown health or write brittle custom Lua health checks that parse controller-specific fields.

Add standard Kubernetes Conditions to both CRDs so Argo CD (and kubectl/CI) can reason about readiness consistently.

#### TemporalConnection

Add:

* `status.conditions: []metav1.Condition`
* (recommended) `status.observedGeneration`

Semantics:

* `Ready=True` when the controller can successfully establish/authenticate a Temporal client using this connection.
* `Ready=False` with clear `reason/message` when:

  * referenced Secret missing/invalid
  * TLS config invalid
  * auth fails
  * dial/connect fails

#### TemporalWorkerDeployment

Add:

* `status.conditions: []metav1.Condition`
* (recommended) `status.observedGeneration`

Minimal condition set:

* `Ready`: True when an effective “current” version is registered in Temporal AND the backing K8s Deployment is healthy.
* `Progressing`: True while rollout is in-flight (target exists but not yet current / ramping / waiting on pollers / waiting on Deployment health).
* `Degraded`: True on reconcile errors that block progress.

Suggested mappings (examples):

* `targetVersion.status=NotRegistered` -> `Progressing=True`, `Ready=False`, reason `WaitingForPollers`
* backing Deployment not healthy / `healthySince` unset -> `Progressing=True`, reason `WaitingForDeployment`
* `currentVersion.status=Current` AND backing Deployment healthy -> `Ready=True`, `Progressing=False`

This enables:

* `kubectl wait --for=condition=Ready temporalconnection/<name>`
* `kubectl wait --for=condition=Ready temporalworkerdeployment/<name>`
* simple Argo CD health checks based only on `status.conditions` (stable API surface)

(Optional) Add kubebuilder printcolumns for READY + CURRENT/TARGET build IDs to improve `kubectl get` UX.

### Additional context

Argo CD supports custom health checks for CRDs, but they work best when controllers expose conventional `status.conditions[]`. This mirrors patterns used by other controllers (e.g., Crossplane resources) to integrate cleanly with Argo CD health assessment.

### PR DESCRIPTION
**Pull request by carlydf**
## What was changed
Deprecate the controller-specific TemporalConnectionHealthy and RolloutComplete conditions with the two standard Kubernetes conditions (Ready, Progressing) that GitOps tools like ArgoCD and Flux can consume without custom parsing logic.

- Ready=True when the target version is Current in Temporal
- Progressing=True while a rollout is in-flight (WaitingForPollers, WaitingForPromotion, Ramping)
- Progressing=False on any blocking error (connection failures, plan errors), regardless of whether a rollout is active

Also adds docs/cd-rollouts.md with integration guidance for kubectl, Helm 3/4, ArgoCD (Lua health check), and Flux (Kustomization + HelmRelease).

## Why?
Previous conditions were not easy for CD systems to consume. They were only merged to main, not included in any release, so we can still change them.

## Checklist

1. Closes #198 

2. How was this tested:
Functional and unit tests.

3. Any docs updates needed?

### DIFF
--- a/.github/workflows/helm.yml
+++ b/.github/workflows/helm.yml
@@ -4,14 +4,20 @@ on:
   workflow_dispatch:
     inputs:
       version_bump:
-        description: 'Type of version bump to perform'
+        description: 'Type of helm chart version bump to perform (none = publish as-is)'
         required: true
         default: 'patch'
         type: choice
         options:
+          - none
           - patch
           - minor
           - major
+      ref:
+        description: 'Branch or tag to publish from (defaults to main)'
+        required: false
+        default: 'main'
+        type: string
 
 permissions:
   contents: read
@@ -32,6 +38,7 @@ jobs:
         uses: actions/checkout@v4
         with:
           token: ${{ steps.generate_token.outputs.token }}
+          ref: ${{ inputs.ref }}
           fetch-depth: 0
 
       - name: Configure Git
@@ -50,8 +57,14 @@ jobs:
           VERSION_BUMP: ${{ inputs.version_bump }}
         run: |
           # Get current version from Chart.yaml
-          CURRENT_VERSION=$(grep 'version:' helm/temporal-worker-controller/Chart.yaml | awk '{print $2}')
+          CURRENT_VERSION=$(grep '^version:' helm/temporal-worker-controller/Chart.yaml | awk '{print $2}')
           echo "Current version: $CURRENT_VERSION"
+
+          if [[ "${VERSION_BUMP}" == "none" ]]; then
+            echo "No version bump requested, publishing existing version: $CURRENT_VERSION"
+            echo "version=$CURRENT_VERSION" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
 
           # Split version into parts
           IFS='.' read -r -a VERSION_PARTS <<< "$CURRENT_VERSION"
@@ -77,15 +90,17 @@ jobs:
           # Update Chart.yaml with new version
           sed -i "s/version: $CURRENT_VERSION/version: $NEW_VERSION/g" helm/temporal-worker-controller/Chart.yaml
 
-          # Also bump CRDs chart version
-          sed -i "s/version: $CURRENT_VERSION/version: $NEW_VERSION/g" helm/temporal-worker-controller-crds/Chart.yaml
+          # Also bump CRDs chart version if it exists
+          if [ -f helm/temporal-worker-controller-crds/Chart.yaml ]; then
+            sed -i "s/version: $CURRENT_VERSION/version: $NEW_VERSION/g" helm/temporal-worker-controller-crds/Chart.yaml
+            git add helm/temporal-worker-controller-crds/Chart.yaml
+          fi
 
           # Set output variable for use in later steps
           echo "version=$NEW_VERSION" >> "$GITHUB_OUTPUT"
 
-          # Commit both Chart.yaml files
+          # Commit Chart.yaml files
           git add helm/temporal-worker-controller/Chart.yaml
-          git add helm/temporal-worker-controller-crds/Chart.yaml
           git commit -m "Bump chart version to $NEW_VERSION [skip ci]"
           git push
 
@@ -102,10 +117,14 @@ jobs:
           VERSION=${{ steps.bump_version.outputs.version }}
           echo "Chart version: $VERSION"
 
-          # Package and push the CRDs chart
-          helm package ./helm/temporal-worker-controller-crds
-          helm push temporal-worker-controller-crds-${VERSION}.tgz oci://docker.io/temporalio
-          echo "✅ CRDs chart pushed successfully to oci://docker.io/temporalio/temporal-worker-controller-crds:${VERSION}"
+          # Package and push the CRDs chart if it exists
+          if [ -d "./helm/temporal-worker-controller-crds" ]; then
+            helm package ./helm/temporal-worker-controller-crds
+            helm push temporal-worker-controller-crds-${VERSION}.tgz oci://docker.io/temporalio
+            echo "✅ CRDs chart pushed successfully to oci://docker.io/temporalio/temporal-worker-controller-crds:${VERSION}"
+          else
+            echo "⚠️ CRDs chart not found, skipping"
+          fi
 
           # Package and push the controller chart
           helm package ./helm/temporal-worker-controller

### CODE REVIEW

File: docs/cd-rollouts.md [hunk 1]
```diff
@@ -0,0 +1,215 @@
+# CD Rollouts with the Temporal Worker Controller
```
Shivs11, line 1: nit: took me a second to realize CD meant "continous deployment", maybe let's have the full form be present in the start and then use the abbreviated form everywhere else later on

internal/controller/worker_controller.go
```diff
@@ -312,17 +313,59 @@ func (r *TemporalWorkerDeploymentReconciler) setCondition(
 	})
 }
 
-// recordWarningAndSetConditionFalse emits a warning event, sets a condition to False, and persists the status update.
-func (r *TemporalWorkerDeploymentReconciler) recordWarningAndSetConditionFalse(
+// syncConditions sets Ready, Progressing, and Degraded based on the current rollout state.
+// It must be called at the end of a successful reconcile (no errors) so that Degraded is
+// cleared and Progressing/Ready reflect the latest Temporal version status.
+func (r *TemporalWorkerDeploymentReconciler) syncConditions(twd *temporaliov1alpha1.TemporalWorkerDeployment) {
```
Shivs11, line 319: (correct me if I am wrong in my interpretation)

from what I understand, you have changed status recording from being imperative (updating the status of the TWD at the moment of promotion/any failure) to now updating it at the end of a reconcile loop, in this function.

In the happy path, I wonder then, if we are gonna be a lag of one reconcile that is present on the twd status object. A simple example would be that if we are promoting a version from inactive to current and have successfully done so, we would update the status on the twd to have ConditionReady to be false, which would not be 100% accurate. Thus, we could have something like this:

>   TargetVersion.Status: Inactive    ← stale (Temporal already has Current)
>   Ready: False                       ← wrong for this moment in time
>   Progressing: True                  ← wrong for this moment in time

Moreover, in the happy path, this auto-resolves since the next reconciliation loop would fix things as we would read a fresh value from the server. However, what happens if we crash/face an error during this reconciliation loop? I think then, we would never show in our TWD status that the previous version was indeed promoted to current successfully.

docs/cd-rollouts.md
```diff
@@ -0,0 +1,215 @@
...
+      containers:
+        - name: worker
+          image: my-registry/my-worker:v2.3.0
+```
+
+```bash
+kubectl apply -f twd.yaml
+```
+
+The controller picks up the change on the next reconcile loop (within seconds) and begins the rollout.
```
mnichols, line 79: Is the "reconcile loop" something we can link to for description? I cant recall if this is in the architecture docs or not...

docs/cd-rollouts.md
```diff
@@ -0,0 +1,215 @@
...
+spec:
+  template:
+    spec:
+      containers:
+        - name: worker
+          image: my-registry/my-worker:v2.3.0
+```
+
+```bash
+kubectl apply -f twd.yaml
```
mnichols, line 76: Or if you dont want to create a template or risk with apply, you can
```
kubectl patch temporalworkerdeployment my-worker-deployment \
  -n temporal-oms-processing \
  --type=merge \
  -p "{\"spec\":{\"template\":{\"spec\":{\"containers\":[{\"name\":\"worker\",\"image\":\"${IMAGE}\"}]}}}}"

```

docs/cd-rollouts.md
```diff
@@ -0,0 +1,215 @@
...
+
+```bash
+helm upgrade my-worker ./chart --values values.yaml
+kubectl wait temporalworkerdeployment/my-worker \
+  --for=condition=Ready \
+  --timeout=10m \
+  --namespace my-namespace
+```
+
+## ArgoCD
```
mnichols, line 118: I wonder if you shouldnt consider a `recipes` directory where you can explicitly have different vendors like Flux, Argo, etc general guidance? Seems like it'd be easier to discover if it is a separate file in that kind of directory.

docs/README.md
```diff
@@ -11,6 +11,9 @@ This documentation structure is designed to support various types of technical d
 
 ## Index
 
+### [CD Rollouts](cd-rollouts.md)
```
mnichols, line 14: I'd over-index on fully qualified names and avoid abbreviations except for inside paragraphs that have `CD` already defined.

internal/controller/execplan.go
```diff
@@ -202,8 +202,6 @@ func (r *TemporalWorkerDeploymentReconciler) updateVersionConfig(ctx context.Con
 				"Failed to set buildID %q as current version: %v", vcfg.BuildID, err)
 			return fmt.Errorf("unable to set current deployment version: %w", err)
 		}
-		r.setCondition(workerDeploy, temporaliov1alpha1.ConditionRolloutComplete, metav1.ConditionTrue,
```
carlydf, line 205: add: mark version as active, and update CurrentVersion / RampingVersion in the TWD.Status

internal/controller/worker_controller.go
```diff
@@ -312,17 +313,59 @@ func (r *TemporalWorkerDeploymentReconciler) setCondition(
...
+	// Happy path: clear Degraded unconditionally — reaching this point means no errors occurred.
+	r.setCondition(twd, temporaliov1alpha1.ConditionDegraded,
+		metav1.ConditionFalse, temporaliov1alpha1.ReasonAsExpected, "No reconciliation errors")
+
+	switch twd.Status.TargetVersion.Status {
```
carlydf, line 325: this is not up to date after exec plan. need to update TargetVersion.Status after exec plan if we want Ready/Progressing condition to be correct after exec plan

api/v1alpha1/worker_types.go
```diff
@@ -88,12 +88,28 @@ type TemporalWorkerDeploymentSpec struct {
...
+	ConditionProgressing = "Progressing"
+
+	// ConditionDegraded is True when a blocking error prevents the controller
+	// from making progress. The Reason field identifies the specific failure.
+	ConditionDegraded = "Degraded"
+)
+
+// Deprecated condition type constants. Maintained for backward compatibility with
+// monitoring and automation built against v1.3.x. Use Ready, Progressing, and Degraded
+// instead. These will be removed in the next major version.
```
jaypipes: Next major (API) Version of the CRDs? or next major version of the controller itself? :)

api/v1alpha1/worker_types.go
```diff
@@ -88,12 +88,28 @@ type TemporalWorkerDeploymentSpec struct {
...
+	// ArgoCD and Flux use this condition to gate deployment success.
+	ConditionReady = "Ready"
+
+	// ConditionProgressing is True while a rollout is actively in-flight —
+	// i.e., the target version has not yet been promoted to current.
+	ConditionProgressing = "Progressing"
+
+	// ConditionDegraded is True when a blocking error prevents the controller
+	// from making progress. The Reason field identifies the specific failure.
+	ConditionDegraded = "Degraded"
```
jaypipes: Recommend removing this Condition. It's not a Kubernetes standard condition and generally best practice to avoid mixing negative polarity Conditions with positive polarity Conditions. I think it's a better idea to use `ConditionProgressing` and when the value of the condition is false, set the Reason field to indicate an error that is blocking progress.

api/v1alpha1/worker_types.go
```diff
@@ -104,33 +120,50 @@ const (
...
+	// to current or ramping.
+	ReasonWaitingForPromotion = "WaitingForPromotion"
+
+	// ReasonRamping is set on ConditionProgressing=True when the target version is
+	// the ramping version and is receiving a configured percentage of new workflows.
+	ReasonRamping = "Ramping"
+
+	// ReasonAsExpected is set on ConditionDegraded=False to indicate that no
+	// reconciliation errors occurred.
+	ReasonAsExpected = "AsExpected"
```
jaypipes: recommend removing this along with ConditionDegraded.

api/v1alpha1/worker_types.go
```diff
@@ -104,33 +120,50 @@ const (
...
+
+	// ReasonTemporalConnectionNotFound is set on ConditionDegraded=True when the
+	// referenced TemporalConnection resource cannot be found.
```
jaypipes: Recommend putting this on ConditionProgressing=false as a reason why progress is blocked.

docs/cd-rollouts.md
carlydf: oo good idea. btw, I think we might address some of your feedback in a separate followup PR, so that we can get the code changes in this PR in today. but this feedback is very appreciated!

docs/cd-rollouts.md
```diff
@@ -8,7 +8,7 @@ For migration help, see [migration-to-versioned.md](migration-to-versioned.md).
 
 ## Understanding the conditions
 
-The `TemporalWorkerDeployment` resource exposes three standard conditions on `status.conditions` that CD tools and scripts can consume.
+The `TemporalWorkerDeployment` resource exposes two standard conditions on `status.conditions` that CD tools and scripts can consume.
```
Shivs11, line 11: @carlydf - would love your review on the cd-rollouts.md file changes specifically; i know we are gonna come back to Mark's comments later on, but would still love to have something sitting on main that is right in it's essence

docs/cd-rollouts.md
```diff
@@ -125,28 +125,25 @@ The two standard conditions (`Ready`, `Progressing`) keep the Lua simple — it
 # In your argocd-cm ConfigMap
 data:
   resource.customizations.health.temporal.io_TemporalWorkerDeployment: |
-    local hs = {status = "Progressing", message = "Waiting for conditions"}
+    local ready = nil
```
Shivs11, line 128: The Kubernetes API does not guarantee the order of entries in status.conditions. The original Lua script iterated the array and returned on the first match, which meant if Progressing=False (reason: RolloutComplete) appeared before Ready=True in the array, ArgoCD would report the resource as Degraded even though the rollout completed successfully.

The fix collects both conditions into local variables first, then evaluates them with a clear priority: Ready=True always wins. This makes the health check order-independent.

PR #238, opened 2026-03-23
Lower reconcile-loop log to debug level

carlydf:
## What was changed and why
"Running Reconcile loop" fires on every 10-second tick for every TWD, making it noise at the default info level. Use V(1) so it only appears when the operator runs with --zap-log-level=debug.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


Shivs11 (approved):
TIL there is no l.Debug on the logr and the only way to have this is to do l.V(1).Info, which is so ugly LOL

ty tho

Issue #55, opened 2025-06-25
Add Finalizer for Deployment Resources


carlydf:

carlydf:
https://github.com/temporalio/temporal-worker-controller/issues/166 possibly related

PR #240, opened 2026-03-24
fix: clean up Temporal server-side versioning data on TWD deletion

### LINKED ISSUE #166: Cleanup of Temporal deployments when TemporalWorker CRD is deleted
## Description

When a TemporalWorkerDeployment custom resource is deleted, the controller does not remove the corresponding Temporal worker deployment (and its versions) from the Temporal Server. Today the controller reconciles only creation/update paths, which leaves a stale deployment record behind.

##  Context

In my team's project we run 2 types of workers: one long-lived worker for workflow logic and a short-lived worker spun up per workflow run to process its activities (so we never share queues across runs). Each run creates a worker deployment, deletes the CRD once finished (usually within a day). With potentially thousands of workflows running daily, we can end up with many stale and short-lived worker deployments. 

##  Questions for the maintainers
 Was there a deliberate rationale for skipping deletion (e.g., audit or durability concerns)?
 Are there recommended scripts or best practices today for pruning stale worker deployments?
 With the model of short lived workers, besides database growth, are there other downsides—memory pressure, lookup costs, UI/API clutter? Any mitigation/guidance that would help teams with this pattern?
Does the “short lived workers deployments” pattern align with how the Worker Controller is intended to be used, or is there a better approach you recommend?

anujagrawal380:
- Add a finalizer to `TemporalWorkerDeployment` to run Temporal server-side cleanup before K8s deletion
- Add a finalizer to `TemporalConnection` to prevent it from being deleted while any TWD still references it
- On TWD deletion, set current version to unversioned, clear ramping version, and delete registered versions

## Problem

When a `TemporalWorkerDeployment` CRD is deleted (e.g., switching back to plain Deployments), the Temporal server retains the build ID routing configuration. The matching service continues routing new tasks to the deleted build ID's physical queue, while unversioned workers poll a different physical queue. Tasks sit in `Scheduled` state indefinitely with no errors.

A secondary race condition exists: Helm deletes both the `TemporalConnection` and `TWD` in the same upgrade. Without the connection, the controller cannot talk to Temporal to clean up. This is solved by adding a finalizer to the `TemporalConnection` that blocks its deletion until all referencing TWDs are gone.

## Changes

**`internal/controller/worker_controller.go`:**

**TWD finalizer (`temporal.io/worker-deployment-cleanup`):**
- Added to all TWD resources during normal reconciliation
- On deletion, triggers `handleDeletion()` which:
  1. Sets the current version to unversioned (`BuildID: ""`) -- the critical step that unblocks task dispatch
  2. Clears any ramping version
  3. Deletes all registered versions with `SkipDrainage: true`
  4. Attempts to delete the deployment record itself
  5. Removes the connection finalizer if no other TWDs reference it
  6. Removes its own finalizer, allowing K8s to complete deletion

**TemporalConnection finalizer (`temporal.io/connection-in-use`):**
- Added to the `TemporalConnection` during normal TWD reconciliation via `ensureConnectionFinalizer()`
- Prevents the connection from being deleted while any TWD still references it
- Removed by `removeConnectionFinalizerIfUnused()` during TWD deletion, after checking no other TWDs in the same namespace reference the connection
- Guarantees the connection is always available during TWD cleanup -- no race condition with Helm deleting both resources simultaneously

**RBAC updates:**
- Added `update;patch` verbs for `temporalconnections` (was `get;list;watch`)
- Added `update` verb for `temporalconnections/finalizers`

## Deletion flow

```
Helm upgrade (TWD disabled)
  |
  v
Helm deletes TWD CRD + TemporalConnection CRD simultaneously
  |
  +--> TemporalConnection: has finalizer, K8s sets DeletionTimestamp but blocks deletion
  |
  +--> TWD: has finalizer, K8s sets DeletionTimestamp, triggers Reconcile
         |
         v
       handleDeletion() runs:
         1. Fetches TemporalConnection (guaranteed to exist via finalizer)
         2. Connects to Temporal server
         3. Sets current version to unversioned
         4. Deletes versions
         5. Removes connection finalizer (no other TWDs reference it)
         6. Removes TWD finalizer
         |
         v
       TWD deleted by K8s
         |
         v
       TemporalConnection: no more finalizers, deleted by K8s
```

Issue #55 
Closes #166

### DIFF
--- a/helm/temporal-worker-controller/templates/rbac.yaml
+++ b/helm/temporal-worker-controller/templates/rbac.yaml
@@ -110,7 +110,15 @@ rules:
     verbs:
       - get
       - list
+      - patch
+      - update
       - watch
+  - apiGroups:
+      - temporal.io
+    resources:
+      - temporalconnections/finalizers
+    verbs:
+      - update
   - apiGroups:
       - temporal.io
     resources:
--- a/helm/temporal-worker-controller/templates/rbac.yaml
+++ b/helm/temporal-worker-controller/templates/rbac.yaml
@@ -107,6 +107,7 @@ rules:
       - temporal.io
     resources:
       - temporalconnections
+      - workerresourcetemplates
     verbs:
       - get
       - list
@@ -117,6 +118,7 @@ rules:
       - temporal.io
     resources:
       - temporalconnections/finalizers
+      - temporalworkerdeployments/finalizers
     verbs:
       - update
   - apiGroups:
@@ -134,28 +136,12 @@ rules:
   - apiGroups:
       - temporal.io
     resources:
-      - temporalworkerdeployments/finalizers
-    verbs:
-      - update
-  - apiGroups:
-      - temporal.io
-    resources:
       - temporalworkerdeployments/status
       - workerresourcetemplates/status
     verbs:
       - get
       - patch
       - update
-  - apiGroups:
-      - temporal.io
-    resources:
-      - workerresourcetemplates
-    verbs:
-      - get
-      - list
-      - patch
-      - update
-      - watch
   # GENERATED RULES END
   # Rules for managing resources attached via WorkerResourceTemplate.
   # Driven entirely by workerResourceTemplate.allowedResources in values.yaml.
--- a/internal/controller/worker_controller.go
+++ b/internal/controller/worker_controller.go
@@ -10,11 +10,13 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/go-logr/logr"
 	temporaliov1alpha1 "github.com/temporalio/temporal-worker-controller/api/v1alpha1"
 	"github.com/temporalio/temporal-worker-controller/internal/controller/clientpool"
 	"github.com/temporalio/temporal-worker-controller/internal/k8s"
 	"github.com/temporalio/temporal-worker-controller/internal/temporal"
 	"go.temporal.io/api/serviceerror"
+	sdkclient "go.temporal.io/sdk/client"
 	"google.golang.org/grpc/codes"
 	grpcstatus "google.golang.org/grpc/status"
 	appsv1 "k8s.io/api/apps/v1"
@@ -28,6 +30,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
@@ -44,6 +47,10 @@ const (
 
 	// wrtWorkerRefKey is the field index key for WorkerResourceTemplate by temporalWorkerDeploymentRef.name.
 	wrtWorkerRefKey = ".spec.temporalWorkerDeploymentRef.name"
+
+	// twdFinalizerName is the finalizer added to TemporalWorkerDeployment resources
+	// to ensure Temporal server-side versioning data is cleaned up before the CRD is deleted.
+	twdFinalizerName = "temporal.io/worker-deployment-cleanup"
 )
 
 // getAPIKeySecretName extracts the secret name from a SecretKeySelector
@@ -137,6 +144,35 @@ func (r *TemporalWorkerDeploymentReconciler) Reconcile(ctx context.Context, req 
 		// the next reconcile fires naturally when the TWD is created. If the List or
 		// status updates fail (transient API errors), return the error to requeue with backoff.
 		return ctrl.Result{}, r.markWRTsTWDNotFound(ctx, req.NamespacedName)
+	}
+
+	// Handle deletion: clean up Temporal server-side versioning data before allowing
+	// the CRD to be deleted. Without this, stale build ID routing persists in Temporal
+	// and prevents unversioned workers from picking up tasks on the same task queue.
+	if !workerDeploy.DeletionTimestamp.IsZero() {
+		if controllerutil.ContainsFinalizer(&workerDeploy, twdFinalizerName) {
+			l.Info("TemporalWorkerDeployment is being deleted, running cleanup")
+			if err := r.handleDeletion(ctx, l, &workerDeploy); err != nil {
+				l.Error(err, "failed to clean up Temporal server-side deployment data")
+				return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
+			}
+
+			// Cleanup succeeded, remove the finalizer so K8s can delete the resource
+			controllerutil.RemoveFinalizer(&workerDeploy, twdFinalizerName)
+			if err := r.Update(ctx, &workerDeploy); err != nil {
+				return ctrl.Result{}, err
+			}
+			l.Info("Temporal server-side cleanup complete, finalizer removed")
+		}
+		return ctrl.Result{}, nil
+	}
+
+	// Ensure finalizer is present on non-deleted resources
+	if !controllerutil.ContainsFinalizer(&workerDeploy, twdFinalizerName) {
+		controllerutil.AddFinalizer(&workerDeploy, twdFinalizerName)
+		if err := r.Update(ctx, &workerDeploy); err != nil {
+			return ctrl.Result{}, err
+		}
 	}
 
 	// TODO(jlegrone): Set defaults via webhook rather than manually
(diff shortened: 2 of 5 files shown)


internal/controller/worker_controller.go
```diff
@@ -38,6 +42,15 @@ const (
...
 	buildIDLabel   = "temporal.io/build-id"
+
+	// twdFinalizerName is the finalizer added to TemporalWorkerDeployment resources
+	// to ensure Temporal server-side versioning data is cleaned up before the CRD is deleted.
+	twdFinalizerName = "temporal.io/worker-deployment-cleanup"
+
+	// tcFinalizerName is the finalizer added to TemporalConnection resources to prevent
+	// them from being deleted while any TemporalWorkerDeployment still references them.
+	// This ensures the connection is available during TWD deletion cleanup.
+	tcFinalizerName = "temporal.io/connection-in-use"
```
jaypipes: I don't think you need separate finalizer strings/names for each resource kind. Having a single finalizer should work just fine:

```go
finalizerName = "finalizer.temporal.io"
```

internal/controller/worker_controller.go
```diff
@@ -296,6 +352,132 @@ func (r *TemporalWorkerDeploymentReconciler) Reconcile(ctx context.Context, req
...
+	if routingConfig.RampingVersion != nil {
+		l.Info("Clearing ramping version", "buildID", routingConfig.RampingVersion.BuildID)
+		if _, err := deploymentHandler.SetRampingVersion(ctx, sdkclient.WorkerDeploymentSetRampingVersionOptions{
+			BuildID:    "",
+			Percentage: 0,
+			Identity:   getControllerIdentity(),
+		}); err != nil {
+			l.Info("Failed to clear ramping version (may have been cleared by SetCurrentVersion)", "error", err)
+		}
+	}
```
jaypipes: maybe put the following in here to follow the example info log from above.

```go
} else {
	l.Info("No ramping version set, skipping clear ramping version")
}
```

jaypipes (commented):
@anujagrawal380 awesome contribution, thank you so much for this PR! I couple really minor comments below, but overall excellent work.

internal/controller/worker_controller.go
anujagrawal380: @jaypipes You are right that we can use one finalizer string for both resources. But using separate names is clearer, when you see `temporal.io/connection-in-use` on a TemporalConnection, you immediately know why it's there. With a generic `finalizer.temporal.io`, you'd have to dig into code to understand which controller added it and why. Please share your thoughts, we will make changes as per your direction.

jaypipes: > You are right that we can use one finalizer string for both resources. But using separate names is clearer, when you see `temporal.io/connection-in-use` on a TemporalConnection, you immediately know why it's there. With a generic `finalizer.temporal.io`, you'd have to dig into code to understand which controller added it and why. Please share your thoughts, we will make changes as per your direction.

Finalizers only have one purpose: to prevent the deletion of something before some cleanup/action can be taken. So, there's not really a point in having more than one string name for a finalizer in the temporal-worker-controller, IMHO. Note that all builtin Kubernetes controllers use the exact same string (`resource.kubernetes.io/delete-protection`) ([ref](https://github.com/kubernetes/kubernetes/blob/1ed5965382e09c1fdc8b363c153e9ebee0942481/staging/src/k8s.io/api/resource/v1/types.go#L33)) so I'd go with either `finalizer.temporal.io` or `temporal.io/delete-protection` to follow in that naming convention/practice.

anujagrawal380: Cool, makes sense. Making the change!

jaypipes (approved):
rock on :) nice work on this @anujagrawal380!

carlydf (changes requested):
we need integration tests for this before merging to main / including it in a release

carlydf (approved):

carlydf:
Hi @anujagrawal380 , could you fix the linters! Would love to include this in our next release

anujagrawal380:
> Hi @anujagrawal380 , could you fix the linters! Would love to include this in our next release

@carlydf @jaypipes Added few more minor improvements here: https://github.com/temporalio/temporal-worker-controller/pull/240/changes/9fd0c740bfa25a367e8f5b194dd18675bf7a7890 . PTAL!

internal/controller/worker_controller.go
```diff
@@ -41,6 +45,18 @@ const (
...
+	// resources to prevent deletion before cleanup actions are taken. On TWD resources, it
+	// ensures Temporal server-side versioning data is cleaned up. On TemporalConnection
+	// resources, it prevents deletion while any TWD still references the connection.
+	finalizerName = "temporal.io/delete-protection"
+
+	// deletionCleanupTimeout is the maximum duration to retry Temporal server-side
+	// cleanup before giving up and allowing the K8s resource to be deleted.
+	// This prevents the TWD from being stuck in Terminating state indefinitely
+	// if the Temporal server is unavailable or a version has persistent active pollers.
+	deletionCleanupTimeout = 5 * time.Minute
```
carlydf: I think the goal of this finalizer is for the kubernets / `TemporalWorkerDeployment` perspective and the Temporal server perspective to be aligned. So if the server-side object was created by creating the k8s-side object, the server-side object should also be deleted in the same way.

This timeout breaks that expectation. If the server is temporarily unavailable and this finalizer gives up and deletes the k8s-side object, the server-side object would never be deleted.

I'm curious if you ran into this while testing? The default TTL for "active pollers" in server is 5 minutes, so if when the `TWD` enters Terminating state it is running active pollers, the controller needs to kill those Deployments and then wait 5 minutes before the "no active pollers" check passes. If all versions were Drained and the pods had been scaled down for a while, this delay wouldn't exist. 

Because of that 5 minute poller TTL, a 5 minute `deletionCleanupTimeout` would frequently be used in case of deletion before natural scaledown, which would IMO not be good (because of the leftover server-side object thing I explained above).

If we decide to keep this, I would advocate for:
- A very long threshold, like 1h
- Only timing out on unavailable errors from the server, not precondition failed (which is the "active pollers" thing)

carlydf (changes requested):

internal/controller/worker_controller.go
carlydf: I will defer to @jaypipes opinion here though!

jaypipes: @carlydf @anujagrawal380 I would go further than Carly here and actually recommend removing this timeout entirely. I don't think there's actually anything wrong with a TWD "being stuck in Terminating state". If the Temporal server is unavailable or a version has persistent active pollers, we don't want to delete TWD resources (or the Kubernetes resources that back the TWD). Having the TWD continue to be in a Terminating state for an extended period of time is perfectly acceptable.

jaypipes: Keep in mind that if something is truly "stuck" the user can always force-delete the k8s resource anyway which would delete the resource from the API server, though the user might need to manually force-delete other owned resources in that case.

jaypipes (changes requested):
yeah, I agree with @carlydf's concern about the deleteCleanupTimeout. I think that should be removed.

carlydf:
@anujagrawal380 we hope to include this change in our release early next week, without the timeout addition. Hopefully you can respond to the feedback when you get back on Monday! To allow the tests to pass in a reasonable amount of time, you can override the pollerTTL to make pollers "die" faster on the server side.
[Example](https://github.com/temporalio/temporal-worker-controller/blob/main/internal/tests/internal/integration_test.go#L718)

If we don't get a response by Tuesday, we will pick your commits into a separate PR (so you will still be attributed), remove the timeout, make the tests pass and merge it. Hopefully we can come to an agreement about the timeout behavior though!

internal/controller/worker_controller.go
anujagrawal380: Makes Sense! Aligned, made the changes as suggested.

anujagrawal380:
> @anujagrawal380 we hope to include this change in our release early next week, without the timeout addition. Hopefully you can respond to the feedback when you get back on Monday! To allow the tests to pass in a reasonable amount of time, you can override the pollerTTL to make pollers "die" faster on the server side. [Example](https://github.com/temporalio/temporal-worker-controller/blob/main/internal/tests/internal/integration_test.go#L718)
> 
> If we don't get a response by Tuesday, we will pick your commits into a separate PR (so you will still be attributed), remove the timeout, make the tests pass and merge it. Hopefully we can come to an agreement about the timeout behavior though!

Thanks @carlydf @jaypipes for the thorough review. Made the changes as suggested. Please let me know if any more changes are required!

internal/controller/worker_controller.go
```diff
@@ -139,6 +149,41 @@ func (r *TemporalWorkerDeploymentReconciler) Reconcile(ctx context.Context, req
...
+		if controllerutil.ContainsFinalizer(&workerDeploy, finalizerName) {
+			l.Info("TemporalWorkerDeployment is being deleted, running cleanup")
+			if err := r.handleDeletion(ctx, l, &workerDeploy); err != nil {
+				l.Error(err, "failed to clean up Temporal server-side deployment data, will retry")
+				return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
+			}
+
+			// Remove our finalizer from the TemporalConnection if no other TWDs reference it.
+			if err := r.removeConnectionFinalizerIfUnused(ctx, l, &workerDeploy); err != nil {
+				l.Error(err, "failed to remove finalizer from TemporalConnection")
```
jaypipes: There's no reason to use `l.Error()` here because the controller-runtime logger will automatically log the error returned.

internal/controller/worker_controller.go
```diff
@@ -359,6 +411,134 @@ func (r *TemporalWorkerDeploymentReconciler) markWRTsTWDNotFound(ctx context.Con
...
+	// Step 2: If there's a ramping version, clear it.
+	if routingConfig.RampingVersion != nil {
+		l.Info("Clearing ramping version", "buildID", routingConfig.RampingVersion.BuildID)
+		if _, err := deploymentHandler.SetRampingVersion(ctx, sdkclient.WorkerDeploymentSetRampingVersionOptions{
+			BuildID:    "",
+			Percentage: 0,
+			Identity:   getControllerIdentity(),
+		}); err != nil {
+			l.Info("Failed to clear ramping version (may have been cleared by SetCurrentVersion)", "error", err)
+		}
```
jaypipes: @carlydf @Shivs11 should we be clearing the ramping version *before* setting the current version to unversioned?

internal/tests/internal/deletion_integration_test.go
```diff
@@ -0,0 +1,301 @@
...
+	}
+	if !hasFinalizer {
+		t.Fatalf("TWD does not have expected finalizer %q", deletionFinalizerName)
+	}
+
+	// Delete the TWD
+	t.Log("Deleting the TemporalWorkerDeployment")
+	if err := k8sClient.Delete(ctx, &twdBeforeDelete); err != nil {
+		t.Fatalf("failed to delete TWD: %v", err)
+	}
```
jaypipes, line 176: Before we delete the TWD, let's add another version to the TWD, set the ramping version to that new version. That way, we test the code paths that clear the ramping version on the TWD during deletion.

jaypipes (changes requested):

internal/tests/internal/deletion_integration_test.go
anujagrawal380: Makes sense

internal/controller/worker_controller.go
anujagrawal380: Cool

internal/controller/worker_controller.go
anujagrawal380: Does `SetCurrentVersion` with an empty BuildID guaranteed to clear the ramping version atomically on the server side?

jaypipes: @anujagrawal380 not sure :) that's why I'm asking @carlydf and @Shivs11

internal/tests/internal/deletion_integration_test.go
```diff
@@ -0,0 +1,360 @@
...
+		t.Fatalf("failed to describe deployment before setting ramping version: %v", err)
+	}
+	if _, err := deploymentHandle.SetRampingVersion(ctx, sdkclient.WorkerDeploymentSetRampingVersionOptions{
+		BuildID:       buildIDv2,
+		Percentage:    50,
+		ConflictToken: descResp.ConflictToken,
+		Identity:      "test",
+	}); err != nil {
+		t.Fatalf("failed to set v2.0 as ramping version: %v", err)
+	}
```
jaypipes: 👍

internal/controller/worker_controller.go
anujagrawal380: https://deepwiki.com/search/-step-2-if-theres-a-ramping-ve_d61611c4-01e5-432c-ae27-effd7e0ddcae?mode=fast

anujagrawal380: Deepwiki says no @jaypipes

jaypipes: @anujagrawal380 actually reading deepwiki there, it is saying that we *should* be clearing the ramping version before clearing the current version.

anujagrawal380: Yeah, went into details now. Lets swap steps 1 and 2 in handleDeletion, clear ramping first, then set current to unversioned.

internal/controller/worker_controller.go
```diff
@@ -359,6 +410,143 @@ func (r *TemporalWorkerDeploymentReconciler) markWRTsTWDNotFound(ctx context.Con
 	return errors.Join(errs...)
 }
 
+// handleDeletion cleans up Temporal server-side deployment versioning data when
+// a TemporalWorkerDeployment CRD is deleted. This prevents stale build ID routing
+// from blocking unversioned workers on the same task queue.
+//
+// The cleanup sequence:
+//  1. Set the current version to "unversioned" (empty BuildID) so new tasks route to unversioned workers
+//  2. Delete all non-current/non-ramping versions (drained/inactive ones)
```
jaypipes: Need to update these code comments to match the new sequence.

anujagrawal380: Resolved

jaypipes (approved):
excellent work on this @anujagrawal380 thank you so much! :)

helm/temporal-worker-controller/templates/rbac.yaml
```diff
@@ -110,7 +110,15 @@ rules:
...
       - list
+      - patch
+      - update
       - watch
+  - apiGroups:
+      - temporal.io
+    resources:
+      - temporalconnections/finalizers
+    verbs:
+      - update
```
carlydf, line 123: to get past the CI linter, add these permissions as kubebuilder annotations (I think [here](https://github.com/temporalio/temporal-worker-controller/blob/main/internal/controller/worker_controller.go#L101-L111) but I could be wrong ) and run `make manifests`

carlydf (approved):
thank you!

helm/temporal-worker-controller/templates/rbac.yaml
anujagrawal380: Done

Shivs11 (approved):
ty!

anujagrawal380:
@carlydf Can we please rerun tests again here?

anujagrawal380:
@carlydf Sorry, lint failed. Please again!

carlydf:
> @carlydf Sorry, lint failed. Please again!

@anujagrawal380, this test failed in https://github.com/temporalio/temporal-worker-controller/actions/runs/25011104275/job/73248144790. I'm re-running to see if it was just flaky and maybe will pass the second time. At the same time, ideally we would not have a flaky test since that will cause issues for future PRs. Could you take a look at the failure please?
```
    --- FAIL: TestIntegration/deletion-sets-current-to-unversioned (5.61s)
```

anujagrawal380:
> > @carlydf Sorry, lint failed. Please again!
> 
> @anujagrawal380, this test failed in https://github.com/temporalio/temporal-worker-controller/actions/runs/25011104275/job/73248144790. I'm re-running to see if it was just flaky and maybe will pass the second time. At the same time, ideally we would not have a flaky test since that will cause issues for future PRs. Could you take a look at the failure please?
> 
> ```
>     --- FAIL: TestIntegration/deletion-sets-current-to-unversioned (5.61s)
> ```

@carlydf Added a commit, please rerun once more!

carlydf:
> > > @carlydf Sorry, lint failed. Please again!
> > 
> > 
> > @anujagrawal380, this test failed in https://github.com/temporalio/temporal-worker-controller/actions/runs/25011104275/job/73248144790. I'm re-running to see if it was just flaky and maybe will pass the second time. At the same time, ideally we would not have a flaky test since that will cause issues for future PRs. Could you take a look at the failure please?
> > ```
> >     --- FAIL: TestIntegration/deletion-sets-current-to-unversioned (5.61s)
> > ```
> 
> @carlydf Added a commit, please rerun once more!

@anujagrawal380 it failed again. Have not looked at why. Is this passing locally? To run locally you can do:

```
KUBEBUILDER_ASSETS=/Users/cdf/Desktop/temporal-worker-controller/bin/k8s/1.27.1-darwin-arm64 go test -tags test_dep ./internal/tests/internal
   -run "TestIntegration/deletion-sets-current-to-unversioned" -timeout 600s 
```
but replace the `KUBEBUILDER_ASSETS` env var with the correct path for your system. To generate those assets and find out the right path, you can run `make test-integration` and then Ctrl+C the test after the KUBEBUILDER_ASSETS line is printed.
```
 % make test-integration
test -s /Users/cdf/Desktop/temporal-worker-controller/bin/controller-gen && /Users/cdf/Desktop/temporal-worker-controller/bin/controller-gen --version | grep -q v0.19.0 || \
        GOBIN=/Users/cdf/Desktop/temporal-worker-controller/bin go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.19.0
GOWORK=off GO111MODULE=on /Users/cdf/Desktop/temporal-worker-controller/bin/controller-gen rbac:roleName=manager-role crd:allowDangerousTypes=true,maxDescLen=0,generateEmbeddedObjectMeta=true paths=./api/... paths=./internal/... paths=./cmd/... \
    output:crd:artifacts:config=helm/temporal-worker-controller-crds/templates
python3 hack/sync-rbac-rules.py
Synced RBAC rules from config/rbac/role.yaml → helm/temporal-worker-controller/templates/rbac.yaml
GOWORK=off GO111MODULE=on /Users/cdf/Desktop/temporal-worker-controller/bin/controller-gen object:headerFile="hack/boilerplate.go.txt" paths=./api/... paths=./internal/... paths=./cmd/...
test -s /Users/cdf/Desktop/temporal-worker-controller/bin/setup-envtest || GOBIN=/Users/cdf/Desktop/temporal-worker-controller/bin go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
Running integration tests...
KUBEBUILDER_ASSETS="/Users/cdf/Desktop/temporal-worker-controller/bin/k8s/1.27.1-darwin-arm64" go test -v -tags test_dep ./internal/tests/internal -run TestIntegration
^Cmake: *** [test-integration] Interrupt: 2
```

PR #242, opened 2026-03-24
update the helm.yml workflow to now publish helm charts from feature branches without bumping up the chart version

Shivs11:
## What was changed
- WISOTT

## Why?
- We want to be able to now publish a helm chart from a tag, which shall be cut from a release branch, so that we can fix the divergence between our appVersion and our helm chart version.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/internal/controller/clientpool/clientpool.go
+++ b/internal/controller/clientpool/clientpool.go
@@ -58,13 +58,24 @@ type ClientPool struct {
 	logger    log.Logger
 	clients   map[ClientPoolKey]ClientInfo
 	k8sClient runtimeclient.Client
+	// dialFn establishes a Temporal SDK connection from the given options. In production
+	// this is sdkclient.Dial; in tests it can be replaced with a function that returns a
+	// mock client without making any network calls.
+	dialFn func(sdkclient.Options) (sdkclient.Client, error)
+	// systemCertPoolFn loads the host OS certificate pool used as the base when a custom
+	// CA is appended via ca.crt. In production this is x509.SystemCertPool; in tests it
+	// can be replaced to inject a known set of "system" root CAs without depending on the
+	// OS trust store.
+	systemCertPoolFn func() (*x509.CertPool, error)
 }
 
 func New(l log.Logger, c runtimeclient.Client) *ClientPool {
 	return &ClientPool{
-		logger:    l,
-		clients:   make(map[ClientPoolKey]ClientInfo),
-		k8sClient: c,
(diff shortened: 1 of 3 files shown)


.github/workflows/helm.yml
```diff
@@ -50,9 +57,15 @@ jobs:
           VERSION_BUMP: ${{ inputs.version_bump }}
         run: |
           # Get current version from Chart.yaml
-          CURRENT_VERSION=$(grep 'version:' helm/temporal-worker-controller/Chart.yaml | awk '{print $2}')
+          CURRENT_VERSION=$(grep '^version:' helm/temporal-worker-controller/Chart.yaml | awk '{print $2}')
```
Shivs11, line 60: i'm just being as explicit as i can here by choosing the 'version' rather than the appVersion in the chart.yaml file.

.github/workflows/helm.yml
```diff
@@ -77,15 +90,17 @@ jobs:
           # Update Chart.yaml with new version
           sed -i "s/version: $CURRENT_VERSION/version: $NEW_VERSION/g" helm/temporal-worker-controller/Chart.yaml
 
-          # Also bump CRDs chart version
-          sed -i "s/version: $CURRENT_VERSION/version: $NEW_VERSION/g" helm/temporal-worker-controller-crds/Chart.yaml
+          # Also bump CRDs chart version if it exists
```
Shivs11, line 93: remembered that in this very specific case of publishing the helm chart from the to-be cut chart-0.20 tag, the repo checked out at the tag would not have these sep crds, which would fail this step if we did not have an if condition here.

.github/workflows/helm.yml
```diff
@@ -77,15 +90,17 @@ jobs:
...
+            git add helm/temporal-worker-controller-crds/Chart.yaml
+          fi
 
           # Set output variable for use in later steps
           echo "version=$NEW_VERSION" >> "$GITHUB_OUTPUT"
 
-          # Commit both Chart.yaml files
+          # Commit Chart.yaml files
           git add helm/temporal-worker-controller/Chart.yaml
-          git add helm/temporal-worker-controller-crds/Chart.yaml
```
carlydf, line 88: if the CRDs directory exists, we should git add it

Shivs11: Line 96 in my change does add it - we just do a git add only if the dir exists

PR #255, opened 2026-04-01
Fix greedy sed in release.yml and restore cert-manager constraint

Shivs11:
## Summary
- Restores cert-manager dependency version constraint to `">=v1.0.0"` (was overwritten to `0.24.0` by the `v1.5.1` release pipeline)
- Fixes the root cause: `sed` commands in `release.yml` used `s/version: .*/...` which matched **all** `version:` lines, including indented dependency versions. Now anchored with `^` to only match the top-level chart version
- Updates `Chart.lock` to be in sync with the restored `Chart.yaml`

## Context
When the `v1.5.1` release was published, `release.yml`'s helm job:
1. Overwrote the cert-manager constraint from `">=v1.0.0"` to `0.24.0` (same bug as #252)
2. `helm package` then failed because cert-manager `0.24.0` doesn't exist in the jetstack repo
3. The controller chart `0.24.0` was **never published**, but the CRDs chart `0.24.0` was

## After merge
Trigger `helm.yml` manually with `version_bump: none`, `ref: main` to publish the controller chart `0.24.0` with `appVersion: 1.5.1`.

## Test plan
- [ ] CI passes
- [ ] After merge, trigger `helm.yml` with `version_bump: none`, `ref: main`
- [ ] Verify controller chart `0.24.0` is published to Docker Hub

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -96,7 +96,7 @@ jobs:
         id: bump_version
         run: |
           # Get current version from Chart.yaml
-          CURRENT_VERSION=$(grep 'version:' helm/temporal-worker-controller/Chart.yaml | awk '{print $2}')
+          CURRENT_VERSION=$(grep '^version:' helm/temporal-worker-controller/Chart.yaml | awk '{print $2}')
           echo "Current version: $CURRENT_VERSION"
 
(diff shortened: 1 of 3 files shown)


jaypipes (approved):
makes sense.

Issue #263, opened 2026-04-07
Cannot delete WorkerResourceTemplate due to no permissions


jinjiaKarl:
`The WorkerResourceTemplate "*" is invalid: spec.template: Forbidden: requesting user "xxxxx" is not authorized to delete ScaledObject in namespace "xxx": permission denied
`

We are using GKE, and I have Kubernetes Engine Admin role. The validation webhook only checks if RBAC/SubjectAccessReview is applied to user properly. Is it possible to delete WorkerResourceTemplate without adding RBAC roles?

```
> kubectl auth can-i delete ScaledObject -n x
yes
```

jinjiaKarl:
The detailed info you can find is in Temporal Community Slack, https://temporalio.slack.com/archives/C07MDJ6S3HP/p1775564362647909. Adding the Extra field can leverage the GKE IAM permissions.

PR #265, opened 2026-04-08
Add extra field when doing SubjectAccessReview

jinjiaKarl:
## What was changed

Pass UserInfo.Extra fields when constructing the SubjectAccessReview for the requesting user in the WorkerResourceTemplate validating webhook.

## Why?

Previously, the webhook only forwarded User and Groups to the SAR spec. This omitted the Extra field, which cloud providers like GKE use to carry IAM identity information (e.g. iam.gke.io/user-assertion). Without it, the SAR could not evaluate permissions granted via GKE IAM or GCP group membership, only direct RBAC bindings to a specific username were recognized.

## Checklist

1. Closes  #263

2. How was this tested:

deployed to one GKE cluster, people with GKE IAM Admin permission can create/update/delete WorkerResourceTemplate resource without granting Kubernetes built-in RBAC bindings

3. Any docs updates needed?

### DIFF
--- a/api/v1alpha1/workerresourcetemplate_webhook.go
+++ b/api/v1alpha1/workerresourcetemplate_webhook.go
@@ -470,7 +470,11 @@ func (v *WorkerResourceTemplateValidator) validateWithAPI(ctx context.Context, w
 			Spec: authorizationv1.SubjectAccessReviewSpec{
 				User:   req.UserInfo.Username,
 				Groups: req.UserInfo.Groups,
-				Extra:  convertUserInfoExtra(req.UserInfo.Extra),
+				// Some authentication plugins like GKE's IAM plugin rely on certain fields being
+				// present in the UserInfo.Extra field, so here we make sure to copy any extra
+				// field values from the authentication request into the extra field of the
+				// authorization review.
+				Extra: convertUserInfoExtra(req.UserInfo.Extra),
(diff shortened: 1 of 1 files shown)


api/v1alpha1/workerresourcetemplate_webhook.go
```diff
@@ -456,6 +470,7 @@ func (v *WorkerResourceTemplateValidator) validateWithAPI(ctx context.Context, w
 			Spec: authorizationv1.SubjectAccessReviewSpec{
 				User:   req.UserInfo.Username,
 				Groups: req.UserInfo.Groups,
+				Extra:  convertUserInfoExtra(req.UserInfo.Extra),
```
jaypipes: Let's add a little comment above here to let our future selves know why we added this :)

```go
// Some authentication plugins like GKE's IAM plugin rely on certain fields being
// present in the UserInfo.Extra field, so here we make sure to copy any extra
// field values from the authentication request into the extra field of the
// authorization review.
```

jaypipes (commented):
thanks very much for this @jinjiaKarl! one tiny little request for a code comment, otherwise looks fantastic. :)

jaypipes (approved):
👍

Shivs11 (approved):

PR #274, opened 2026-04-14
[FR-126] Rollback

### LINKED ISSUE #126: [Feature Request] Rollback mode

For users who use ramping.
Having upgraded from v1 to v2 and set v2 to current, some users would want to instantly roll back to v1 if they discover an issue. 

Can allow users to configure a rollout strategy for rollbacks?

### Additional context

Currently there is only a rollout strategy. 

Could do:
- Check if version is in DeprecatedVersions (still in sunset) **and that version was previously a CurrentVersion**
  - Temporal server provides `LastDeactivationTime` and `LastActivationTime` but those are not enough because they include if the version was only ramping. Check `LastCurrentTime` != nil to confirm that it was nil.
- If still in sunset, could automatically set to AllAtOnce. 
  - No need to intertwine these.
- Do we want to provide a parameter such as "fastRollbackAgeLimit" or just default to 24h?
  - 1 vote for default to 24h
  - Idea here is that a much older version is more likely to be very different from what the running workflows expect.
- Do we want to only support rolling one version back?
  - We can add a limit like this if requested. To start with, any previously-Current version within default of 24h age is eligible for #fastrollback

Open questions:

- If we change LastCurrentTime to be the last time a version _was_ current, instead of the last time a version _was set to current_, the 24h `fastRollbackAgeLimit` makes sense. --> Will do in a future server release.

- Should we add a knob for `fastRollbackAgeLimit` --> yes, and default to 1h

- What should the default rollback strategy be? --> AllAtOnce with a 1h time limit. Disable rollback if rollout mode is manual.

- [Add to a separate doc on rollback] Operator might want to decide to pin all AutoUpgrade workflows that are running on the bad version to the bad version, so that they don't roll back to the previous version. (Or workflow pause?)

eniko-dif:
## What was changed

Detect when a target version was current within the last 1h, and if so, do an immediate rollback to this version (AllAtOnce rollout).

## Why?

Open feature issue: https://github.com/temporalio/temporal-worker-controller/issues/126

## Checklist

1. Closes #126

2. How was this tested:

Unit and integration tests (ongoing actual test with local setup).

3. Any docs updates needed?

Added in `configuration.md`.


api/v1alpha1/workerdeployment_types.go
```diff
@@ -83,6 +83,10 @@ type TemporalWorkerDeploymentSpec struct {
 	// How to rollout new workflow executions to the target version.
 	RolloutStrategy RolloutStrategy `json:"rollout"`
 
+	// How to rollback to a previous version. If not specified, defaults to AllAtOnce strategy.
+	// +optional
+	RollbackStrategy *RollbackStrategy `json:"rollback,omitempty"`
```
jlegrone: I think it would be helpful to add a comments clarifying a couple things:
1. How/when a rollback is triggered. I think right now it's only when the worker's pod spec is updated, AND the build ID has been previously set as the default worker version?
2. How rollback strategy applies to new vs. currently running workflows. Ie. running workflows that are pinned will remain on the version we are rolling back from, while new workflows will start on the rollback version.

api/v1alpha1/worker_types.go
```diff
@@ -355,6 +359,18 @@ const (
...
+// DefaultVersionRollbackStrategy describes how to cut over during rollback to a previous version.
+// +kubebuilder:validation:Enum=AllAtOnce;Progressive
+type DefaultVersionRollbackStrategy string
+
+const (
+	// RollbackAllAtOnce immediately switches 100% of traffic back to the previous version.
+	RollbackAllAtOnce DefaultVersionRollbackStrategy = "AllAtOnce"
+
+	// RollbackProgressive gradually ramps traffic back to the previous version.
+	RollbackProgressive DefaultVersionRollbackStrategy = "Progressive"
```
jlegrone: I think if we have a progressive rollout mode, we should also require that there be at least 1 step defined. It could be reasonable to assume that choosing the progressive rollout mode without also adding steps would reuse the existing steps from the rollout config. Without steps, progressive and all at once modes would behave the same way, right?

api/v1alpha1/worker_types.go
```diff
@@ -355,6 +359,18 @@ const (
 	UpdateProgressive DefaultVersionUpdateStrategy = "Progressive"
 )
 
+// DefaultVersionRollbackStrategy describes how to cut over during rollback to a previous version.
+// +kubebuilder:validation:Enum=AllAtOnce;Progressive
+type DefaultVersionRollbackStrategy string
```
jlegrone: Nit: we have a mix of existing styles between `SunsetStrategy` and `DefaultVersionUpdateStrategy`, but I think I prefer dropping the `Default` prefix.

```suggestion
type VersionRollbackStrategy string
```

internal/planner/planner.go
```diff
@@ -759,17 +761,49 @@ func getTestWorkflows(
...
+	targetVersionInfo, exists := temporalState.Versions[status.TargetVersion.BuildID]
+	if !exists {
+		return false
+	}
+
+	if targetVersionInfo.LastCurrentTime == nil {
+		return false
+	}
+
+	l.Info("Detected rollback scenario using LastCurrentTime. "+
```
jlegrone: I think we should add an additional condition to only consider it a rollback if `LastCurrentTime` is recent (maybe defaulting to <24h before now). Otherwise accidentally deploying a version from weeks ago could skip the normal rollout steps, which might be unexpected.

api/v1alpha1/workerdeployment_types.go
```diff
@@ -396,6 +412,21 @@ type RolloutStrategy struct {
 	Steps []RolloutStep `json:"steps,omitempty" protobuf:"bytes,3,rep,name=steps"`
 }
 
+// RollbackStrategy defines strategy to apply when rolling back to a previous version.
+// This is separate from RolloutStrategy because rollbacks have different requirements:
+// - No gate workflow (already trusted version)
+// - No manual mode (rollbacks should be automatic)
+// - Default to AllAtOnce for fast recovery
+type RollbackStrategy struct {
```
jlegrone: Can we add a `MaxRollbackAge` field (struggling to come up with a better name)? I think a reasonable default would be 24h. If the `LastCurrentTime` is less than `time.Now() - MaxRollbackAge`, then the controller should treat the deployment as a standard rollout rather than a rollback.

api/v1alpha1/workerdeployment_webhook.go
```diff
@@ -57,6 +57,11 @@ func (s *TemporalWorkerDeploymentSpec) Default(ctx context.Context) error {
 		s.SunsetStrategy.DeleteDelay = &v1.Duration{Duration: defaults.DeleteDelay}
 	}
 
+	if s.RollbackStrategy == nil {
+		s.RollbackStrategy = &RollbackStrategy{Strategy: RollbackAllAtOnce}
```
jlegrone: This is a change in behavior that should be called out in release notes.

Is there a way to disable rollbacks and keep the existing behavior of treating every version change as a new rollout? This should be documented as well.

api/v1alpha1/workerdeployment_types.go
```diff
@@ -396,6 +412,21 @@ type RolloutStrategy struct {
...
+	// Strategy for rollback. Valid values are "AllAtOnce" or "Progressive".
+	// Defaults to "AllAtOnce" for fast recovery.
+	Strategy DefaultVersionRollbackStrategy `json:"strategy"`
+
+	// Steps to execute progressive rollbacks. Only required when strategy is "Progressive".
+	// +optional
+	Steps []RolloutStep `json:"steps,omitempty"`
```
jlegrone: I'm not quite convinced that progressive rollbacks are necessary. What are the reasons to prefer a progressive rollback vs. disabling rollback and using the progressive rollout strategy?

It's probably not a big deal to maintain this functionality, but it also looks straightforward to add in a future release without introducing breaking changes.

api/v1alpha1/workerdeployment_webhook.go
eniko-dif: Now it's possible to disable the rollback by setting the max version age to 0. I've also updated the docs.

api/v1alpha1/workerdeployment_types.go
eniko-dif: This is done now, but with 1h default as per the meeting last week.

api/v1alpha1/workerdeployment_types.go
eniko-dif: As per our discussion last week, it could still be meaningful to have different progressive steps (for example 10% for a few minutes to verify that workflows start properly) and then bump to 100%).

internal/planner/planner.go
eniko-dif: It's do now with max version age.

api/v1alpha1/worker_types.go
eniko-dif: The validation is the same as for rollout, so there cannot be a no-step progressive strategy (done in the webhook). "It could be reasonable to assume that choosing the progressive rollout mode without also adding steps would reuse the existing steps from the rollout config." I would not assume this, it seems like a nasty side-effect, but if you think it's a good way forward, I can implement it.

internal/planner/planner.go
```diff
@@ -759,17 +761,66 @@ func getTestWorkflows(
...
+	if !exists {
+		return false
+	}
+
+	if targetVersionInfo.LastCurrentTime == nil {
+		return false
+	}
+
+	if config.RollbackStrategy.MaxVersionAge != nil && time.Since(*targetVersionInfo.LastCurrentTime) > config.RollbackStrategy.MaxVersionAge.Duration {
+		l.Info("Skipping rollback: the version's last current time exceeds MaxVersionAge",
```
jlegrone:
```suggestion
		l.Debug("Skipping rollback: the version's last current time exceeds MaxVersionAge",
```

internal/planner/planner.go
```diff
@@ -759,17 +761,66 @@ func getTestWorkflows(
...
+			"targetBuildID", status.TargetVersion.BuildID,
+			"lastCurrentTime", targetVersionInfo.LastCurrentTime,
+			"maxVersionAge", config.RollbackStrategy.MaxVersionAge.Duration)
+		return false
+	}
+
+	l.Info("Detected rollback scenario using LastCurrentTime. "+
```
jlegrone:
```suggestion
	l.Debug("Detected rollback scenario using LastCurrentTime. "+
```

api/v1alpha1/temporalworker_webhook.go
```diff
@@ -94,41 +102,20 @@ func validateForUpdateOrCreate(old, new *TemporalWorkerDeployment) (admission.Wa
 	}
 
 	allErrs = append(allErrs, validateRolloutStrategy(new.Spec.RolloutStrategy)...)
+	allErrs = append(allErrs, validateRollbackStrategy(*new.Spec.RollbackStrategy)...)
```
jlegrone: Could `new.Spec.RollbackStrategy` ever be nil? Should probably add a nil check here or update the type signature for `validateRollbackStrategy` instead of dereferencing.

api/v1alpha1/workerdeployment_types.go
```diff
@@ -396,6 +424,27 @@ type RolloutStrategy struct {
...
+	// Defaults to "AllAtOnce" for fast recovery.
+	Strategy VersionRollbackStrategy `json:"strategy"`
+
+	// Steps to execute progressive rollbacks. Only required when strategy is "Progressive".
+	// +optional
+	Steps []RolloutStep `json:"steps,omitempty"`
+
+	// MaxVersionAge limits which versions are eligible as rollback targets.
+	// A version is only considered a rollback target if it was last current within this duration.
+	// If nil, there is no age limit.
```
jlegrone: The default applied in the webhook is 1h though, so this comment doesn't match current behavior. We should also document the meaning of setting to 0.

Right now I don't think there is a way to achieve the "no age limit" behavior, beyond setting a very high value? (I think this is fine, but just clarifying since it seems like you can either specify a limit or disable rollback entirely by setting to 0).

api/v1alpha1/workerdeployment_webhook.go
```diff
@@ -57,6 +57,14 @@ func (s *TemporalWorkerDeploymentSpec) Default(ctx context.Context) error {
...
+	} else if s.RollbackStrategy.Strategy == "" {
+		s.RollbackStrategy.Strategy = RollbackAllAtOnce
+	}
```
jlegrone: Nit: this is entirely a matter of taste but I think two `if` statements is easier to read than the `if/else` flow (more clear that the strategy is always all at once if left unset).

```suggestion
	if s.RollbackStrategy == nil {
		s.RollbackStrategy = &RollbackStrategy{}
	}
	if s.RollbackStrategy.Strategy == "" {
		s.RollbackStrategy.Strategy = RollbackAllAtOnce
	}
```

internal/planner/planner.go
```diff
@@ -840,6 +891,25 @@ func getVersionConfigDiff(
...
+// Convert to reuse rollout logic with different settings
+func convertRollbackToRolloutStrategy(rb temporaliov1alpha1.RollbackStrategy) temporaliov1alpha1.RolloutStrategy {
+	var strategy temporaliov1alpha1.VersionRolloutStrategy
+	switch rb.Strategy {
+	case temporaliov1alpha1.RollbackAllAtOnce:
+		strategy = temporaliov1alpha1.UpdateAllAtOnce
+	case temporaliov1alpha1.RollbackProgressive:
+		strategy = temporaliov1alpha1.UpdateProgressive
+	default:
+		strategy = temporaliov1alpha1.UpdateAllAtOnce
```
jlegrone: 🤔 I think this ought to be an error condition; if the controller sees an unknown strategy it should fall back to manual mode and stop attempting updates to the default version.

internal/planner/planner_test.go
```diff
@@ -3581,3 +3581,448 @@ func TestGetWRTOwnerRefPatches(t *testing.T) {
...
+
+	defaultConfig := &Config{
+		RolloutStrategy: temporaliov1alpha1.RolloutStrategy{Strategy: temporaliov1alpha1.UpdateAllAtOnce},
+		RollbackStrategy: &temporaliov1alpha1.RollbackStrategy{
+			Strategy:      temporaliov1alpha1.RollbackAllAtOnce,
+			MaxVersionAge: &metav1.Duration{Duration: defaults.RollbackMaxVersionAge},
+		},
+	}
+
+	testCases := []struct {
```
jlegrone, line 3698: I think there should be a test cases for

1.`RollbackStrategy == nil` (even though the webhook should set a default, it's still possible to run the controller without the webhook enabled). When `nil` I would assert that the behavior matches the default rollback strategy configured by the webhook.
2. `RollbackStrategy.MaxVersionAge == nil` (assert default max version age applied, again to match the default webhook behavior)
3. `RollbackStrategy.MaxVersionAge == 0` (assert false, even when the rest of the current status would normally have resulted in a rollback)

api/v1alpha1/workerdeployment_webhook_test.go
```diff
@@ -23,8 +24,9 @@ func TestTemporalWorkerDeployment_ValidateCreate(t *testing.T) {
 	tests := map[string]struct {
 		obj      runtime.Object
 		errorMsg string
+		warnMsg  string
```
jlegrone: Should there be a warning emitted for a combination of Manual rollout + Progressive rollback? That seems in the same vein as rollout steps being faster than rollback steps. Please add a test case for that if so.

internal/planner/planner.go
eniko-dif: The logger has no Debug level, only Info and Error. (If I remember correctly, it was related to K8s events?)

api/v1alpha1/temporalworker_webhook.go
eniko-dif: The rollback strategy is a pointer, because it's not mandatory in the spec. At the point of the validation it should not be nil, because of the default we set, but just in case, I added a nil check.

api/v1alpha1/workerdeployment_webhook.go
eniko-dif: I've done the change, but I remember the times, where I had to optimize even the number of if calls... :P

internal/planner/planner.go
eniko-dif: Before this function is used, there is a check to see if the rollout strategy is set to manual, and if it is, rollback is disabled. In any case, I updated the default to be manual.

internal/planner/planner_test.go
eniko-dif: I changed the Config to have rollback strategy at all times, so the first two are no longer needed and the third existed already.

api/v1alpha1/workerdeployment_webhook_test.go
eniko-dif: When rollback is Manual, rollback is disabled in the planner, so I don't think we need a warning.

api/v1alpha1/workerdeployment_types.go
```diff
@@ -83,6 +83,22 @@ type WorkerDeploymentSpec struct {
...
+	//
+	// A rollback is triggered automatically when the target version's pod spec is updated and
+	// the resulting build ID has previously been set as the default (current) version of the
+	// worker deployment. The controller detects this by checking whether Temporal recorded a
+	// non-nil LastCurrentTime for that build ID.
+	//
+	// The rollback strategy controls routing of NEW workflow executions only. Workflows already
+	// running are pinned to the version they started on and continue executing there; they are
+	// not affected by the rollback. Only new workflow executions will be routed to the rollback
+	// target version.
```
jaypipes:
```suggestion
	// not affected by the rollback. Only new workflow executions that are not pinned
	// to a specific build ID will be routed to the rollback target version.
```

api/v1alpha1/workerdeployment_types.go
```diff
@@ -83,6 +83,22 @@ type WorkerDeploymentSpec struct {
...
+	//
+	// Rollback is suppressed when the rollout strategy is Manual.
```
jaypipes: I'm just curious, but why is rollback suppressed when the rollout stategy is Manual?

api/v1alpha1/workerdeployment_types.go
```diff
@@ -405,6 +433,30 @@ type RolloutStrategy struct {
...
+type RollbackStrategy struct {
+	// Strategy for rollback. Valid values are "AllAtOnce" or "Progressive".
+	// Defaults to "AllAtOnce" for fast recovery.
+	Strategy VersionRollbackStrategy `json:"strategy"`
+
+	// Steps to execute progressive rollbacks. Only required when strategy is "Progressive".
+	// +optional
+	Steps []RolloutStep `json:"steps,omitempty"`
+
+	// MaxVersionAge limits which versions are eligible as rollback targets.
```
jaypipes: I have a suspicion this is overly complicated configuration for most users. Would a simpler "just roll back to the previous working (i.e. passes gates) version" behaviour be easier for users to grok? Where the controller simply would attempt rolling back to N-1, then N-2, etc... until landing on a working version?

api/v1alpha1/workerdeployment_types.go
```diff
@@ -405,6 +433,30 @@ type RolloutStrategy struct {
...
+// This is separate from RolloutStrategy because rollbacks have different requirements:
+// - No gate workflow (already trusted version)
+// - No manual mode (rollbacks should be automatic)
+// - Default to AllAtOnce for fast recovery
+// The rollback strategy is ignored if the rollout strategy is set to Manual and
+// rollback is disabled if the MaxVersionAge is set to 0.
+type RollbackStrategy struct {
+	// Strategy for rollback. Valid values are "AllAtOnce" or "Progressive".
+	// Defaults to "AllAtOnce" for fast recovery.
+	Strategy VersionRollbackStrategy `json:"strategy"`
```
jaypipes: Is there really a valid reason or use case for having a rollback strategy not match the rollout strategy? Why not just have the rollback strategy always be the same as rollout?

api/v1alpha1/workerdeployment_webhook.go
```diff
@@ -117,6 +137,67 @@ func validateRolloutStrategy(s RolloutStrategy) []*field.Error {
...
+			if rollbackTotal > rolloutTotal {
+				return admission.Warnings{fmt.Sprintf(
+					"rollback strategy is slower than rollout: progressive rollback total duration (%s) exceeds progressive rollout total duration (%s) — is that intended?",
+					rollbackTotal, rolloutTotal,
+				)}
+			}
+		}
+	}
+	return nil
+}
```
jaypipes: see my comment above... I suspect there's not really a valid use case for having a rollback strategy not match the rollout strategy... if we can just say the rollback == rollout, we can simplify things quite a bit and remove the need for these warnings.

jaypipes (commented):
@eniko-dif thanks very much for this! I think the UX can be simplified quite a bit by just making the rollback strategy match the rollout strategy. Also, I recommend removing the `MaxVersionAge` configuration option. Instead, simply roll back to the previous WDV's build ID. I really don't see a use case for configuring a rollback target version of anything other than the previous version...

docs/configuration.md
```diff
@@ -110,6 +113,14 @@ rollout:
       pauseDuration: 0s   # Full rollout after canary validation
 ```
 
+## Rollback Strategy
+
+The controller automatically detects rollbacks: if the target version was previously the current version within the last hour, it immediately routes 100% of traffic back to that version (AllAtOnce), regardless of the configured rollout strategy. This means setting your target version back to a prior build ID is enough to trigger a rollback — no special action or configuration required.
```
jaypipes, line 118: 👍

docs/configuration.md
```diff
@@ -110,6 +113,14 @@ rollout:
...
+
+Rollback behavior is not configurable: it always uses the AllAtOnce strategy with a fixed 1-hour eligibility window, so there is no `rollback` field in the spec.
```
jaypipes, line 120: For completeness, it might be worth adding a callout in the [CD Rollout doc](https://github.com/temporalio/temporal-worker-controller/blob/main/docs/cd-rollouts.md#triggering-a-rollout) that notes that setting the `image` field to an image tag related to a past active version in the last hour effectively issues a rollback.

internal/planner/planner_test.go
```diff
@@ -3686,3 +3686,304 @@ func TestGetWRTOwnerRefPatches(t *testing.T) {
...
+
+			if result == nil {
+				t.Fatal("expected non-nil VersionConfig")
+			}
+
+			assert.Equal(t, tc.expectSetCurrent, result.SetCurrent, tc.description)
+			assert.Equal(t, tc.status.VersionConflictToken, result.ConflictToken)
+		})
+	}
+}
```
jaypipes, line 3989: solid unit tests, thank you @eniko-dif :)

internal/planner/planner.go
```diff
@@ -905,6 +909,52 @@ func getVersionConfigDiff(
...
+	// The target version was last current more than an hour ago, making it too old to trust immediate rollback
+	if time.Since(*targetVersionInfo.LastCurrentTime) > defaults.RollbackMaxVersionAge {
+		l.Info("Skipping rollback: the version's last current time exceeds the max rollback version age",
+			"targetBuildID", status.TargetVersion.BuildID,
+			"lastCurrentTime", targetVersionInfo.LastCurrentTime,
+			"maxVersionAge", defaults.RollbackMaxVersionAge)
+		return false
+	}
+
+	// The target version was current in the last 1h, so rollback immediatley
```
```suggestion
	// The target version was current in the last 1h, so rollback immediately
```

internal/tests/internal/deployment_controller.go
```diff
@@ -56,11 +59,10 @@ func startAndStopWorker(t *testing.T, ctx context.Context, k8sClient client.Clie
 	// wait for worker to start
 	<-startedCh
 
-	time.Sleep(1 * time.Second)
```
jaypipes, line 59: Will removing this sleep cause flakiness? It's not clear why this sleep was originally added but I have to assume it was there for a reason :)

internal/tests/internal/integration_test.go
```diff
@@ -1024,6 +1063,17 @@ func testWorkerDeploymentCreation(
 		t.Fatalf("failed to create WorkerDeployment: %v", err)
 	}
 
+	// Immediately apply the input status before the controller's first reconcile.
```
jaypipes: How do you know that the controller's first reconcile hasn't already started? Line 1062 above (`k8sClient.Create()` will end up sending the new WD record out to the controller, right? So, the controller may have already started processing that loop before line 1072 executes below...

jaypipes (commented):
Code looks great, thank you @eniko-dif! I have two questions for you to address, both in the integration test additions. Otherwise, this looks good to get in for [vNext](https://github.com/temporalio/temporal-worker-controller/milestone/8) (which would make the next release version v1.9.0)

internal/tests/internal/deployment_controller.go
eniko-dif: It's not needed anymore, because the function changed. Now instead of starting the worker for one second, it starts it and returns the stop function. This way the worker is alive exactly for the time the test needs to register its version. So instead of being more flaky, it should be not flaky at all anymore.

internal/tests/internal/integration_test.go
eniko-dif: I updated the comment to match the behavior and expectations. :)

internal/tests/internal/integration_test.go
```diff
@@ -1024,6 +1063,16 @@ func testWorkerDeploymentCreation(
 		t.Fatalf("failed to create WorkerDeployment: %v", err)
 	}
 
+	// k8sClient.Create strips the status subresource, so the TWD starts with an empty
+	// status. Not guaranteed to precede the controller's first reconcile, but seeding it
+	// here helps test cases with pre-existing target/deprecated versions converge faster
+	// and avoid flaking against the eventually timeouts below.
```
jaypipes, line 1069: 👍

jaypipes (approved):
rock on @eniko-dif :)

internal/planner/planner.go
```diff
@@ -905,6 +909,52 @@ func getVersionConfigDiff(
...
 }
 
+func isRollbackScenario(
+	l logr.Logger,
+	status *temporaliov1alpha1.WorkerDeploymentStatus,
+	temporalState *temporal.TemporalWorkerState,
+	config *Config,
+) bool {
+	// Do not rollback when the user takes control of deployments with manual mode
+	if config.RolloutStrategy.Strategy == temporaliov1alpha1.UpdateManual {
```
jlegrone, line 919: Is a nil check needed here? Ie. if someone is running the controller without the defaulting webhook enabled, could config.RolloutStrategy be nil?

jlegrone (approved):

internal/planner/planner.go
eniko-dif: No, because  `config.RolloutStrategy` is a value type and it is a required field in the worker spec.

PR #276, opened 2026-04-14
fix: accept Opaque secrets for mTLS auth

### LINKED ISSUE #275: [Bug] ParseClientSecret rejects Opaque secrets for AuthModeTLS
## Summary

`ParseClientSecret` enforces `corev1.SecretTypeTLS` for mTLS connections, but many organizations
store TLS credentials in `Opaque` secrets to support multiple files in a single secret (e.g.,
`tls.crt`, `tls.key`, and `ca.crt` together). The Kubernetes `kubernetes.io/tls` secret type only
natively supports two keys (`tls.crt` and `tls.key`), so teams that need to bundle a CA cert
alongside the keypair must use `Opaque`.

## Steps to Reproduce

1. Create a Kubernetes `Opaque` secret containing `tls.crt`, `tls.key`, and optionally `ca.crt`
2. Configure a `TemporalWorker` with `AuthModeTLS` referencing that secret
3. Observe: reconciler fails with `secret <name> must be of type kubernetes.io/tls`

## Expected Behavior

The controller should accept both `corev1.SecretTypeTLS` and `corev1.SecretTypeOpaque` for
`AuthModeTLS`, as both can contain the required `tls.crt` / `tls.key` data. The actual credential
parsing in `fetchClientUsingMTLSSecret` already reads the keys by name (`secret.Data["tls.crt"]`,
`secret.Data["tls.key"]`), so no further changes are needed there — only the type guard in
`ParseClientSecret` needs relaxing.

## Affected Code

**File:** `internal/controller/clientpool/clientpool.go:252`

```go
case AuthModeTLS:
    if secret.Type != corev1.SecretTypeTLS {
        err := fmt.Errorf("secret %s must be of type kubernetes.io/tls", secret.Name)
        return nil, nil, nil, err
    }
    return cp.fetchClientUsingMTLSSecret(secret, opts)
```

## Suggested Fix

```go
case AuthModeTLS:
    if secret.Type != corev1.SecretTypeTLS && secret.Type != corev1.SecretTypeOpaque {
        err := fmt.Errorf("secret %s must be of type kubernetes.io/tls or Opaque", secret.Name)
        return nil, nil, nil, err
    }
    return cp.fetchClientUsingMTLSSecret(secret, opts)
```

## Additional Context

- The `fetchClientUsingMTLSSecret` function is already key-name-driven and works correctly with
  `Opaque` secrets containing the right keys — only the type assertion blocks it.
- Standard Kubernetes tooling (e.g., cert-manager) can issue either secret type depending on
  configuration.
- This restriction breaks organizations that bundle `tls.crt`, `tls.key`, and `ca.crt` into a
  single `Opaque` secret, which is a common pattern for managing multiple file-based credentials.

ConnorGraham:
## What was changed

Relax the secret type check in `ParseClientSecret` to accept both `kubernetes.io/tls` and `Opaque` secret types for `AuthModeTLS`. Update the `MutualTLSSecretRef` field comment in the CRD types to reflect the expanded accepted types.

## Why?

Previously, the controller enforced `type: kubernetes.io/tls` for mTLS secrets. This blocked organizations that store TLS credentials in `Opaque` secrets — a common pattern when bundling `tls.crt`, `tls.key`, and `ca.crt` into a single secret (e.g. multi-file cert-manager outputs, or any tooling that produces a keypair alongside a custom CA).

The `kubernetes.io/tls` secret type natively supports only two keys (`tls.crt` and `tls.key`), so teams needing to include a CA cert alongside the keypair must use `Opaque`. The downstream handler `fetchClientUsingMTLSSecret` already accesses secret data by key name and works correctly with either type — only the type guard was blocking it.

## Checklist

1. Closes #275

2. How was this tested:

Unit test `TestParseClientSecret_OpaqueSecretType` (renamed from `TestParseClientSecret_WrongSecretType`) now asserts that an `Opaque` secret containing `tls.crt` and `tls.key` is accepted and produces a valid `ClientAuth` with `AuthModeTLS`. All 316 existing tests continue to pass.

3. Any docs updates needed?

`MutualTLSSecretRef` field comment in `api/v1alpha1/temporalconnection_types.go` updated to document both accepted secret types.

### DIFF
--- a/api/v1alpha1/temporalconnection_types.go
+++ b/api/v1alpha1/temporalconnection_types.go
@@ -27,8 +27,10 @@ type TemporalConnectionSpec struct {
 	HostPort string `json:"hostPort"`
 
 	// MutualTLSSecretRef is the name of the Secret that contains the TLS certificate and key
-	// for mutual TLS authentication. The secret must be `type: kubernetes.io/tls` and exist
-	// in the same Kubernetes namespace as the TemporalConnection resource.
+	// for mutual TLS authentication. The secret must be `type: kubernetes.io/tls` or
+	// `type: Opaque` and exist in the same Kubernetes namespace as the TemporalConnection
+	// resource. Opaque secrets are useful when bundling tls.crt, tls.key, and ca.crt into
+	// a single secret (e.g. multi-file cert-manager outputs).
 	//
 	// More information about creating a TLS secret:
 	// https://kubernetes.io/docs/concepts/configuration/secret/#tls-secrets
--- a/internal/controller/clientpool/clientpool.go
+++ b/internal/controller/clientpool/clientpool.go
@@ -249,8 +249,8 @@ func (cp *ClientPool) ParseClientSecret(
 	// Check the secret type
 	switch authMode {
 	case AuthModeTLS:
-		if secret.Type != corev1.SecretTypeTLS {
-			err := fmt.Errorf("secret %s must be of type kubernetes.io/tls", secret.Name)
+		if secret.Type != corev1.SecretTypeTLS && secret.Type != corev1.SecretTypeOpaque {
+			err := fmt.Errorf("secret %s must be of type kubernetes.io/tls or Opaque", secret.Name)
 			return nil, nil, nil, err
 		}
 		return cp.fetchClientUsingMTLSSecret(secret, opts)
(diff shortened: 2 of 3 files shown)


carlydf (approved):
thanks!

PR #288, opened 2026-04-22
doc: release policy and versioning

### LINKED ISSUE #264: Document release policy
Explain what Semantic Versioning ✨means to us✨, when we will do Github releases vs Helm Chart releases, which version boundaries we do upgrade-downgrade tests on (hint: only minor version to minor version, so in production people should upgrade one minor version at a time), and anything else the people should know about our releases.

jaypipes:
Adds documentation that explains how we think about releases of the application and Helm Charts and when we will bump major, minor and patch versions for both Application and Chart Releases.

Closes #264

### DIFF
--- a/docs/release.md
+++ b/docs/release.md
@@ -62,3 +62,8 @@ Kubernetes resources in the `templates/` directory.
 The patch version of the Semantic Version string for the Chart Release is
 bumped when we release a new Application Release and update the `Chart.yaml`'s
 `appVersion` field to point to a new Application Release.
+
+We will *always* cut a new Chart Release patch version for each new Application
+Release. However, the Chart Release patch version may not be cut at the exact
+same *time* as the Application Release. In this way, we can test the Helm Chart
+release process in isolation from the Application release process.
--- a/docs/release.md
+++ b/docs/release.md
@@ -1,0 +1,64 @@
+# Releases
+
+This document explains our thinking around Releases of Temporal Worker
+Controller and associated Helm Charts.
+
+A Release is a coordination point for publishing a versioned set of artifacts.
+Each Release has a Github Release page which describes the changes in the
+Release and the artifacts associated with it.
+
+A Release Tag is simply the string name given to the Release.
+
+In the Temporal Worker Controller project, we cut two different kinds of Releases:
+
+* **Application Release**: contains the Temporal Worker Controller container image.
+* **Chart Release**: contains the Helm Charts that install the Temporal Worker
+  Controller and the Custom Resource Definitions (CRDs) used by the controller.
+
+These different kinds of Releases are on different cadences and use different
+version numbering.
+
+## Application Releases
+
+Application Releases refer to the Temporal Worker Controller code and the
+`temporalio/temporal-worker-controller` container image artifacts.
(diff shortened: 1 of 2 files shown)


docs/release.md
```diff
@@ -0,0 +1,64 @@
...
+`temporalio/temporal-worker-controller` container image artifacts.
+
+Release Tags for official Application Releases will always use Semantic Version
+strings, e.g. `v1.42.1`.
+
+We bump the minor version number of the Application Release version when new
+features are added **to the controller code itself**.
+
+Likewise, We bump the patch version number of the Application Release version
+when bug fixes are added to the controller code itself.
```
carlydf, line 33: if we release a bug fix to the controller code itself, will we also cut a patch of the corresponding Helm Release so that users can install the bug fix via Helm?

This section made me concerned that maybe we would patch the image and not package that bug fix into a helm chart.

To take a step back, my understanding is that because many users install this code via Helm, whenever we release new controller _code_, we will always release it in some Helm Chart Version. And also if we change the chart but not the controller code, we might release a new Helm Chart Version. Are there cases where we would release new code but not put it in any Helm Chart release?

carlydf: overall looks good though! just want to make sure I understand this and we are communicating clearly

jaypipes: > if we release a bug fix to the controller code itself, will we also cut a patch of the corresponding Helm Release so that users can install the bug fix via Helm?

Eventually, yes :)

> This section made me concerned that maybe we would patch the image and not package that bug fix into a helm chart.

We will always *eventually* cut a Helm Chart release that includes the patched application code. I deliberately avoided saying that we would cut a *coordinated* release of the application and Helm Charts, though :)

> To take a step back, my understanding is that because many users install this code via Helm, whenever we release new controller _code_, we will always release it in some Helm Chart Version. 

We will always release it in some Helm Chart version, yes. Just not guaranteed to be done at the same time.

> And also if we change the chart but not the controller code, we might release a new Helm Chart Version. 

Yes, absolutely.

> Are there cases where we would release new code but not put it in any Helm Chart release?

No, we will always *eventually* cut a Helm Chart version that includes that patched Application Release.

I'll add some wording saying essentially we will always cut a Helm Chart version that includes new Application Releases but won't commit to doing coordinated releasing of Application and Chart releases.

jaypipes: @carlydf I added a note, please let me know if that wording makes sense.

carlydf (approved):

docs/release.md
```diff
@@ -0,0 +1,69 @@
...
+> version or the APIVersion of the CRDs supported by the controller**. 
+
+## Chart Releases
+
+Chart Releases refer to the two Helm Charts that install the Temporal Worker
+Controller and manage the Custom Resource Definitions (CRDs) used by the
+controller.
+
+Release Tags for official Chart Releases are prefixed with the string `helm-`
+and the Semantic Version string for the Chart Release, e.g. `helm-v0.24.0`.
```
Shivs11, line 46: oooo nice

Shivs11 (approved):
tysm for doing this

PR #291, opened 2026-04-23
Back off on `DescribeWorkerDeployment` `ResourceExhausted` error

### LINKED ISSUE #278: [Bug] Worker Deployment Read API rate limit still exceeded in v1.4.0
## Description

v1.4.0 release notes mention:
> "Omit `DescribeVersion` calls for drained versions to avoid hitting RPS limits"

However, I am still hitting Worker Deployment Read API rate limit errors in v1.4.0.
The error is coming from `DescribeWorkerDeployment` (the top-level deployment describe),
not from `DescribeVersion` on individual versions.

## Error

```
"error":"unable to get Temporal worker deployment state: unable to describe worker
deployment X/y: Worker Deployment Read API rate limit exceeded for namespace "ABCD""
```

The stacktrace points to the standard reconciler loop in controller-runtime,

## Root Cause (suspected)

The v1.4.0 optimization skips `DescribeVersion` for drained versions, which reduces
per-version calls. However, `DescribeWorkerDeployment` is called unconditionally on
every reconcile iteration regardless of the deployment's state. With multiple
`TemporalWorkerDeployment` CRs in the same Temporal namespace and a short reconcile
interval, these calls aggregate and exceed the APS (actions-per-second) limit on
Temporal Cloud.

carlydf:
#### Problem

Namespaces with many TemporalWorkerDeployment objects can trigger Temporal's per-namespace DescribeWorkerDeployment rate limit (frontend.globalNamespaceWorkerDeploymentReadRPS, default 50 RPS). When this happens, the reconciler was returning the error immediately, causing the workqueue to requeue with an exponential backoff starting at ~5ms — effectively a tight retry loop that makes the rate-limit problem worse.

The error comes back as *serviceerror.ResourceExhausted (not a standard gRPC codes.ResourceExhausted status), so it must be detected with errors.As rather than grpcstatus.FromError.

#### Changes

  - internal/controller/worker_controller.go: detect *serviceerror.ResourceExhausted from GetWorkerDeploymentState and return RequeueAfter: 30s instead of an immediate error requeue. Sets ConditionProgressing=False with ReasonTemporalStateFetchFailed and a "Rate limited" message so the condition is visible to users.
  - internal/tests/internal/rate_limit_integration_test.go: new integration test that creates 10 TWDs against a 1 RPS limit, confirming the error surfaces with the expected condition reason and message.
  - go.mod / go.sum / go.work: bump Temporal server dependency to v1.31.0-154.2, which is the first version that enforces globalNamespaceWorkerDeploymentReadRPS.

## Checklist

1. Closes #278 

2. How was this tested:
  - KUBEBUILDER_ASSETS=.../bin/k8s/1.27.1-darwin-arm64 go test -tags test_dep ./internal/tests/internal -run "TestIntegration/rate-limit" -timeout 120s -v passes
  - Same test fails when the errors.As block is removed (reverts to immediate retry with generic message)
  - Full integration suite passes: go test -tags test_dep ./internal/tests/internal -run TestIntegration -timeout 600s

3. Any docs updates needed?


carlydf:
integration test will pass once I pull in #290

Shivs11 (approved):

internal/tests/internal/rate_limit_integration_test.go
```diff
@@ -0,0 +1,95 @@
...
+						strings.Contains(c.Message, "Rate limited") {
+						t.Logf("TWD %s confirmed rate-limited: %s", twd.Name, c.Message)
+						return nil
+					}
+				}
+			}
+			return errors.New("no TWD has been rate-limited yet")
+		})
+	})
+}
```
jaypipes, line 95: Nice test and nice comments explaining the test assertions and setup. ++

jaypipes (approved):
muy bueno.

PR #292, opened 2026-04-23
Fix events RBAC and automate Helm ClusterRole generation from markers

### LINKED ISSUE #277: [Bug] ClusterRole missing events patch/create permissions for cross-namespace TemporalWorkerDeployment resources (v1.4.0)
## Description

Starting in v1.4.0, the controller attempts to patch Kubernetes `events` for
`TemporalWorkerDeployment` objects. However, the `ClusterRole` provisioned by
the Helm chart does not include `create`/`patch` verbs on the `events` resource,
causing the controller to continuously log permission errors and fail to record
events.

## Setup

- **Controller chart version:** 1.4.0
- **Controller namespace:** `X`
- **TemporalWorkerDeployment namespace:** `Y`

The controller is deployed in a different namespace than the `TemporalWorkerDeployment`
CRs it manages. This is a common and supported pattern (deploying the controller
once cluster-wide and managing CRs in application namespaces).

## Error

```
Server rejected event (will not retry!)" err="events \"dbtworker.18a6745739983ba1\" is forbidden: User \"system:serviceaccount:X:temporal-worker-controller-service-account\" cannot patch resource \"events\" in │
│  API group \"\" in the namespace \"Y\""
```

## Root Cause

The `ClusterRole` created by the Helm chart is missing `events` permissions.
The controller-runtime library used by the controller (`sigs.k8s.io/controller-runtime@v0.21.0`)
records events against the namespace where the watched resource lives (in this case `X`),
not where the controller itself is deployed.

Since a `ClusterRole` + `ClusterRoleBinding` is used (as expected for a cluster-wide
controller), the fix is straightforward — add `events` to the ClusterRole rules.

## Fix

Add to the controller's `ClusterRole`:

```yaml
- apiGroups:
  - ""
  resources:
  - events
  verbs:
  - create
  - patch
```

## Workaround

Manually patching the `ClusterRole` to add the above rule resolves the issue
until a chart fix is released.

carlydf:
### Two related issues:

1. Events permission used the wrong API group.
  The +kubebuilder:rbac marker for events specified groups=events.k8s.io (the newer structured events API), but controller-runtime records v1.Event objects against the core "" API group. This caused the controller to log Server rejected event (will not retry!) errors when deployed in a different namespace from the TemporalWorkerDeployment CRs it manages — common cluster-wide deployment pattern.

2. The Helm ClusterRole was hand-maintained and had drifted from the Go markers.
The +kubebuilder:rbac markers in worker_controller.go were incomplete: workerresourcetemplates, workerresourcetemplates/status, and subjectaccessreviews were all present in the Helm chart but had no corresponding markers. This made it easy for RBAC rules to fall out of sync — which is exactly how the events bug happened in the first place.

## Changes

Fix the events API group (worker_controller.go, config/rbac/role.yaml): correct groups=events.k8s.io → groups=core in the marker; the generated manifest now uses apiGroups: [""].

Add missing markers (worker_controller.go): add markers for workerresourcetemplates (get/list/watch/patch/update), workerresourcetemplates/status (get/patch/update), and authorization.k8s.io/subjectaccessreviews (create) to match what was already deployed by the Helm chart.

Automate Helm ClusterRole sync (hack/sync-rbac-rules.py, Makefile, helm/.../rbac.yaml): make manifests now runs a script that reads the controller-gen-generated config/rbac/role.yaml and replaces the # GENERATED RULES BEGIN / # GENERATED RULES END section in the Helm template. The Helm-templated dynamic rules (the allowedResources range) are left untouched. Going forward, adding a +kubebuilder:rbac marker and running make manifests is all that's needed to update the deployed ClusterRole.

Closes #277

### DIFF
--- a/hack/sync-rbac-rules.py
+++ b/hack/sync-rbac-rules.py
@@ -1,0 +1,65 @@
+#!/usr/bin/env python3
+"""Sync rules from config/rbac/role.yaml into the Helm ClusterRole template.
+
+The Helm template must contain:
+  # GENERATED RULES BEGIN
+  ...
+  # GENERATED RULES END
+markers inside its ClusterRole rules section.  Everything between the markers
+is replaced with the rules from config/rbac/role.yaml (indented by two spaces).
+"""
+import re
+import sys
+
+ROLE_YAML = "config/rbac/role.yaml"
+HELM_RBAC = "helm/temporal-worker-controller/templates/rbac.yaml"
+BEGIN_MARKER = "  # GENERATED RULES BEGIN"
+END_MARKER = "  # GENERATED RULES END"
+
+
+def extract_rules_text(path):
+    with open(path) as f:
+        content = f.read()
+    idx = content.find("\nrules:\n")
+    if idx == -1:
+        print(f"ERROR: 'rules:' not found in {path}", file=sys.stderr)
+        sys.exit(1)
+    rules_body = content[idx + len("\nrules:\n"):]
+    # Indent lines relative to the `rules:` key in the Helm template.
+    # controller-gen emits two indentation levels:
+    #   col 0: outer list items  (e.g. "- apiGroups:")   → add 2 spaces
+    #   col 2: inner list values (e.g. "  - events")     → add 4 spaces
+    #   col 2: mapping keys      (e.g. "  resources:")   → add 2 spaces
+    # The extra indent on inner list values matches the style used by the
+    # hand-authored rules in the Helm template.
+    lines = rules_body.splitlines(keepends=True)
+    result = []
+    for line in lines:
+        if not line.strip():
+            result.append(line)
+        elif line.startswith("  - "):
+            result.append("    " + line)  # inner list value: 2 global + 2 extra
+        else:
+            result.append("  " + line)    # outer list item or mapping key: 2 global
+    return "".join(result)
+
+
+def update_helm(path, rules_text):
+    with open(path) as f:
+        content = f.read()
+    pattern = re.compile(
+        r"(" + re.escape(BEGIN_MARKER) + r"[^\n]*\n)(.*?)(" + re.escape(END_MARKER) + r")",
+        re.DOTALL,
+    )
+    if not pattern.search(content):
+        print(f"ERROR: markers not found in {path}", file=sys.stderr)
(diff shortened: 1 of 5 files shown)


internal/controller/worker_controller.go
```diff
@@ -102,7 +102,10 @@ type TemporalWorkerDeploymentReconciler struct {
 //+kubebuilder:rbac:groups=core,resources=secrets,verbs=get;list;watch
 //+kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=apps,resources=deployments/scale,verbs=update
-// +kubebuilder:rbac:groups=events.k8s.io,resources=events,verbs=create;patch
+//+kubebuilder:rbac:groups=core,resources=events,verbs=create;patch
+//+kubebuilder:rbac:groups=temporal.io,resources=workerresourcetemplates,verbs=get;list;watch;patch;update
+//+kubebuilder:rbac:groups=temporal.io,resources=workerresourcetemplates/status,verbs=get;patch;update
+//+kubebuilder:rbac:groups=authorization.k8s.io,resources=subjectaccessreviews,verbs=create
```
carlydf: the latter 3 of these are just adding what was previously hand-added to the rbac.yaml in helm

hack/sync-rbac-rules.py
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env python3
```
carlydf, line 1: could fix this issue simply by updating rbac.yaml by hand if we don't want this script, but the surface is low and I think this method is better so we don't forget to add things in future

jaypipes (commented):
lgtm.

Might I suggest adding a check to the `.github/workflows/helm-validate.yml` workflow file that double-checks nobody manually updated the Helm RBAC stuff?

```yml
jobs:
  helm-check-rbac:
    name: Check Helm RBAC wasn't manually updated
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - run: make manifests && git diff --exit-code helm/temporal-worker-controller/templates/rbac.yaml
```

jaypipes (approved):
excellent work on this, thank you @carlydf! :)

PR #293, opened 2026-04-23
Validate TWD spec via CRD CEL rules at apply time (fixes #62)

### LINKED ISSUE #62: Validate `TemporalWorkerDeployment` spec at apply time instead of reconcile time
validate max name length <= 63

carlydf:
## Summary

Fixes #62 — validates `TemporalWorkerDeployment` spec at apply time instead of reconcile time.

### Approach

Embed CEL validation rules directly in the CRD schema (`x-kubernetes-validations`) so the API server enforces them universally at admission time — no webhook infrastructure required. The optional TWD webhook remains to provide apply-time feedback to those who enable it.

### CRD CEL rules added (`worker_types.go` → regenerated CRD manifest)

| Rule | Placed on | Enforced by |
|---|---|---|
| `metadata.name` ≤ 63 chars | `TemporalWorkerDeployment` root | CRD |
| Progressive strategy requires steps | `RolloutStrategy` | CRD |
| `rampPercentage` steps capped at 20 | `steps` field (`maxItems`) | CRD |
| `pauseDuration` ≥ 30s per step | `RolloutStep` | CRD |
| `gate.inputFrom`: exactly one of `configMapKeyRef`/`secretKeyRef` | `RolloutStrategy` | CRD |

Two constraints from the original webhook could **not** be expressed as CRD CEL rules in k8s 1.27 (cost budget limit for list traversal; `gate.input` is `x-kubernetes-preserve-unknown-fields` and invisible to CEL):
- `rampPercentage` strictly increasing across steps
- `gate.input` and `gate.inputFrom` mutually exclusive

### Fallback for webhook-only constraints

The reconciler calls `ValidateCreate()` as a fallback for the two constraints the CRD cannot enforce. On failure it emits a `Warning` event and sets `ConditionProgressing=False` / `ConditionReady=False` with reason `InvalidSpec`, then returns without requeueing (the watch re-triggers when the user corrects the spec). Adds `ReasonInvalidSpec` constant.

### Webhook cleanup

`validateRolloutStrategy` in the webhook now only checks the two constraints not in the CRD. All checks that duplicated CRD rules were removed.

### Testing

- **Unit tests**: updated to match narrowed webhook scope (removed cases now enforced by CRD)
- **Reconciler unit test**: `TestReconcile_InvalidSpec_EmitsEventAndSetsCondition` verifies the event/condition path
- **envtest integration tests** (`temporalworkerdeployment_cel_validation_test.go`): hit a real kube-apiserver to verify each CRD CEL rule is syntactically valid and semantically correct, independently of the webhook

## Test plan
- [ ] `KUBEBUILDER_ASSETS=... go test ./...` passes locally
- [ ] Apply a TWD with name > 63 chars — API server rejects immediately
- [ ] Apply a TWD with `strategy: Progressive` and no steps — API server rejects immediately
- [ ] Apply a TWD with `strategy: Progressive` and > 20 steps — API server rejects immediately
- [ ] Apply a TWD with a step `pauseDuration < 30s` — API server rejects immediately
- [ ] Apply a TWD with decreasing `rampPercentage` (webhook disabled) — reconciler sets `InvalidSpec` condition and emits Warning event, does not act on the spec

### DIFF
--- a/api/v1alpha1/temporalworker_webhook.go
+++ b/api/v1alpha1/temporalworker_webhook.go
@@ -7,7 +7,6 @@ package v1alpha1
 import (
 	"context"
 	"fmt"
-	"time"
 
 	"github.com/temporalio/temporal-worker-controller/internal/defaults"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -19,9 +18,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 )
 
-const (
-	maxTemporalWorkerDeploymentNameLen = 63
-)
 
 func (r *TemporalWorkerDeployment) SetupWebhookWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewWebhookManagedBy(mgr).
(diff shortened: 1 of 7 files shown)


api/v1alpha1/worker_types.go
```diff
@@ -379,6 +379,10 @@ type GateInputSource struct {
 }
 
 // RolloutStrategy defines strategy to apply during next rollout
+// +kubebuilder:validation:XValidation:rule="self.strategy != 'Progressive' || (has(self.steps) && size(self.steps) > 0)",message="steps are required for Progressive rollout"
+// +kubebuilder:validation:XValidation:rule="self.strategy != 'Progressive' || !has(self.steps) || self.steps.size() <= 1 || (self.steps.map(s, s.rampPercentage).isSorted() && self.steps.map(s, s.rampPercentage).all(v, self.steps.filter(s, s.rampPercentage == v).size() == 1))",message="rampPercentage must increase between each step"
+// +kubebuilder:validation:XValidation:rule="!has(self.gate) || !(has(self.gate.input) && has(self.gate.inputFrom))",message="only one of input or inputFrom may be set"
+// +kubebuilder:validation:XValidation:rule="!has(self.gate) || !has(self.gate.inputFrom) || (has(self.gate.inputFrom.configMapKeyRef) != has(self.gate.inputFrom.secretKeyRef))",message="exactly one of configMapKeyRef or secretKeyRef must be set"
```
jaypipes: would it make more sense to put these validation expressions on GateWorkflowConfig itself instead of RolloutStrategy?

carlydf: I actually ended up removing these and validating them elsewhere because:
> Two issues with the initial CEL rules surfaced in k8s 1.27 envtest:
>
> 1. gate.input is x-kubernetes-preserve-unknown-fields: true — CEL cannot reference it structurally, so has(self.gate.input) failed to compile. Removed the gate input/inputFrom mutual-exclusivity rule; the webhook Go code still enforces it when enabled.
>
> 2. Any list traversal on self.steps at the RolloutStrategy level (map, isSorted) hits the schema-level CEL cost budget in k8s 1.27 because the cost estimator does not propagate maxItems from the steps array schema up to the parent object's rule. Removed the rampPercentage ordering rule from the CRD; the webhook Go code still enforces it when enabled.
>
> Remaining CRD rules (all O(1)): name length, Progressive requires steps, pauseDuration >= 30s per step, gate.inputFrom exclusivity.

api/v1alpha1/temporalworker_webhook.go
```diff
@@ -19,9 +18,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 )
 
-const (
```
carlydf, line 22: not needed anymore since this is handled through CRD CEL rules

api/v1alpha1/temporalworker_webhook_test.go
```diff
@@ -27,10 +27,6 @@ func TestTemporalWorkerDeployment_ValidateCreate(t *testing.T) {
 		"valid temporal worker deployment": {
 			obj: testhelpers.MakeTWDWithName("valid-worker", ""),
 		},
-		"temporal worker deployment with name too long": {
```
carlydf, line 30: covered by api/v1alpha1/temporalworkerdeployment_cel_validation_test.go (same with other removed webhook tests)

jaypipes (approved):
Very nice, @carlydf. Thank you for putting the detailed comments in here to explain why some checks could only be done in the reconciler loop as opposed to in the API server's CEL check stage.

PR #294, opened 2026-04-23
CRD rename: TemporalWorkerDeployment → WorkerDeployment, TemporalConnection → Connection

### LINKED ISSUE #256: remove `Temporal` prefix on CRD names
We'd like to remove the `Temporal` prefix from the `TemporalWorkerDeployment` and `TemporalConnection` custom resource definition names.

Tentative plan is to introduce identical CRDs but with that prefix removed and a small piece of code in the controller that auto-converts older named CRs to newer named CRs.

carlydf:
## What was changed

Renames the two primary CRDs and a `WorkerResourceTemplate` field:

| Old | New |
|---|---|
| `TemporalWorkerDeployment` | `WorkerDeployment` |
| `TemporalConnection` | `Connection` |
| `WorkerResourceTemplate.spec.temporalWorkerDeploymentRef` | `WorkerResourceTemplate.spec.workerDeploymentRef` |

In v1.7, the old CRD kinds are not actively managed: existing objects are not reconciled against Temporal server state, and new objects of these kinds cannot be created. The deprecated CRDs exist only to support zero-downtime migration of resources already on the cluster.

### Key implementation details

- **Migration helper reconcilers** (`DeprecatedTWDReconciler`, `DeprecatedTCReconciler`) set a `Ready=False` status condition on deprecated resources, progressing through `Deprecated` → `WorkerDeploymentExists` (ownership transfer in progress) → `MigratedToWorkerDeployment` (safe to delete).
- **Ownership transfer**: when the v1.7 controller first sees a `WorkerDeployment` with the same name as an existing `TemporalWorkerDeployment`, it patches the ownerRefs on all child Deployments and WorkerResourceTemplates to point to the new owner — no pods are restarted.
- **Migration-guard finalizer** (`temporal.io/migration-guard`): added to every `TemporalWorkerDeployment` and `TemporalConnection` on controller startup. Prevents deletion until migration is confirmed — TWD requires the `migrated-to-wd: "true"` label, TC requires a matching `Connection` to exist. Resources stuck in `Terminating` surface a `DeletingPendingMigration` condition with an actionable message.
- **Deprecated CRD manifests** are generated by `make manifests` alongside the new CRDs. CEL rule `oldSelf != null` blocks new creates; `deprecated: true` + `deprecationWarning` surfaces a warning on every API call.
- **Deprecated Go types** (`deprecated_temporalworkerdeployment_types.go`, `deprecated_temporalconnection_types.go`) retain the full v1.6 typed spec with renamed sub-structs (e.g. `DeprecatedWorkerOptions`) to avoid collision with the new types, keeping `json:` tags identical so the schema is unchanged.
- **RBAC changes are additive only**: all existing permissions kept, new permissions added only for the two new CRDs.

## Migration flow

1. **Upgrade the CRDs chart** — installs new CRDs, marks old kinds deprecated. Running v1.6 controller is unaffected.
2. **Upgrade the controller** — v1.7 controller adds `temporal.io/migration-guard` finalizers to all deprecated resources.
3. **Find-and-replace kind names** in your manifests (`TemporalWorkerDeployment` → `WorkerDeployment`, `TemporalConnection` → `Connection`) and apply via `helm upgrade` / `kubectl apply` / GitOps. This simultaneously creates the new resources and marks the old ones for deletion; the finalizer holds the deprecated resources in `Terminating` until the controller confirms ownership transfer, then releases them. Workers are continuously managed throughout.

No need to create new resources before upgrading the controller — the finalizer ensures deprecated resources cannot disappear before migration is complete regardless of order.

## Why?

The `Temporal` prefix was redundant given the `temporal.io` API group. We wanted to clean this up before GA.

## Checklist

1. Closes #256

2. Tested via existing integration test suite (dual-field tests cover both old and new CRD kinds)
TODO: test migration in real life

3. Docs: all existing docs updated to use new names; `docs/migration-crd-rename.md` added with migration steps


internal/controller/deprecated_reconciler.go
```diff
@@ -0,0 +1,103 @@
...
+// resources that share the same Temporal deployment name.
+//
+// +kubebuilder:rbac:groups=temporal.io,resources=temporalworkerdeployments,verbs=get;list;watch;update;patch
+// +kubebuilder:rbac:groups=temporal.io,resources=temporalworkerdeployments/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=temporal.io,resources=temporalworkerdeployments/finalizers,verbs=update
+type DeprecatedTWDReconciler struct {
+	client.Client
+}
+
+func (r *DeprecatedTWDReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
carlydf, line 43: note: this file is actually interesting to review

internal/controller/reconciler_events_test.go
```diff
@@ -349,8 +349,8 @@ func TestSyncConditions(t *testing.T) {
 		assertCondition(t, twd, temporaliov1alpha1.ConditionReady, metav1.ConditionTrue, temporaliov1alpha1.ReasonRolloutComplete)
 		assertCondition(t, twd, temporaliov1alpha1.ConditionProgressing, metav1.ConditionFalse, temporaliov1alpha1.ReasonRolloutComplete)
 		// Deprecated conditions
-		assertCondition(t, twd, temporaliov1alpha1.ConditionTemporalConnectionHealthy, metav1.ConditionTrue, temporaliov1alpha1.ReasonTemporalConnectionHealthy) //nolint:staticcheck // backward compat
-		assertCondition(t, twd, temporaliov1alpha1.ConditionRolloutComplete, metav1.ConditionTrue, temporaliov1alpha1.ReasonRolloutComplete)                     //nolint:staticcheck // backward compat
+		assertCondition(t, twd, temporaliov1alpha1.ConditionConnectionHealthy, metav1.ConditionTrue, temporaliov1alpha1.ReasonConnectionHealthy) //nolint:staticcheck // backward compat
```
carlydf, line 352: if we were only keeping these conditions for "backward compatibility" do we want to keep them now that we are moving to a new CRD?

internal/controller/worker_controller.go
```diff
@@ -383,59 +390,156 @@ func (r *TemporalWorkerDeploymentReconciler) Reconcile(ctx context.Context, req
 }
 
 // markWRTsTWDNotFound sets Ready=False/TWDNotFound on all WorkerResourceTemplates that reference
-// a TemporalWorkerDeployment that could not be found. This covers the case where the WRT is
+// a WorkerDeployment that could not be found. This covers the case where the WRT is
+// migrateFromDeprecatedTWD checks for a TemporalWorkerDeployment stub with the
+// same name/namespace as wd. If one exists and has not yet been migrated, it
+// transfers ownership of child Deployments and WorkerResourceTemplates from the
+// deprecated TWD to wd, then labels the TWD as migrated.
+func (r *WorkerDeploymentReconciler) migrateFromDeprecatedTWD(ctx context.Context, l logr.Logger, wd *temporaliov1alpha1.WorkerDeployment) error {
```
carlydf, line 409: note for reviewers: this is another actually interesting section

internal/tests/internal/integration_test.go
```diff
@@ -827,10 +827,19 @@ func TestIntegration(t *testing.T) {
 		},
 	}
 
+	// TODO: "6th-rollout-unblocked-after-pollers-die-max-ctrlr-versions" is flaky after the
```
carlydf: come back to this!

hack/sync-rbac-rules.py
```diff
@@ -1,28 +1,27 @@
 #!/usr/bin/env python3
-"""Sync rules from config/rbac/role.yaml into the Helm ClusterRole template.
```
carlydf: config is no longer in the gitignore, so I almost committed it. We could put it back in gitignore, but this solves the problem more fully.

helm/temporal-worker-controller/templates/rbac.yaml
```diff
@@ -188,7 +299,7 @@ metadata:
   name: {{ .Release.Name }}-{{ .Release.Namespace }}-temporalconnection-editor-role
 rules:
   - apiGroups:
-      - temporal.io.temporal.io
```
carlydf, line 197: mistake previously in main meant no one was using this anyway, but just fix it for symmetry with the other roles. will remove in v1.8

internal/controller/deprecated_tc_reconciler.go
```diff
@@ -0,0 +1,72 @@
...
+// DeprecatedTCReconciler reconciles TemporalConnection stubs.
+//
+// IMPORTANT: this reconciler must NEVER call Temporal APIs or manage k8s
+// resources. Its only job is to surface a status condition guiding users to
+// migrate to the Connection CRD.
+type DeprecatedTCReconciler struct {
+	client.Client
+}
+
+func (r *DeprecatedTCReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
carlydf, line 28: other actually interesting piece

internal/controller/deprecated_tc_reconciler.go
```diff
@@ -0,0 +1,72 @@
...
+		reason  string
+		message string
+	)
+
+	var conn temporaliov1alpha1.Connection
+	err := r.Get(ctx, req.NamespacedName, &conn)
+	switch {
+	case err == nil:
+		reason = "MigratedToConnection"
+		message = "Migration complete. Delete this TemporalConnection."
```
carlydf, line 87: as soon as a Connection exists with the same name, migration can be considered complete. No ownership transfers needed.

internal/controller/reconciler_events_test.go
carlydf: keeping them for now to make all these name changes as simple as possible to reason about

internal/controller/worker_controller.go
```diff
@@ -136,32 +143,40 @@ func (r *TemporalWorkerDeploymentReconciler) Reconcile(ctx context.Context, req
...
+		// the next reconcile fires naturally when the WD is created. If the List or
 		// status updates fail (transient API errors), return the error to requeue with backoff.
-		return ctrl.Result{}, r.markWRTsTWDNotFound(ctx, req.NamespacedName)
+		return ctrl.Result{}, r.markWRTsWDNotFound(ctx, req.NamespacedName)
+	}
+
+	// Migration: if a deprecated TemporalWorkerDeployment with the same name/namespace
+	// exists and has not yet been migrated, transfer ownership of its child Deployments
+	// and WorkerResourceTemplates to this WorkerDeployment.
+	if err := r.migrateFromDeprecatedTWD(ctx, l, &workerDeploy); err != nil {
```
carlydf, line 172: interesting to review

api/v1alpha1/deprecated_temporalconnection_types.go
```diff
@@ -48,15 +40,19 @@ type TemporalConnectionSpec struct {
 
 // TemporalConnectionStatus defines the observed state of TemporalConnection
 type TemporalConnectionStatus struct {
-	// TODO(jlegrone): Add additional status fields following Kubernetes API conventions
-	// https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status
+	// +optional
+	// +listType=map
+	// +listMapKey=type
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
```
carlydf, line 46: so that we can mark Connection.Status.Conditions Ready=False

api/v1alpha1/deprecated_temporalconnection_types.go
```diff
@@ -48,15 +40,19 @@ type TemporalConnectionSpec struct {
...
 }
 
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
 //+kubebuilder:resource:shortName=tconn
 //+kubebuilder:printcolumn:name="Host",type="string",JSONPath=".spec.hostPort",description="Temporal server endpoint"
 //+kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp",description="Age"
+// +kubebuilder:validation:XValidation:rule="oldSelf != null",message="TemporalConnection is deprecated and cannot be created. Use Connection instead."
```
jaypipes, line 54: Nice.

docs/migration-crd-rename.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Why the rename?
+
+The `Temporal` prefix was redundant — all resources in the `temporal.io` API group are already scoped to Temporal. The shorter names are consistent with Kubernetes naming conventions and reduce verbosity in manifests and CLI commands.
+After this release, the Worker Controller will be Generally Available (GA), which means no more breaking changes will be introduced. We wanted to make this transition before GA, and make it as clean as possible.
+
+## Migration steps
+
+> **Dev / non-production environments:** If you don't need to preserve any worker state, the simplest path is to delete all your `TemporalWorkerDeployment` and `TemporalConnection` resources while the v1.6 controller is still running. At that point no migration-guard finalizer has been added yet, so deletion completes after the v1.6 finalizer completes. Note that all related Worker Deployment state in the Temporal server will also be deleted. Then upgrade the controller and create fresh `WorkerDeployment` and `Connection` resources.
```
jaypipes, line 20: 👍

jaypipes (approved):
Super thorough. Really impressed with this, thank you @carlydf you rock.

PR #304, opened 2026-04-27
deprecate authProxy.enabled value option

jaypipes:
Marks the `authProxy.enabled` values file option as deprecated and adds documentation to the `metrics.disableAuth` values file option explaining its effect.

Ensures that metrics port binding to localhost (127.0.0.1) only occurs when the auth proxy is enabled (via metrics.disableAuth=true).

Issue #167

### DIFF
--- a/helm/temporal-worker-controller/values.yaml
+++ b/helm/temporal-worker-controller/values.yaml
@@ -57,9 +57,8 @@ affinity: {}
 # More than one replica is required for high availability.
 replicas: 2
 
-# Opt out of these resources if you want to disable the
-# auth proxy (https://github.com/brancz/kube-rbac-proxy)
-# which protects your /metrics endpoint.
+# deprecated. disable authenticated metrics endpoint access with the
+# metrics.disableAuth value.
 authProxy:
   enabled: true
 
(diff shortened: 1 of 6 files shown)


helm/temporal-worker-controller/values.yaml
```diff
@@ -57,19 +57,28 @@ affinity: {}
 # More than one replica is required for high availability.
 replicas: 2
 
-# Opt out of these resources if you want to disable the
-# auth proxy (https://github.com/brancz/kube-rbac-proxy)
-# which protects your /metrics endpoint.
+# deprecated. disable authenticated metrics endpoint access with the
```
carlydf, line 60: Just to clarify, because in some other Temporal contexts, "deprecated" means "we are removing this soon but for now it still works," this helm value no longer has any impact in helm charts we release after this change goes in. That's ok with me (I think it didn't work well in the first place, which is why someone reported an issue), but just clarifying

carlydf (approved):

helm/temporal-worker-controller/values.yaml
jaypipes: @carlydf I left it in there and marked the value as deprecated in the values.schema.json file so that if someone uses that values field in a previous Helm installation, when they do `helm upgrade --install` it should show a deprecation warning. That's really all I think is possible here.... thoughts?

carlydf: ah makes sense. (the issue of compatibility with prev helm version).
I think theoretically we could put a comment saying "value only used in helm chart versions <=0.24.1", but not necessary imo

Issue #303, opened 2026-04-25
Remove unnecessary files from source control


jaypipes:
The main Git repository contains some unwanted objects, including [multiple](https://github.com/temporalio/temporal-worker-controller/blob/35371a9ac950005b3f6912ad5f3dce6933bb11f8/internal/demo/helloworld/worker) 30MB binary [executables](https://github.com/temporalio/temporal-worker-controller/blob/35371a9ac950005b3f6912ad5f3dce6933bb11f8/internal/demo/worker).

Issue #302, opened 2026-04-25
remove `go.work[.sum]` from repository


jaypipes:
It's best practice to keep `go.work` and `go.work.sum` files out of the source code repository. If we need to use Go workspaces in our build scripts, we can dynamically create them in our CI/CD pipelines.

PR #305, opened 2026-04-27
remove go.work and binaries from source control

jaypipes:
There were two binary files in the source code repository. Removed them from source control and added them to .gitignore. Removes the go.work and go.work.sum files as well, since it's best practice to not have those in source control. Finally, moved the go-vet Github Action step from the `test-integration` workflow into the `linters` workflow since `go vet` is a linter. Standardized the linter steps to always cache the setup-go Github Action results.

Issue #302
Issue #303


carlydf (approved):

.github/workflows/test-integration.yml
```diff
@@ -61,11 +61,17 @@ jobs:
...
     - name: Checkout code
       uses: actions/checkout@v4
-      
+
+    - name: setup go workspace
+      run: go work init && go work use -r .
+
     - name: Set up Go
-      uses: actions/setup-go@v5
+      uses: actions/setup-go@4b73464bb391d4059bd26b0524d20df3927bd417 # v6.3.0
```
Shivs11, line 69: ???
why can't we just use a v.6.3.0 here (just curious)

Shivs11 (approved):

.github/workflows/test-integration.yml
jaypipes: It's [best practice to pin actions to a specific SHA/commit hash](https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions). `@v5` isn't an immutable tag unfortunately. It's a mutable tag and therefore you don't actually know what code that GH action is installing.

jaypipes: I'll update all the rest of the GH action references to SHA commit hashes in a future PR.

PR #307, opened 2026-04-27
Include cluster UID in CONTROLLER_IDENTITY to prevent cross-cluster conflicts

carlydf:
## Summary

- **Go**: at startup, `main()` reads `POD_NAMESPACE`, fetches the namespace object via `mgr.GetAPIReader()`, and appends the namespace UID to `CONTROLLER_IDENTITY` (format: `{release}/{namespace}/{namespace-uid}`). No Helm changes required.
- **Go**: removed the hardcoded `"temporal-worker-controller"` fallback from `getControllerIdentity()` — the env var is now required.
- **Migration**: `shouldClaimManagerIdentity` is extended to reclaim ownership of Worker Deployments that were managed by an older controller version, so existing deployments are not orphaned after an upgrade.
- **Tests**: use a dedicated `testControllerIdentity` constant (not the deprecated default) so tests exercise the normal identity path rather than the migration branch.

## Why

Two controllers installed in different clusters but with the same release name and namespace shared the same identity string. That allowed them to both claim ownership of the same Temporal Worker Deployment. The controller's namespace UID is stable, unique per cluster, and already available without any Helm templating.

Appending the UID in Go (rather than in the Helm chart) means `helm template | kubectl apply` workflows work without any override — there is no offline vs. live-cluster split.

## Migration

Upgrading changes the controller's identity, so existing Worker Deployments whose `ManagerIdentity` was set by an older version need to be reclaimed. Without handling this, the new controller would see a non-empty `ManagerIdentity` it doesn't recognise as its own, skip the claim step, and be unable to manage those deployments going forward.

Two legacy identity formats are handled:

| Existing `ManagerIdentity` | Reason |
|---|---|
| `"temporal-worker-controller"` | Hardcoded default used before Helm set the identity |
| `"release/namespace"` | Helm-set identity before the namespace UID was appended |

Detection for the second case uses a prefix check: if the new identity is `"release/namespace/{uid}"` and the existing identity is `"release/namespace"`, the new identity starts with `existing + "/"`, so the controller knows it is looking at its own older identity and reclaims.

The old constant is retained in `defaults` as `DeprecatedDefaultControllerIdentity` with a comment explaining its purpose, rather than being deleted, to make the intent explicit.

## Empty identity enforcement

An empty `CONTROLLER_IDENTITY` is dangerous in two ways:
1. `SetManagerIdentity` with an empty string **clears** the `ManagerIdentity` field on the Worker Deployment, leaving it ownerless and claimable by any controller.
2. Other write operations (`SetCurrentVersion`, `SetRampingVersion`, etc.) with an empty `Identity` cause the SDK to substitute its own default (typically hostname), producing junk in the Temporal audit trail.

Three layers of enforcement:
- **`main()`** — fast-fail at process start for normal Helm deployments.
- **`Reconcile()`** — fallback check for when the reconciler is imported as a library and `main()` is not in the call path (e.g. embedded controller managers, tests).
- **`claimManagerIdentity()`** — last-resort guard specifically against the dangerous field-clear: refuses to call `SetManagerIdentity` if identity is empty.

## Test plan

- [ ] `go build ./...` passes
- [ ] `helm template` renders without error and without any cluster lookup
- [ ] Integration tests pass with `CONTROLLER_IDENTITY` set via `t.Setenv`
- [ ] Existing Worker Deployments with old-format `ManagerIdentity` are reclaimed after upgrade

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -5,6 +5,7 @@
 package main
 
 import (
+	"context"
 	"flag"
 	"log/slog"
 	"os"
@@ -13,7 +14,9 @@ import (
 	"github.com/temporalio/temporal-worker-controller/internal/controller"
 	"github.com/temporalio/temporal-worker-controller/internal/controller/clientpool"
 	"go.temporal.io/sdk/log"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
@@ -114,6 +117,20 @@ func main() {
 
 	if os.Getenv(controller.IdentityEnvKey) == "" {
 		setupLog.Error(nil, "CONTROLLER_IDENTITY environment variable must be set")
+		os.Exit(1)
+	}
+	podNamespace := os.Getenv("POD_NAMESPACE")
+	if podNamespace == "" {
+		setupLog.Error(nil, "POD_NAMESPACE environment variable must be set")
+		os.Exit(1)
+	}
+	var ns corev1.Namespace
+	if err := mgr.GetAPIReader().Get(context.Background(), types.NamespacedName{Name: podNamespace}, &ns); err != nil {
+		setupLog.Error(err, "unable to fetch namespace UID for controller identity")
+		os.Exit(1)
+	}
+	if err := os.Setenv(controller.IdentityEnvKey, os.Getenv(controller.IdentityEnvKey)+"/"+string(ns.UID)); err != nil {
+		setupLog.Error(err, "unable to set CONTROLLER_IDENTITY")
 		os.Exit(1)
 	}
 
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -112,6 +112,11 @@ func main() {
 		os.Exit(1)
 	}
 
+	if os.Getenv(controller.IdentityEnvKey) == "" {
+		setupLog.Error(nil, "CONTROLLER_IDENTITY environment variable must be set")
+		os.Exit(1)
+	}
+
 	setupLog.Info("starting manager")
 	if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
 		setupLog.Error(err, "problem running manager")
--- a/helm/temporal-worker-controller/templates/rbac.yaml
+++ b/helm/temporal-worker-controller/templates/rbac.yaml
@@ -74,6 +74,12 @@ rules:
   - apiGroups:
       - ""
     resources:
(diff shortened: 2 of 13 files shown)


jaypipes (changes requested):
Instead of adding a new `clusterID` values.yaml field and using the Helm `lookup` template function, how about just updating the controller Go code to always append the UID of the namespace that the controller is in to the end of the `{release.name}/{release.namespace}` existing value of the `CONTROLLER_IDENTITY` env var?

That way, there'd be no need for this new only-useful-when-running-helm-template values.yaml field and wouldn't need to change anything about the Helm chart at all.

cmd/main.go
```diff
@@ -112,6 +115,25 @@ func main() {
 		os.Exit(1)
 	}
 
+	if os.Getenv(controller.IdentityEnvKey) == "" {
+		setupLog.Error(nil, "CONTROLLER_IDENTITY environment variable must be set")
+		os.Exit(1)
+	}
+	podNamespace := os.Getenv("POD_NAMESPACE")
```
carlydf, line 122: already needed for WRT webhook (SA checks)

cmd/main.go
```diff
@@ -112,6 +115,25 @@ func main() {
...
+	}
+	var ns corev1.Namespace
+	if err := mgr.GetAPIReader().Get(context.Background(), types.NamespacedName{Name: podNamespace}, &ns); err != nil {
+		setupLog.Error(err, "unable to fetch namespace UID for controller identity")
+		os.Exit(1)
+	}
+	if err := os.Setenv(controller.IdentityEnvKey, os.Getenv(controller.IdentityEnvKey)+"/"+string(ns.UID)); err != nil {
+		setupLog.Error(err, "unable to set CONTROLLER_IDENTITY")
+		os.Exit(1)
+	}
```
Shivs11, line 135: so i have a feeling that this won't work only because we don't have the worker-controller's RBAC policies updated

i checked [here](https://github.com/temporalio/temporal-worker-controller/blob/9ff2733463d3c822a21af3abda0e7d0df210aac5/helm/temporal-worker-controller/templates/rbac.yaml#L62) and noticed that the `clusterRole` does not have namespaces in it

my feeling is that this call would then fail but happy to be proven wrong!

Shivs11: note - i just found out that envTest also won't test this since it's spins up a cluster with a permissive auth mode

Shivs11 (changes requested):

cmd/main.go
carlydf: we should just add describe namespace permission to the auth then, right?

helm/temporal-worker-controller/templates/rbac.yaml
```diff
@@ -71,6 +71,12 @@ rules:
     verbs:
       - create
       - patch
+  - apiGroups:
+      - ""
+    resources:
+      - namespaces
+    verbs:
+      - get
```
jaypipes, line 79: 👍

jaypipes (approved):
Thanks @carlydf! Much cleaner :)

PR #308, opened 2026-04-28
Prepare to include cluster UID in CONTROLLER_IDENTITY to prevent cross-cluster conflicts

carlydf:
## What was changed
In `shouldClaimManagerIdentity`, detect future format of manager identity

## Why?
To facilitate clean reclaim after rollback from the next release, which will include https://github.com/temporalio/temporal-worker-controller/pull/309

## Checklist

1. Closes 

2. How was this tested:
Envtest won't test our get ns permissions, so just trusting there (we will test it in CI env because it runs on controller startup)

3. Any docs updates needed?

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -5,6 +5,7 @@
 package main
 
 import (
+	"context"
 	"flag"
 	"log/slog"
 	"os"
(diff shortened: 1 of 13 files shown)


jaypipes (approved):
👍

PR #309, opened 2026-04-28
Include cluster UID in CONTROLLER_IDENTITY to prevent cross-cluster conflicts

carlydf:
Completes the two-release migration for including the cluster namespace UID in the controller identity string.

Previously, the new `{identity}/{namespaceUID}` format was prepared behind a "next release" comment while the bare `{identity}` string remained active. This PR flips that: `getControllerIdentity()` now returns the full `{identity}/{suffix}` format, and `getDeprecatedControllerIdentity()` handles reclamation of Worker Deployments previously claimed under the old format.

**Changes:**
- `getControllerIdentity()` now returns `{CONTROLLER_IDENTITY}/{CONTROLLER_IDENTITY_SUFFIX}`; returns empty string if either is unset
- `getDeprecatedControllerIdentity()` (renamed from `getControllerIdentityWithNamespaceUID`) is used only for backward-compatible reclamation
- Startup guard in `main()` fails fast if `CONTROLLER_IDENTITY` is unset
- Fallback guard in `Reconcile()` for library users who bypass `main()` requires that `getControllerIdentity() != ""`
- Safety check in `claimManagerIdentity()` refuses to call `SetManagerIdentity` with an empty string to prevent accidental field clearing
- Renamed `ToBeDeprecatedDefaultControllerIdentity` → `DeprecatedDefaultControllerIdentity`
- Updated tests to set both `IdentityEnvKey` and `IdentitySuffixEnvKey`


internal/controller/execplan.go
```diff
@@ -205,6 +205,14 @@ func (r *TemporalWorkerDeploymentReconciler) claimManagerIdentity(
 	vcfg *planner.VersionConfig,
 ) error {
 	identity := getControllerIdentity()
+	if identity == "" {
+		// Passing an empty identity to SetManagerIdentity clears the field on the
+		// Worker Deployment, leaving it ownerless. Refuse rather than cause that.
+		// This should never happen, but this is the extra fallback in case somehow
+		// the check in main() and Reconcile() are not sufficient.
```
jaypipes, line 212: 👍

jaypipes (approved):
Excellent, thank you @carlydf!

PR #312, opened 2026-05-01
document downgrade concerns for CRD rename migrate

jaypipes:
Adds a section explaining rollback procedures to the guide for the v1.6 -> v1.7 migration that involves the rename of TemporalWorkerDeployment CRD to WorkerDeployment.


docs/migration-crd-rename.md
```diff
@@ -119,3 +119,91 @@ If you delete a deprecated resource before creating its replacement, the resourc
...
+
+## Downgrading from v1.7 to v1.6
+
+There are some important things to consider if you want to roll back
+(downgrade) an installed version of Temporal Worker Controller.
+
+> **Warning**: You **should not perform a rollback/downgrade of the Temporal
+> Worker Controller CRDs Helm Chart**. Doing so is a potentially
+> **destructive** operation that can cause your Temporal Worker Deployments to
+> be deleted.
```
carlydf: It could make sense to link to https://github.com/temporalio/temporal-worker-controller/blob/main/docs/crd-management.md#crd-rollback-and-field-pruning here with a "See x for more details" or something.

But this is also a good opportunity to get your input on that section I just linked! It might be too gentle and not scary enough about CRD rollback, and we could also change it in this PR to get to the point quicker (It's a bit wordy right now).

carlydf (approved):
small comment but as a whole I approve! maybe we should run the runbook locally before merging (I can do that)

docs/migration-crd-rename.md
```diff
@@ -119,3 +119,91 @@ If you delete a deprecated resource before creating its replacement, the resourc
...
+
+Once you have downgraded the Temporal Worker Controller, you will need to take
+some corrective actions depending on how far down the migration path you went
+when upgrading to the v1.7 Temporal Worker Controller release.
+
+If you upgraded the Temporal Worker Controller to v1.7 -- i.e. you successfully
+completed Step 2 above -- but **did not** complete Step 3 (migrating your
+resources), execute the following `kubectl` command to remove the CRD rename
+validation guard on the old `TemporalWorkerDeployment` Custom Resource
+Definition:
```
Shivs11: for each of these "how deep are you in this game", should we also explain what is the main reason why folks can, for eg in this specific case, just delete the new WD instances and cannot do so directly in the next step?

docs/migration-crd-rename.md
```diff
@@ -119,3 +119,91 @@ If you delete a deprecated resource before creating its replacement, the resourc
 Ready=False reason=DeletingPendingMigration
 message: "This TemporalWorkerDeployment is marked for deletion. Create a WorkerDeployment with the same name and spec to complete migration; deletion will proceed automatically once migration is confirmed."
 ```
+
+## Downgrading from v1.7 to v1.6
```
Shivs11, line 126: I went through this document but did not see a mention of the new finalizer that we had added in the previous renaming PR, specifically:
temporal.io/migration-guard

I think if someone were to upgrade the controller to 1.7, and have the controller write this finalizer on existing TWD's, a rollback to v1.6 controller wouldn't remove them no? 

A potential downside to this could be that those TWD's could never be nuked?

docs/migration-crd-rename.md
```diff
@@ -119,3 +119,91 @@ If you delete a deprecated resource before creating its replacement, the resourc
...
+```
+
+Then get a list of all the Kubernetes Deployments that are now owned by the new
+`WorkerDeployment` resources:
+
+```bash
+kubectl get -n <NAMESPACE> deployments -o jsonpath='{range .items[*]}{.metadata.name}{"\t"}{.metadata.ownerReferences[0].kind}{"/"}{.metadata.ownerReferences[0].name}{"\n"}{end}'
+```
+
+Then, for each of the Kubernetes Deployments listed above, execute the
```
Shivs11: should we also have this being mentioned for WorkerResourceTemplates too?

docs/migration-crd-rename.md
```diff
@@ -119,3 +119,91 @@ If you delete a deprecated resource before creating its replacement, the resourc
...
+following `kubectl` command to reset the OwnerReferences of Kubernetes
+Deployments back to the original `TemporalWorkerDeployment` custom resources:
+
+```bash
+kubectl patch -n <NAMESPACE> deployment <DEPLOYMENT_NAME> --type='merge' -p '
+{
+  "metadata": {
+    "ownerReferences": [
+      {
+        "apiVersion": "v1alpha1",
```
Shivs11: shouldn't this be temporal.io/v1alpha1?

docs/migration-crd-rename.md
```diff
@@ -119,3 +119,91 @@ If you delete a deprecated resource before creating its replacement, the resourc
...
+when upgrading to the v1.7 Temporal Worker Controller release.
+
+If you upgraded the Temporal Worker Controller to v1.7 -- i.e. you successfully
+completed Step 2 above -- but **did not** complete Step 3 (migrating your
+resources), execute the following `kubectl` command to remove the CRD rename
+validation guard on the old `TemporalWorkerDeployment` Custom Resource
+Definition:
+
+```bash
+kubectl patch crd temporalworkerdeployments.temporal.io --type='json' -p='[{"op": "remove", "path": "/spec/versions/0/schema/openAPIV3Schema/properties/spec/x-kubernetes-validations"}]'
```
Shivs11: i think this is right btw, but I do think we should be having this on our TemporalConnections CRD type too no?

wdyt

Shivs11: that is: temporalconnections.temporal.io

docs/migration-crd-rename.md
carlydf: Can I piggy back on this PR to change this suggestion slightly?
```suggestion
> **Dev / non-production environments:** If you don’t need to preserve worker state, you can delete your `TemporalWorkerDeployment` and `TemporalConnection` resources while the v1.6 controller is still running. This will cause the controller to remove the associated Worker Deployment state in Temporal, leaving Task Queues unversioned. Once cleanup completes, upgrade the controller and recreate them as `WorkerDeployment` and `Connection` resources.
>
> In most cases, following the migration steps below is simpler.
```

docs/migration-crd-rename.md
carlydf: yes, we should add a line telling people to remove the migration-guard finalizer on their TWDs, if it exists (ie. if they upgraded to v1.7 but did not migrate resources all the way to the point of deleting their TWDs)

docs/migration-crd-rename.md
carlydf: agree people should be able to copy-paste the command for both

docs/migration-crd-rename.md
carlydf: I feel like giving people less to read is important for comprehension and compliance with the instructions, so we need to balance giving enough details with too many details. 

If the instructions are easy enough to read, they should just follow them. If we include a too-long "what happens if you don't follow these instructions" section, that could cause people to think "not following the instructions is an option" which we don't want.

docs/migration-crd-rename.md
jaypipes: OK, added.

docs/migration-crd-rename.md
jaypipes: OK, added a note about this.

docs/migration-crd-rename.md
jaypipes: added a link. and I think that section is great. well worded and just the right amount of scary. :)

docs/migration-crd-rename.md
jaypipes: done.

docs/migration-crd-rename.md
jaypipes: good catch, sorry. fixed.

docs/migration-crd-rename.md
jaypipes: added.

docs/migration-crd-rename.md
```diff
@@ -19,6 +19,11 @@ After this release, the Worker Controller will be Generally Available (GA), whic
 
 > **Dev / non-production environments:** If you don't need to preserve any worker state, the simplest path is to delete all your `TemporalWorkerDeployment` and `TemporalConnection` resources while the v1.6 controller is still running. At that point no migration-guard finalizer has been added yet, so deletion completes after the v1.6 finalizer completes. Note that all related Worker Deployment state in the Temporal server will also be deleted. Then upgrade the controller and create fresh `WorkerDeployment` and `Connection` resources.
```
carlydf: need to remove the top one still

docs/migration-crd-rename.md
```diff
@@ -119,3 +124,137 @@ If you delete a deprecated resource before creating its replacement, the resourc
...
+when upgrading to the v1.7 Temporal Worker Controller release.
+
+If you upgraded the Temporal Worker Controller to v1.7 -- i.e. you successfully
+completed Step 2 above -- but **did not** complete Step 3 (migrating your
+resources), execute the following `kubectl` command to remove the CRD rename
+validation guard on the old `TemporalWorkerDeployment` and `TemporalConnection`
+Custom Resource Definitions:
+
+```bash
+kubectl patch crd temporalworkerdeployments.temporal.io --type='json' -p='[{"op": "remove", "path": "/spec/versions/0/schema/openAPIV3Schema/properties/spec/x-kubernetes-validations"}]'
```
carlydf: needs to be
```
kubectl patch crd temporalworkerdeployments.temporal.io --type='json' -p='[{"op": "remove", "path": "/spec/versions/0/schema/openAPIV3Schema/x-kubernetes-validations"}]'
```

docs/migration-crd-rename.md
```diff
@@ -119,3 +124,137 @@ If you delete a deprecated resource before creating its replacement, the resourc
...
+```bash
+kubectl get -n <NAMESPACE> temporalworkerdeployments.temporal.io -o jsonpath='{range .items[*]}{.metadata.name}{"\t"}{.metadata.uid}{"\n"}{end}'
+```
+
+For each of the `TemporalWorkerDeployments` listed above, you will need to
+manually remove the finalizer that was added to your `TemporalWorkerDeployment`
+resource by the 1.7 controller:
+
+```bash
+kubectl patch -n <NAMESPACE> temporalworkerdeployments/<TWD_NAME> --type json -p='[{"op": "remove", "path": "/metadata/finalizers/0"}]'
```
carlydf: needs to be 
```
kubectl patch -n default temporalworkerdeployments/helloworld --type=merge -p='{"metadata":{"finalizers":["temporal.io/delete"]}}'
```
and should be done even if resource migration did not complete (v1.7 controller alone will set the `migration-guard` finalizer on this resource, which needs to be removed after rollback)

carlydf: also, same thing needed for temporalconnections

docs/migration-crd-rename.md
```diff
@@ -119,3 +124,137 @@ If you delete a deprecated resource before creating its replacement, the resourc
...
+        "blockOwnerDeletion": true
+      }
+    ]
+  }
+}'
+```
+
+Again, replace `<TWD_NAME>` and `<TWD_UID>` with the correct
+`TemporalWorkerDeployment` custom resource's name and UID you printed out
+earlier.
```
carlydf: now you can:
- delete your WorkerDeployment and Connection
patch to remove `kubectl patch -n default workerdeployments/helloworld --type=merge -p='{"metadata":{"finalizers":[]}}'`
- patch away the deprecation warning in the TWD and TConn CRDs or CAREFULLY roll back your CRD chart (rolling it back earlier will delete the

carlydf:
```
kubectl patch crd temporalworkerdeployments.temporal.io --type=json -p='[
    {"op":"remove","path":"/spec/versions/0/deprecated"},
    {"op":"remove","path":"/spec/versions/0/deprecationWarning"}
  ]'
```

carlydf (approved):

PR #314, opened 2026-05-01
upgrade SDK and server to latest for demo Worker UI

carlydf:
## What was changed

**SDK/server upgrade**
- Bumped `go.temporal.io/sdk` to v1.43.0 and `go.temporal.io/api` to v1.62.11 in `internal/demo/`.
- Set `WorkerHeartbeatInterval = 5s` in the demo client so the new Worker UI populates quickly.

**Demo tuning for conference**
- Faster sunset on the helloworld WRT: `scaledownDelay: 0s`, `deleteDelay: 30s` (was 30s / 5m) so deprecated versions retire visibly within a demo timeframe.
- `skaffold.yaml`: switched the helloworld profile to semantic image tags driven by `WORKER_VERSION`, so each `skaffold run` produces a meaningful build ID instead of skaffold's default content hash.

**Drop the `worker_version` label_replace workaround**
- Temporal Cloud now emits `temporal_worker_deployment_name` and `temporal_worker_build_id` as separate opt-in labels, so the regex-based recording-rule shim is no longer needed.
- `prometheus-stack-values.yaml`: opted in to the new labels via the scrape config's `params.labels`, and replaced the nested `label_replace` rule with a simple `sum by(...)` aggregation. The recording rule is renamed `temporal_backlog_count_by_version` → `temporal_approximate_backlog_count` to reflect that it's now a pure cardinality reducer over `temporal_cloud_v1_approximate_backlog_count`.
> Note: I don't know why it didn't work to simply use `temporal_cloud_v1_approximate_backlog_count` directly :( I tried for a few hours to use it directly with no rename + metric cardinality-reduce recording rule, but could not get that to work, so went with this rename + cardinality reduction because it worked and I just needed the demo to function. Now that the conference is over, I am trying again to make this work without a recording rule, but for now my preference would be to merge this PR since it is still better than what is currently in `main` using the deprecated `worker_version` label. I added the following note to the README to mention this, and will remove this note once I figure out the actual reason for the recording rule, or how to make it work without a recording rule:
>> **Note:** This demo ships a Prometheus recording rule that renames `temporal_cloud_v1_approximate_backlog_count` to `temporal_approximate_backlog_count` and reduces it to the labels the HPA cares about. In principle the HPA can consume the raw Cloud metric directly (set `namespaced: false` on the prometheus-adapter rule so it doesn't auto-inject a `namespace` label filter), but this demo uses the recording rule as a known-working path.

- Updated `prometheus-adapter-values.yaml`, `grafana-dashboard.json`, and `examples/wrt-hpa-backlog.yaml` to reference the new metric name and labels. The HPA example now also pins `temporal_task_queue` since it's preserved in the aggregation.

**Demo docs polish**
- `internal/demo/README.md`: link to the [Temporal OpenMetrics reference](https://docs.temporal.io/cloud/metrics/openmetrics) and the [opt-in labels docs](https://docs.temporal.io/cloud/metrics/openmetrics/metrics-reference#opt-in-labels); added a metrics-API-key rotation snippet; switched all install instructions to `helm upgrade --install` so they're idempotent on re-runs.

## Why?

Three motivations:
1. The demo Worker UI needs Go SDK ≥ v1.41.0, and the 5s heartbeat makes the worker visible there nearly instantly.
2. Conference-demo pacing: the previous 5-minute delete delay made sunset invisible during a live walkthrough.
3. Now that Temporal Cloud emits the per-version labels natively, the regex shim was extra code with a footgun (broke on build IDs containing `_`). Removing it simplifies the HPA recipe and unblocks documenting it as the recommended pattern.

## Checklist

1. Closes N/A — demo polish, no tracking issue.

2. How was this tested:
   - `skaffold run -p helloworld-worker` against a local cluster, verified semantic tags land on the Deployment.
   - Port-forwarded Prometheus and confirmed `temporal_approximate_backlog_count` returns series with the new labels.
   - Watched a rollout in the demo to confirm sunset timing and Worker UI visibility.

3. Any docs updates needed?
   - `internal/demo/README.md` is updated in this PR. No changes needed in docs.temporal.io.


api/v1alpha1/workerdeployment_webhook.go
```diff
@@ -19,8 +19,9 @@ import (
 )
 
 func (r *WorkerDeployment) SetupWebhookWithManager(mgr ctrl.Manager) error {
-	return ctrl.NewWebhookManagedBy(mgr).
-		For(r).
+	return ctrl.NewWebhookManagedBy(mgr, r).
```
carlydf, line 22: needed by new k8s version

api/v1alpha1/workerresourcetemplate_webhook.go
```diff
@@ -84,9 +84,8 @@ func NewWorkerResourceTemplateValidator(mgr ctrl.Manager) *WorkerResourceTemplat
 //
 // +kubebuilder:webhook:path=/validate-temporal-io-v1alpha1-workerresourcetemplate,mutating=false,failurePolicy=fail,sideEffects=None,groups=temporal.io,resources=workerresourcetemplates,verbs=create;update;delete,versions=v1alpha1,name=vworkerresourcetemplate.kb.io,admissionReviewVersions=v1
 func (v *WorkerResourceTemplateValidator) SetupWebhookWithManager(mgr ctrl.Manager) error {
-	return ctrl.NewWebhookManagedBy(mgr).
-		For(&WorkerResourceTemplate{}).
-		WithValidator(v).
+	return ctrl.NewWebhookManagedBy(mgr, &WorkerResourceTemplate{}).
+		WithCustomValidator(v).
```
carlydf, line 88: needed by new k8s version

internal/demo/helloworld/helm/helloworld/templates/deployment.yaml
```diff
@@ -28,8 +28,8 @@ spec:
     gate:
       workflowType: "RolloutGate"
   sunset:
-    scaledownDelay: 30s
-    deleteDelay: 5m
+    scaledownDelay: 0s
```
carlydf, line 31: faster for demo

api/v1alpha1/workerresourcetemplate_webhook_test.go
```diff
@@ -253,7 +253,7 @@ func TestWorkerResourceTemplate_ValidateCreate(t *testing.T) {
 							"type": "External",
 							"external": map[string]interface{}{
 								"metric": map[string]interface{}{
-									"name": "temporal_backlog_count_by_version",
+									"name": "temporal_approximate_backlog_count",
```
Shivs11: could we also have a pr description as to why we are changing this value here? (as in like we have our metric have these new sep labels now and we don't need the recording rules)

Shivs11 (approved):

api/v1alpha1/workerresourcetemplate_webhook_test.go
jaypipes: Agree with @Shivs11 here. This change was a surprise for me.

carlydf: the metric name in these tests doesn't actually matter, because these tests are just confirming that the WRT is validated correctly (which does not include any validation that this metric actually exists). so this file didn't need to be changed at all, it was just spillover from a find-and-replace that Claude did. Sorry this slipped through my human review!

carlydf: reverted this file to main

jaypipes (approved):
Thanks @carlydf for the updated PR summary explaining the change in metric.

Issue #286, opened 2026-04-21
Support KEDA ScaledObject Temporal trigger as `WorkerResourceTemplate`


carlydf:
### Is your feature request related to a problem? Please describe.

Some people may prefer to use KEDA for scaling for various reasons. For example:
- already have the KEDA External Metrics Provider in their cluster, and therefore are not able to install another metrics adapter as the External Metrics Provider in the same cluster
- already use KEDA Temporal trigger for unversioned workers, and want to migrate to controller-supported versioning without changing their scaling resources

### Describe the solution you'd like

https://github.com/kedacore/keda/pull/7672 adds support for Worker Deployment Versioning to KEDA [`temporal` trigger](https://keda.sh/docs/2.20/scalers/temporal/). 

The `WorkerResourceTemplate` templating code in the worker controller will need to populate the new `workerDeploymentName` and `workerDeploymentBuildId` fields for each per-version KEDA ScaledObject it creates.

It will also need to fill in the `scaleTargetRef` of the KEDA ScaledObject with the per-version k8s Deployment name (might already do this, but need to test).

### Additional context

We should also decide whether KEDA ScaledObject should be a white-listed resource in the default value for `workerResourceTemplate.allowedResources`. That is optional.

dkozlov:
Similar issue https://github.com/temporalio/temporal-worker-controller/issues/320

jaypipes:
Closing this out. This work is included in the [v1.8.0 release](https://github.com/temporalio/temporal-worker-controller/releases/tag/v1.8.0) of Temporal Worker Controller.

PR #324, opened 2026-05-14
Add docs recommending autoscaling setup

carlydf:
Adds documentation outlining the tradeoffs between two autoscaling solutions:

1. HPA+prometheus adapter
2. KEDA Temporal Scaler

Documentation focuses on straightforward descriptions of the pros and cons of each solution.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -39,6 +39,9 @@ See [Migration to Unversioned](migration-to-unversioned.md) for how to migrate b
 ### [Ownership](manager-identity.md)
 How the controller gets permission to manage a Worker Deployment, how a human client can take or give back control.
 
+### [Scaling Recommendations](scaling-recommendations.md)
+Practical reactivity and reliability tradeoffs between HPA + prometheus-adapter and KEDA when scaling Temporal workers per worker-deployment-version. Covers steady-state reactivity (~3:15 via the metric path), task-queue unloading, scale-from-zero limits, and when to pick which tool.
+
 ### [WorkerResourceTemplate](worker-resource-templates.md)
 How to attach HPAs, PodDisruptionBudgets, and other Kubernetes resources to each active versioned Deployment. Covers the auto-injection model, RBAC setup, webhook TLS, and examples.
 
--- a/docs/scaling-recommendations.md
+++ b/docs/scaling-recommendations.md
@@ -29,13 +29,17 @@ backlog appears at T0
 
 **Typical end-to-end reactivity is ≈ 85 seconds + your stabilization window.** Empirically, sample age in Prometheus for a single series follows a sawtooth between 0 and 60 seconds (matching the gateway's ~1/min emission cadence). p50 sample age ≈ 30s, p95 ≈ 50s. The 60-second emission cadence is the inherent floor — smaller scrape intervals, tighter `metricsRelistInterval`, or recording rules cannot improve it because they all consume the same upstream cadence.
 
-### Caveat: gateway-wide stalls
+### Caveat: gateway delivery delay
 
-During our investigation we observed one period of several minutes during which Temporal Cloud's OpenMetrics endpoint returned frozen timestamps for *every* series across the account — backlog series, action counts, error counts, every queue, every namespace, all showing the exact same staleness simultaneously (e.g. all ~30 visible series reading 239s old, identical to the second). The Prometheus scrape continued to succeed (`up{job="temporal_cloud"}` stayed 1, HTTP 200 responses) — only the embedded timestamps in the response body did not advance. During such a period, HPAs see the last known value (via Prometheus staleness lookback) until either (a) fresh samples resume, or (b) the staleness window (5 min default) expires and the metric disappears entirely.
+During our investigation we observed periods of several minutes during which Temporal Cloud's OpenMetrics endpoint returned the same embedded timestamps on repeated scrapes for *every* series across the account simultaneously — backlog series, action counts, error counts, every queue, every namespace, all showing identical staleness to the second (e.g. all ~30 visible series reading 239s old at once). The Prometheus scrape continued to succeed (`up{job="temporal_cloud"}` stayed 1, HTTP 200 responses) — the response body simply repeated already-known samples instead of advancing.
 
-We have only directly characterized this once, so frequency and typical duration are not yet known. The behavior is open with Temporal's Observability team. If your workload cannot tolerate occasional multi-minute scaling pauses, prefer KEDA.
+Once the delay resolved, the gateway delivered the missing samples with their original minute-aligned timestamps in a burst, so Prometheus's storage ends up with a complete 1/minute series in retrospect. We verified this directly: across a 3-hour window covering one such delay event, every gap between consecutive sample timestamps was exactly 60 seconds, no exceptions.
 
-This is also why `metricsRelistInterval: 5m` is the recommended setting: the discovery window must comfortably exceed the longest expected gap so the metric does not deregister, otherwise re-registration waits up to one more relist cycle after data flows again.
+The retrospective completeness is helpful for dashboards and post-hoc analysis, but it does **not** help an HPA, which queries the *latest available* value at decision time. During a delivery delay, the latest available sample is the one from before the delay started. The HPA sees real staleness even though the underlying record will eventually be filled in.
+
+We have only directly characterized this pattern during one investigation session (seeing it twice in ~2 hours of close observation). Frequency in normal operation is not yet known and is open with Temporal's Observability team. If your workload cannot tolerate occasional multi-minute scaling pauses, prefer KEDA.
+
+This is also why `metricsRelistInterval: 5m` is the recommended setting: the discovery window must comfortably exceed the longest expected delay so the metric does not deregister, otherwise re-registration waits up to one more relist cycle after delivery resumes.
 
 ### Slot utilization is a much faster leading signal
 
--- a/docs/scaling-recommendations.md
+++ b/docs/scaling-recommendations.md
@@ -31,11 +31,11 @@ backlog appears at T0
 
 ### Caveat: gateway-wide stalls
 
-We have observed occasional periods of several minutes during which Temporal Cloud's OpenMetrics gateway returns frozen timestamps for *every* series across the account — backlog series, action counts, error counts, every queue, every namespace. The Prometheus scrape continues to succeed (`up{job="temporal_cloud"}` stays 1, HTTP 200 responses), but the embedded timestamps in the response do not advance. During such a stall, HPAs see the last known value (via Prometheus staleness lookback) until either (a) fresh samples resume, or (b) the staleness window (5 min default) expires and the metric disappears entirely.
+During our investigation we observed one period of several minutes during which Temporal Cloud's OpenMetrics endpoint returned frozen timestamps for *every* series across the account — backlog series, action counts, error counts, every queue, every namespace, all showing the exact same staleness simultaneously (e.g. all ~30 visible series reading 239s old, identical to the second). The Prometheus scrape continued to succeed (`up{job="temporal_cloud"}` stayed 1, HTTP 200 responses) — only the embedded timestamps in the response body did not advance. During such a period, HPAs see the last known value (via Prometheus staleness lookback) until either (a) fresh samples resume, or (b) the staleness window (5 min default) expires and the metric disappears entirely.
 
-Frequency and duration of these stalls is still being characterized. They are the dominant reactivity-tail risk for the metric path. If your workload cannot tolerate occasional multi-minute scaling pauses, prefer KEDA.
+We have only directly characterized this once, so frequency and typical duration are not yet known. The behavior is open with Temporal's Observability team. If your workload cannot tolerate occasional multi-minute scaling pauses, prefer KEDA.
 
-This is also why `metricsRelistInterval: 5m` is the recommended setting: the discovery window must comfortably exceed the longest expected stall, otherwise the metric will deregister during a stall and only re-register on the next adapter discovery cycle after data flows again.
+This is also why `metricsRelistInterval: 5m` is the recommended setting: the discovery window must comfortably exceed the longest expected gap so the metric does not deregister, otherwise re-registration waits up to one more relist cycle after data flows again.
 
 ### Slot utilization is a much faster leading signal
 
--- a/docs/scaling-recommendations.md
+++ b/docs/scaling-recommendations.md
@@ -1,0 +1,152 @@
+# Scaling Recommendations: HPA + prometheus-adapter vs KEDA
+
+This document describes practical reactivity and reliability tradeoffs when scaling Temporal workers per worker-deployment-version on Kubernetes, and recommends which tool fits which workload pattern.
+
+The `internal/demo/` example wires the HPA path described here. The KEDA path is mentioned for comparison and as a recommendation for workloads that cannot tolerate the HPA path's limits.
+
+## TL;DR — Pick by workload pattern
+
+| Workload pattern | Recommendation |
+|------------------|----------------|
+| Continuous traffic (task queue always loaded) | HPA + prometheus-adapter, scaling on slot utilization + backlog count |
+| Idle periods >5 min between work; needs scale-from-zero | KEDA Temporal scaler |
+| Required reactivity < ~3 min from first backlog | KEDA Temporal scaler |
+| Required reactivity ~3–5 min, queue always loaded | HPA + prometheus-adapter is fine |
+| Very large namespaces (N × M HPAs polling) | HPA + prometheus-adapter; KEDA hits the 50 RPS namespace rate limit |
+
+## The reactivity model for HPA + prometheus-adapter
+
+For a continuously-loaded task queue, the end-to-end delay from "backlog appears" to "HPA scales up" decomposes as:
+
+```
+backlog appears at T0
+  └─ Temporal Cloud metric pipeline aggregation     +~3 min  (out of your control)
+       └─ Prometheus scrape interval                 +~10 s
+            └─ HPA poll interval                     +~15 s
+                 └─ scale-up stabilization window    +~your config
+                      └─ first replica added
+```
+
+**Steady-state reactivity is ≈ 3 minutes 15 seconds + your stabilization window.** Empirically measured against Temporal Cloud, this is dominated by the embedded-timestamp lag in the OpenMetrics `temporal_cloud_v1_approximate_backlog_count` series — the metric values land in Prometheus with timestamps roughly 3 minutes behind wall clock. With `honor_timestamps: true` in the scrape config (the correct setting; preserves source-of-truth timing), this lag flows through to the HPA.
+
+You cannot improve this with smaller scrape intervals, tighter `metricsRelistInterval`, or recording rules. The Temporal Cloud aggregation pipeline is the floor.
+
+### Slot utilization is a much faster leading signal
+
+`temporal_slot_utilization` is emitted directly by worker pods (no Temporal Cloud aggregation), scraped at the ServiceMonitor interval (~10–30 s), and reflects current state. It also rises *before* backlog accumulates — slots saturate first, then queueing starts. So a two-metric HPA with both slot util and backlog gives you fast scale-up via slot util and a backlog-driven backstop.
+
+The demo HPA uses both. For production scaling we recommend keeping both as well.
+
+## When backlog metric goes silent
+
+Two distinct failure modes that look similar in HPA events but have different meanings:
+
+### Mode 1: adapter-level deregistration (rare)
+- Trigger: prometheus-adapter pod restart, or *no* series matching the rule's `seriesQuery` exist in Prometheus.
+- Symptom in HPA events: `the server could not find the metric ...`.
+- Recovery: up to one `metricsRelistInterval` after data flows again.
+
+prometheus-adapter periodically asks Prometheus "what series exist in the last `metricsRelistInterval`?" — see the [prometheus-adapter README](https://github.com/kubernetes-sigs/prometheus-adapter/blob/master/README.md). If the discovery query window is shorter than the embedded-timestamp lag of your source data, the discovery returns empty and the metric name disappears from the External Metrics API. That's why this demo configures `metricsRelistInterval: 5m` — empirically the smallest value that reliably catches the 3-min-stale Temporal Cloud samples.
+
+### Mode 2: series-level silence (common in low-traffic workloads)
+- Trigger: a task queue with no polls or new tasks for >5 minutes. Temporal unloads it from memory and stops emitting `temporal_cloud_v1_approximate_backlog_count` for that specific `(task_queue, build_id, ...)` labelset. Other queues' series continue to emit.
+- Symptom in HPA events: `no metrics returned from external metrics API`. The metric *name* is still registered; the HPA's specific label selector just matches zero rows now.
+- Recovery: traffic resumes → queue reloads → next emission cycle (~1 min) + 3-min aggregation lag → HPA can read value again.
+
+In a two-metric HPA configured with slot utilization, this is mostly fine: the HPA reports `ScalingActive=True` based on slot utilization while backlog is unavailable, and rejoins backlog scaling once it returns. We've confirmed this empirically in this demo cluster — the HPA continued scaling correctly on slot utilization through 1000+ backlog `FailedGetExternalMetric` events.
+
+## Why this demo does not use a backlog recording rule
+
+A prior version of this demo wrapped the raw Cloud series in a Prometheus recording rule:
+
+```yaml
+- record: temporal_approximate_backlog_count
+  expr: sum by (...) (temporal_cloud_v1_approximate_backlog_count)
+```
+
+The rule was originally added to work around a label-formatting issue in an older Temporal Cloud release. With native per-version labels (`temporal_worker_deployment_name`, `temporal_worker_build_id`) now opt-in, the rule no longer earns its keep:
+
+- **It doesn't reduce reactivity.** The HPA reactivity floor is the upstream Temporal Cloud aggregation lag, not anything the rule could fix.
+- **It duplicates the cardinality bill.** Per-`(task_queue, build_id)` labels are already opt-in at the OpenMetrics level *because* of cardinality. Adding a recording rule on top means storing the same high-cardinality series twice.
+- **It hides a `sum(...)` that the adapter already does.** prometheus-adapter's `metricsQuery: sum(<<.Series>>{<<.LabelMatchers>>})` performs the same collapsing at query time. Pedagogically, "the adapter does the sum" is cleaner than "a recording rule sums first, then the adapter sums again."
+- **It does not solve series-level silence.** When the source goes silent (task queue unloaded), the rule output also goes silent eventually (once Prometheus's staleness lookback expires).
+
+What the recording rule *does* buy is registration stability after operational events: when the source series is sparse-by-timestamp, the rule produces a dense 10-second sample stream that lets the adapter discover with a tight `metricsRelistInterval`. If you find yourself fighting registration flicker on every adapter restart and would rather pay the cardinality cost than tune `metricsRelistInterval`, a recording rule is a reasonable choice. Otherwise, prefer the raw metric.
+
+In this demo we set `metricsRelistInterval: 5m` and consume the raw metric directly.
+
+## Why prometheus-adapter cannot do scale-from-zero
+
+Scale-from-zero on backlog through the metric path requires the metric to exist while there are zero workers. It does not:
+
+1. Zero workers means no polls.
+2. No polls for >5 minutes means the task queue is unloaded from Temporal Cloud's memory.
+3. An unloaded queue emits no metric.
+4. Adapter discovery returns no series, or HPA queries return no rows.
+5. HPA cannot scale up because there's no signal to scale on.
+
+Submitting a workflow does load the task queue back into memory, but the metric still won't reach the HPA for ~3 minutes (Temporal Cloud aggregation lag + scrape + poll). By the time the HPA reacts, you've already had 3+ minutes of unprovisioned work.
+
+KEDA's Temporal scaler calls `DescribeTaskQueue(stats=true)` (or `DescribeWorkerDeploymentVersion`), which loads the queue synchronously and returns the backlog directly. No metric pipeline involved. Scale-from-zero in seconds.
+
+## When KEDA hits its own limits
+
+KEDA bypasses the metric pipeline but uses Temporal API calls, which are subject to a per-namespace rate limit:
+
+```
+FrontendGlobalWorkerDeploymentReadRPS = 50  # per namespace, evenly distributed across frontend instances
+```
+
+For a namespace with N task queues × M worker-deployment-versions = K HPAs, each KEDA poll uses ~1 API call. The polling budget:
+
+| HPA count | Poll every 30s | Poll every 10s | Poll every 5s |
+|-----------|----------------|----------------|---------------|
+| 50        | 1.7 RPS (3%)   | 5 RPS (10%)    | 10 RPS (20%)  |
+| 250       | 8 RPS (17%)    | 25 RPS (50%)   | 50 RPS (100%) |
+| 1500      | 50 RPS (100%)  | exceeds limit  | exceeds limit |
+
+prometheus-adapter has no equivalent per-namespace bottleneck — one OpenMetrics scrape returns all series for the namespace in a single HTTP request, scaling independently of HPA count.
+
+So for very large namespaces (hundreds–thousands of HPAs) needing fast reactivity, neither path is great: KEDA hits the API rate limit, and the metric path has the 3-min aggregation floor. In practice this is a "talk to your account team" situation.
+
+## Recommended configuration for the HPA + prometheus-adapter path
+
+This demo's configuration represents the recommendation, in compact form:
+
+**Scrape config** (`internal/demo/k8s/prometheus-stack-values.yaml`):
+```yaml
+- job_name: temporal_cloud
+  scrape_interval: 10s
+  honor_timestamps: true
+  metrics_path: /v1/metrics
+  params:
+    labels:
+      - temporal_worker_deployment_name
+      - temporal_worker_build_id
+```
+
+**prometheus-adapter rule** (`internal/demo/k8s/prometheus-adapter-values.yaml`):
+```yaml
+metricsRelistInterval: 5m   # must accommodate Cloud's ~3-min embedded-timestamp lag
+rules:
+  external:
+    - seriesQuery: 'temporal_cloud_v1_approximate_backlog_count{temporal_worker_build_id!="__unversioned__"}'
+      metricsQuery: 'sum(<<.Series>>{<<.LabelMatchers>>})'
+      name:
+        as: "temporal_cloud_v1_approximate_backlog_count"
+      resources:
+        namespaced: false
+```
+
+The `seriesQuery` filter excludes `__unversioned__` series. Without it, accounts with many unversioned namespaces produce 5000+ series in the discovery response, which slows or breaks adapter discovery. The filter scopes discovery to versioned workloads — exactly the ones HPAs need.
+
+**HPA template** (`examples/wrt-hpa-backlog.yaml`): two metrics — slot utilization (fast leading signal, scale-up gate) and backlog count (confirming signal, AverageValue target).
+
+## References
+
+- [Temporal Cloud OpenMetrics](https://docs.temporal.io/cloud/metrics/openmetrics) — endpoint and opt-in labels
+- [prometheus-adapter README](https://github.com/kubernetes-sigs/prometheus-adapter/blob/master/README.md) — `metrics-relist-interval` and discovery window semantics
+- [prometheus-adapter externalmetrics.md](https://github.com/kubernetes-sigs/prometheus-adapter/blob/master/docs/externalmetrics.md) — external rules, `namespaced: false` for cluster-scoped metrics
+- [Prometheus HTTP API: `/api/v1/series`](https://prometheus.io/docs/prometheus/latest/querying/api/#finding-series-by-label-matchers) — series discovery semantics
+- [Prometheus scrape config: `honor_timestamps`](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config) — preserving source timestamps
+- [KEDA Temporal scaler](https://keda.sh/docs/latest/scalers/temporal/) — direct API polling alternative
--- a/docs/scaling-recommendations.md
+++ b/docs/scaling-recommendations.md
@@ -10,8 +10,8 @@ The `internal/demo/` example wires the HPA path described here. The KEDA path is
 |------------------|----------------|
 | Continuous traffic (task queue always loaded) | HPA + prometheus-adapter, scaling on slot utilization + backlog count |
 | Idle periods >5 min between work; needs scale-from-zero | KEDA Temporal scaler |
-| Required reactivity < ~3 min from first backlog | KEDA Temporal scaler |
-| Required reactivity ~3–5 min, queue always loaded | HPA + prometheus-adapter is fine |
+| Required reactivity < ~60 s from first backlog | KEDA Temporal scaler |
+| Required reactivity ~90 s typical, tolerant of occasional multi-minute stalls | HPA + prometheus-adapter is fine |
 | Very large namespaces (N × M HPAs polling) | HPA + prometheus-adapter; KEDA hits the 50 RPS namespace rate limit |
 
 ## The reactivity model for HPA + prometheus-adapter
@@ -20,16 +20,22 @@ For a continuously-loaded task queue, the end-to-end delay from "backlog appears
 
 ```
 backlog appears at T0
-  └─ Temporal Cloud metric pipeline aggregation     +~3 min  (out of your control)
+  └─ Temporal Cloud OpenMetrics emission cadence    +~60 s worst-case  (~1 sample/minute)
        └─ Prometheus scrape interval                 +~10 s
             └─ HPA poll interval                     +~15 s
                  └─ scale-up stabilization window    +~your config
                       └─ first replica added
 ```
 
-**Steady-state reactivity is ≈ 3 minutes 15 seconds + your stabilization window.** Empirically measured against Temporal Cloud, this is dominated by the embedded-timestamp lag in the OpenMetrics `temporal_cloud_v1_approximate_backlog_count` series — the metric values land in Prometheus with timestamps roughly 3 minutes behind wall clock. With `honor_timestamps: true` in the scrape config (the correct setting; preserves source-of-truth timing), this lag flows through to the HPA.
+**Typical end-to-end reactivity is ≈ 85 seconds + your stabilization window.** Empirically, sample age in Prometheus for a single series follows a sawtooth between 0 and 60 seconds (matching the gateway's ~1/min emission cadence). p50 sample age ≈ 30s, p95 ≈ 50s. The 60-second emission cadence is the inherent floor — smaller scrape intervals, tighter `metricsRelistInterval`, or recording rules cannot improve it because they all consume the same upstream cadence.
 
-You cannot improve this with smaller scrape intervals, tighter `metricsRelistInterval`, or recording rules. The Temporal Cloud aggregation pipeline is the floor.
+### Caveat: gateway-wide stalls
+
+We have observed occasional periods of several minutes during which Temporal Cloud's OpenMetrics gateway returns frozen timestamps for *every* series across the account — backlog series, action counts, error counts, every queue, every namespace. The Prometheus scrape continues to succeed (`up{job="temporal_cloud"}` stays 1, HTTP 200 responses), but the embedded timestamps in the response do not advance. During such a stall, HPAs see the last known value (via Prometheus staleness lookback) until either (a) fresh samples resume, or (b) the staleness window (5 min default) expires and the metric disappears entirely.
+
+Frequency and duration of these stalls is still being characterized. They are the dominant reactivity-tail risk for the metric path. If your workload cannot tolerate occasional multi-minute scaling pauses, prefer KEDA.
+
+This is also why `metricsRelistInterval: 5m` is the recommended setting: the discovery window must comfortably exceed the longest expected stall, otherwise the metric will deregister during a stall and only re-register on the next adapter discovery cycle after data flows again.
 
 ### Slot utilization is a much faster leading signal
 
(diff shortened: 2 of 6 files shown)


docs/README.md
```diff
@@ -39,6 +39,9 @@ See [Migration to Unversioned](migration-to-unversioned.md) for how to migrate b
 ### [Ownership](manager-identity.md)
 How the controller gets permission to manage a Worker Deployment, how a human client can take or give back control.
 
+### [Scaling Recommendations](scaling-recommendations.md)
```
jaypipes, line 42: Recommend adding this new doc to the list [here](https://github.com/temporalio/temporal-worker-controller/blob/d0d1be5d89327f6e1700f43018cd88fcd0981253/README.md#L144).

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,162 @@
+# Scaling Recommendations: HPA + prometheus-adapter vs KEDA
```
jaypipes:
```suggestion
# Scaling Recommendations
```

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,162 @@
...
+
+This document describes practical reactivity and reliability tradeoffs when scaling Temporal workers per worker-deployment-version on Kubernetes, and recommends which tool fits which workload pattern.
```
jaypipes:
```suggestion
This document describes practical reactivity and reliability tradeoffs when scaling Temporal workers per worker deployment version on Kubernetes, and recommends which tool fits which workload pattern.
```

here and throughout the doc... we use "worker deployment version", not "worker-deployment-version", elsewhere in the docs.

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,162 @@
...
+
+The `internal/demo/` example wires the HPA path described here. The KEDA path is mentioned for comparison and as a recommendation for workloads that cannot tolerate the HPA path's limits.
+
+## TL;DR — Pick by workload pattern
```
jaypipes:
```suggestion
## TL;DR

We recommend choosing a scaler approach that aligns with the workload pattern your application exhibits.
```

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,162 @@
...
+
+| HPA count | Poll every 30s | Poll every 10s | Poll every 5s |
+|-----------|----------------|----------------|---------------|
+| 50        | 1.7 RPS (3%)   | 5 RPS (10%)    | 10 RPS (20%)  |
+| 250       | 8 RPS (17%)    | 25 RPS (50%)   | 50 RPS (100%) |
+| 1500      | 50 RPS (100%)  | exceeds limit  | exceeds limit |
+
+prometheus-adapter has no equivalent per-namespace bottleneck — one OpenMetrics scrape returns all series for the namespace in a single HTTP request, scaling independently of HPA count.
+
+So for very large namespaces (hundreds–thousands of HPAs) needing fast reactivity, neither path is great: KEDA hits the API rate limit, and the metric path has the 3-min aggregation floor. In practice this is a "talk to your account team" situation.
```
jaypipes:
```suggestion
If you are using KEDA with Temporal Cloud and hitting the API rate limit described above, you will need to contact your Temporal Cloud account team to discuss increasing the rate limits.
```

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,162 @@
...
+
+For a namespace with N task queues × M worker-deployment-versions = K HPAs, each KEDA poll uses ~1 API call. The polling budget:
+
+| HPA count | Poll every 30s | Poll every 10s | Poll every 5s |
+|-----------|----------------|----------------|---------------|
+| 50        | 1.7 RPS (3%)   | 5 RPS (10%)    | 10 RPS (20%)  |
+| 250       | 8 RPS (17%)    | 25 RPS (50%)   | 50 RPS (100%) |
+| 1500      | 50 RPS (100%)  | exceeds limit  | exceeds limit |
+
+prometheus-adapter has no equivalent per-namespace bottleneck — one OpenMetrics scrape returns all series for the namespace in a single HTTP request, scaling independently of HPA count.
```
jaypipes:
```suggestion
```

(this section is about KEDA's limits, not HPA's. We already cover HPA's limitations in the previous section. I've moved this to a "HPA strengths" section above.)

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,162 @@
...
+2. No polls for >5 minutes means the task queue is unloaded from Temporal Cloud's memory.
+3. An unloaded queue emits no metric.
+4. Adapter discovery returns no series, or HPA queries return no rows.
+5. HPA cannot scale up because there's no signal to scale on.
+
+Submitting a workflow does load the task queue back into memory, but the metric still won't reach the HPA until the next OpenMetrics emission cycle (~1 minute) plus scrape and HPA poll. By the time the HPA reacts, you've already had ~1+ minute of unprovisioned work.
+
+KEDA's Temporal scaler calls `DescribeTaskQueue(stats=true)` (or `DescribeWorkerDeploymentVersion`), which loads the queue synchronously and returns the backlog directly. No metric pipeline involved. Scale-from-zero in seconds.
+
+## When KEDA hits its own limits
```
jaypipes:
```suggestion
## KEDA limitations
```

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,162 @@
...
+- **It hides a `sum(...)` that the adapter already does.** prometheus-adapter's `metricsQuery: sum(<<.Series>>{<<.LabelMatchers>>})` performs the same collapsing at query time. Pedagogically, "the adapter does the sum" is cleaner than "a recording rule sums first, then the adapter sums again."
+- **It does not solve series-level silence.** When the source goes silent (task queue unloaded), the rule output also goes silent eventually (once Prometheus's staleness lookback expires).
+
+What the recording rule *does* buy is registration stability after operational events: when the source series is sparse-by-timestamp, the rule produces a dense 10-second sample stream that lets the adapter discover with a tight `metricsRelistInterval`. If you find yourself fighting registration flicker on every adapter restart and would rather pay the cardinality cost than tune `metricsRelistInterval`, a recording rule is a reasonable choice. Otherwise, prefer the raw metric.
+
+In this demo we set `metricsRelistInterval: 5m` and consume the raw metric directly.
+
+## Why prometheus-adapter cannot do scale-from-zero
```
jaypipes:
```suggestion
## HPA limitations

This section describes two known limitations for HPA + prometheus adapter.

Temporal Cloud's OpenMetrics endpoint may sometimes return the same embedded timestamps on repeated scrapes for each series across the account simultaneously — backlog series, action counts, error counts, every queue, every namespace. This delay in returning fresh metrics data can impact the speed to which HPA + prometheus adapter scales out or in the replica count for a worker deployment version. This means that HPA + prometheus adapter may not be a good solution if your workload cannot tolerate occasional multi-minute scaling pauses.

> **Note**: This is why `metricsRelistInterval: 5m` is the recommended setting: the discovery window must comfortably exceed the longest expected delay so the metric does not deregister, otherwise re-registration waits up to one more relist cycle after delivery resumes.
```

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,162 @@
...
+
+Scale-from-zero on backlog through the metric path requires the metric to exist while there are zero workers. It does not:
```
jaypipes:
```suggestion
HPA cannot scale your Worker Deployment from zero because the signal for scaling does not yet exist. The signal for scaling is the backlog metric for the task queue associated with the workers in the Worker Deployment. This metric will not exist until there is at least one worker polling the task queue.
```

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,162 @@
...
+
+1. Zero workers means no polls.
+2. No polls for >5 minutes means the task queue is unloaded from Temporal Cloud's memory.
+3. An unloaded queue emits no metric.
+4. Adapter discovery returns no series, or HPA queries return no rows.
+5. HPA cannot scale up because there's no signal to scale on.
```
jaypipes:
```suggestion
In addition to the "first worker start" problem, for customers using Temporal Cloud, if there are no polling workers for a task queue for more than 5 minutes, Temporal Cloud will unload the task queue from memory. Unloaded task queues do not emit metrics, and therefore the signal that HPA uses to scale up will not be present.
```

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,162 @@
...
+
+Submitting a workflow does load the task queue back into memory, but the metric still won't reach the HPA until the next OpenMetrics emission cycle (~1 minute) plus scrape and HPA poll. By the time the HPA reacts, you've already had ~1+ minute of unprovisioned work.
```
jaypipes:
```suggestion
Submitting a workflow does load the task queue back into memory, but the metric still won't reach the HPA until the next OpenMetrics emission cycle (~1 minute). By the time the HPA reacts, you've already had ~1+ minute of unprovisioned work.
```

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,162 @@
...
+
+KEDA's Temporal scaler calls `DescribeTaskQueue(stats=true)` (or `DescribeWorkerDeploymentVersion`), which loads the queue synchronously and returns the backlog directly. No metric pipeline involved. Scale-from-zero in seconds.
```
jaypipes:
```suggestion
## KEDA strengths

KEDA's Temporal scaler calls `DescribeTaskQueue(stats=true)` (or `DescribeWorkerDeploymentVersion`), which loads the queue synchronously and returns the backlog directly. This allows KEDA to scale Temporal workers from zero.
```

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,162 @@
...
+- **It hides a `sum(...)` that the adapter already does.** prometheus-adapter's `metricsQuery: sum(<<.Series>>{<<.LabelMatchers>>})` performs the same collapsing at query time. Pedagogically, "the adapter does the sum" is cleaner than "a recording rule sums first, then the adapter sums again."
+- **It does not solve series-level silence.** When the source goes silent (task queue unloaded), the rule output also goes silent eventually (once Prometheus's staleness lookback expires).
+
+What the recording rule *does* buy is registration stability after operational events: when the source series is sparse-by-timestamp, the rule produces a dense 10-second sample stream that lets the adapter discover with a tight `metricsRelistInterval`. If you find yourself fighting registration flicker on every adapter restart and would rather pay the cardinality cost than tune `metricsRelistInterval`, a recording rule is a reasonable choice. Otherwise, prefer the raw metric.
+
+In this demo we set `metricsRelistInterval: 5m` and consume the raw metric directly.
+
```
jaypipes, line 85:
```suggestion
## HPA strengths

Because HPA uses a single OpenMetrics scrape to gather all series for the namespace in a single HTTP request, the HPA approach scales independently of namespace count. The single HTTP request for OpenMetrics more efficient than KEDA's Temporal API-based approach, and will not run into Temporal API rate limiting problems (see section below on [KEDA limitations](#keda-limitations)).

HPA + prometheus adapter configured to look at both slot util and backlog provides fast scale-up via slot util and a backlog-driven backstop to prevent overly reactive replica count adjustment.
```

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,162 @@
...
+  expr: sum by (...) (temporal_cloud_v1_approximate_backlog_count)
+```
+
+The rule was originally added to work around a label-formatting issue in an older Temporal Cloud release. With native per-version labels (`temporal_worker_deployment_name`, `temporal_worker_build_id`) now opt-in, the rule no longer earns its keep:
+
+- **It doesn't reduce reactivity.** The HPA reactivity floor is the upstream OpenMetrics emission cadence (~60s), not anything the rule could fix.
+- **It duplicates the cardinality bill.** Per-`(task_queue, build_id)` labels are already opt-in at the OpenMetrics level *because* of cardinality. Adding a recording rule on top means storing the same high-cardinality series twice.
+- **It hides a `sum(...)` that the adapter already does.** prometheus-adapter's `metricsQuery: sum(<<.Series>>{<<.LabelMatchers>>})` performs the same collapsing at query time. Pedagogically, "the adapter does the sum" is cleaner than "a recording rule sums first, then the adapter sums again."
+- **It does not solve series-level silence.** When the source goes silent (task queue unloaded), the rule output also goes silent eventually (once Prometheus's staleness lookback expires).
```
jaypipes: Remove this. It's not necessary and likely will just confuse the reader.

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,162 @@
...
+
+This document describes practical reactivity and reliability tradeoffs when scaling Temporal workers per worker-deployment-version on Kubernetes, and recommends which tool fits which workload pattern.
+
+The `internal/demo/` example wires the HPA path described here. The KEDA path is mentioned for comparison and as a recommendation for workloads that cannot tolerate the HPA path's limits.
+
+## TL;DR — Pick by workload pattern
+
+| Workload pattern | Recommendation |
+|------------------|----------------|
+| Continuous traffic (task queue always loaded) | HPA + prometheus-adapter, scaling on slot utilization + backlog count |
```
jaypipes:
```suggestion
| Continuous traffic (task queue always loaded) | HPA |
```

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,162 @@
...
+| Idle periods >5 min between work; needs scale-from-zero | KEDA Temporal scaler |
```
jaypipes:
```suggestion
| Idle periods >5 min between work OR needs scale-from-zero | KEDA Temporal scaler |
```

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,162 @@
...
+| Required reactivity < ~60 s from first backlog | KEDA Temporal scaler |
+| Required reactivity ~90 s typical, tolerant of occasional multi-minute stalls | HPA + prometheus-adapter is fine |
```
jaypipes:
```suggestion
| Required reactivity ~90 s typical, tolerant of occasional multi-minute stalls | HPA + prometheus-adapter |
```

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,162 @@
...
+| Very large namespaces (N × M HPAs polling) | HPA + prometheus-adapter; KEDA hits the 50 RPS namespace rate limit |
```
jaypipes:
```suggestion
| 1000s of task queues and worker deployment versions  | HPA + prometheus-adapter |
```

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,162 @@
...
+```
+backlog appears at T0
+  └─ Temporal Cloud OpenMetrics emission cadence    +~60 s worst-case  (~1 sample/minute)
+       └─ Prometheus scrape interval                 +~10 s
+            └─ HPA poll interval                     +~15 s
+                 └─ scale-up stabilization window    +~your config
+                      └─ first replica added
+```
+
+**Typical end-to-end reactivity is ≈ 85 seconds + your stabilization window.** Empirically, sample age in Prometheus for a single series follows a sawtooth between 0 and 60 seconds (matching the gateway's ~1/min emission cadence). p50 sample age ≈ 30s, p95 ≈ 50s. The 60-second emission cadence is the inherent floor — smaller scrape intervals, tighter `metricsRelistInterval`, or recording rules cannot improve it because they all consume the same upstream cadence.
```
jaypipes: What is "your stabilization window"? That is not defined anywhere and this reads like Claude has seen the phrase "stabilization window" in other docs that it scraped from the Internet that described auto-scaling algorithms but doesn't actually understand what "stabilization window" means.

Also, "Typical end-to-end reactivity" doesn't make sense here and sounds like a term Claude either made up or has hallucinated-adopted from the term "end-to-end reactivity" [from frontend software development patterns](https://www.explainthis.io/en/swe/understanding-reactivity-from-frontend).

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,162 @@
...
+
+| Workload pattern | Recommendation |
+|------------------|----------------|
+| Continuous traffic (task queue always loaded) | HPA + prometheus-adapter, scaling on slot utilization + backlog count |
+| Idle periods >5 min between work; needs scale-from-zero | KEDA Temporal scaler |
+| Required reactivity < ~60 s from first backlog | KEDA Temporal scaler |
+| Required reactivity ~90 s typical, tolerant of occasional multi-minute stalls | HPA + prometheus-adapter is fine |
+| Very large namespaces (N × M HPAs polling) | HPA + prometheus-adapter; KEDA hits the 50 RPS namespace rate limit |
+
+## The reactivity model for HPA + prometheus-adapter
```
jaypipes:
```suggestion
## HPA scaling signal

This section describes the signal used by HPA + prometheus adapter to adjust the count of workers in a Kubernetes deployment managed by Temporal Worker Controller.

There are two metric data points that are scraped by HPA + prometheus adapter.

`temporal_cloud_v1_approximate_backlog_count` (or just "backlog") is a measurement of the number of pending tasks on a particular task queue that are waiting for a poller (a worker) to pull that task and process it.

`temporal_slot_utilization` (or just "slot util") is emitted directly by worker pods (no Temporal Cloud aggregation), scraped at the ServiceMonitor interval (~10–30 s), and reflects the current state of a particular worker. This metrics rises *before* backlog accumulates — slots saturate first, then queueing starts.
```

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,162 @@
...
+
+We have only directly characterized this pattern during one investigation session (seeing it twice in ~2 hours of close observation). Frequency in normal operation is not yet known and is open with Temporal's Observability team. If your workload cannot tolerate occasional multi-minute scaling pauses, prefer KEDA.
+
+This is also why `metricsRelistInterval: 5m` is the recommended setting: the discovery window must comfortably exceed the longest expected delay so the metric does not deregister, otherwise re-registration waits up to one more relist cycle after delivery resumes.
+
+### Slot utilization is a much faster leading signal
+
+`temporal_slot_utilization` is emitted directly by worker pods (no Temporal Cloud aggregation), scraped at the ServiceMonitor interval (~10–30 s), and reflects current state. It also rises *before* backlog accumulates — slots saturate first, then queueing starts. So a two-metric HPA with both slot util and backlog gives you fast scale-up via slot util and a backlog-driven backstop.
+
+The demo HPA uses both. For production scaling we recommend keeping both as well.
```
jaypipes: Recommend removing this. I've pulled some of the content into the proposed new `## HPA strengths` and `## HPA scaling signal` sections

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,162 @@
...
+
+The retrospective completeness is helpful for dashboards and post-hoc analysis, but it does **not** help an HPA, which queries the *latest available* value at decision time. During a delivery delay, the latest available sample is the one from before the delay started. The HPA sees real staleness even though the underlying record will eventually be filled in.
+
+We have only directly characterized this pattern during one investigation session (seeing it twice in ~2 hours of close observation). Frequency in normal operation is not yet known and is open with Temporal's Observability team. If your workload cannot tolerate occasional multi-minute scaling pauses, prefer KEDA.
+
+This is also why `metricsRelistInterval: 5m` is the recommended setting: the discovery window must comfortably exceed the longest expected delay so the metric does not deregister, otherwise re-registration waits up to one more relist cycle after delivery resumes.
```
jaypipes: Recommend removing this section. I've moved some of the content and reworded it in the `## HPA limitations` section below.

jaypipes (changes requested):
Thanks @carlydf , I've done a first go-around reviewing this documentation and adding (quite a few) suggested changes and removals to "de-Claude" some of it and make it (hopefully) a bit more readable for a general audience.

docs/scaling-recommendations.md
mnichols: I would recommend clarifying that this is talking about Temporal Cloud...OSS customers won't have this limitation OOTB. 
I would also recommend referencing https://docs.temporal.io/cloud/limits and if this limit is not explicitly there have it added there. Ideally, you'd have a special `Worker Versioning` section on that limits page so customers can just reference that directly

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,162 @@
...
+                      └─ first replica added
+```
+
+**Typical end-to-end reactivity is ≈ 85 seconds + your stabilization window.** Empirically, sample age in Prometheus for a single series follows a sawtooth between 0 and 60 seconds (matching the gateway's ~1/min emission cadence). p50 sample age ≈ 30s, p95 ≈ 50s. The 60-second emission cadence is the inherent floor — smaller scrape intervals, tighter `metricsRelistInterval`, or recording rules cannot improve it because they all consume the same upstream cadence.
+
+### Caveat: gateway delivery delay
+
+During our investigation we observed periods of several minutes during which Temporal Cloud's OpenMetrics endpoint returned the same embedded timestamps on repeated scrapes for *every* series across the account simultaneously — backlog series, action counts, error counts, every queue, every namespace, all showing identical staleness to the second (e.g. all ~30 visible series reading 239s old at once). The Prometheus scrape continued to succeed (`up{job="temporal_cloud"}` stayed 1, HTTP 200 responses) — the response body simply repeated already-known samples instead of advancing.
```
02strich: this reads odd given that the project is from the same people as Temporal Cloud - can this be reworded?

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,162 @@
...
+
+Once the delay resolved, the gateway delivered the missing samples with their original minute-aligned timestamps in a burst, so Prometheus's storage ends up with a complete 1/minute series in retrospect. We verified this directly: across a 3-hour window covering one such delay event, every gap between consecutive sample timestamps was exactly 60 seconds, no exceptions.
+
+The retrospective completeness is helpful for dashboards and post-hoc analysis, but it does **not** help an HPA, which queries the *latest available* value at decision time. During a delivery delay, the latest available sample is the one from before the delay started. The HPA sees real staleness even though the underlying record will eventually be filled in.
+
+We have only directly characterized this pattern during one investigation session (seeing it twice in ~2 hours of close observation). Frequency in normal operation is not yet known and is open with Temporal's Observability team. If your workload cannot tolerate occasional multi-minute scaling pauses, prefer KEDA.
```
02strich: this sounds like sharing internal sausage making - which as a customer I am not sure what to take away from

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,162 @@
...
+
+### Mode 2: series-level silence (common in low-traffic workloads)
+- Trigger: a task queue with no polls or new tasks for >5 minutes. Temporal unloads it from memory and stops emitting `temporal_cloud_v1_approximate_backlog_count` for that specific `(task_queue, build_id, ...)` labelset. Other queues' series continue to emit.
+- Symptom in HPA events: `no metrics returned from external metrics API`. The metric *name* is still registered; the HPA's specific label selector just matches zero rows now.
+- Recovery: traffic resumes → queue reloads → next emission cycle (~1 min) + 3-min aggregation lag → HPA can read value again.
+
+In a two-metric HPA configured with slot utilization, this is mostly fine: the HPA reports `ScalingActive=True` based on slot utilization while backlog is unavailable, and rejoins backlog scaling once it returns. We've confirmed this empirically in this demo cluster — the HPA continued scaling correctly on slot utilization through 1000+ backlog `FailedGetExternalMetric` events.
+
+## Why this demo does not use a backlog recording rule
```
02strich: which demo? "this" is an unclear reference

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,162 @@
...
+
+A prior version of this demo wrapped the raw Cloud series in a Prometheus recording rule:
```
02strich:
```suggestion
A prior version of this demo wrapped the raw Temporal Cloud series in a Prometheus recording rule:
```

docs/scaling-recommendations.md
02strich: Isn't the same problem also seen with OSS?

docs/scaling-recommendations.md
jaypipes: Right, which is why I recommended in my review that this entire section be removed :)

docs/scaling-recommendations.md
jaypipes: Ii recommended removing this entire section

docs/scaling-recommendations.md
jaypipes: not sure. @carlydf do you know the answer to this?

docs/scaling-recommendations.md
carlydf: OSS customers can increase the rate limit, but they still will experience the default rate limit unless they override it. If they do increase the rate limit, they will also need to eventually scale up the per-namespace worker, because the per-namespace worker handles `DescribeWorkerDeployment` requests.

docs/scaling-recommendations.md
carlydf: Sorry about any Claude-isms, the reason this is in Draft mode is because I hadn't done a full pass over it yet. I care deeply about not putting PRs in front of people that I haven't reviewed myself and don't necessarily endorse any of this until it's out of draft mode.

Claude probably came up with this after seeing a bunch of grafana screenshots that I sent it.

docs/scaling-recommendations.md
carlydf: yes, same problem for OSS and Cloud here, Claude hallucinating. (PR in draft mode because I had not fully reviewed yet, didn't intend to waste your energy on review until it was ready!)

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,162 @@
...
+4. Adapter discovery returns no series, or HPA queries return no rows.
+5. HPA cannot scale up because there's no signal to scale on.
+
+Submitting a workflow does load the task queue back into memory, but the metric still won't reach the HPA until the next OpenMetrics emission cycle (~1 minute) plus scrape and HPA poll. By the time the HPA reacts, you've already had ~1+ minute of unprovisioned work.
+
+KEDA's Temporal scaler calls `DescribeTaskQueue(stats=true)` (or `DescribeWorkerDeploymentVersion`), which loads the queue synchronously and returns the backlog directly. No metric pipeline involved. Scale-from-zero in seconds.
+
+## When KEDA hits its own limits
+
+KEDA bypasses the metric pipeline but uses Temporal API calls, which are subject to a per-namespace rate limit:
```
carlydf: Other KEDA limitation (for now) https://github.com/temporalio/temporal-worker-controller/issues/355

KEDA also does not actually work with TWC until https://github.com/temporalio/temporal-worker-controller/issues/286 is closed. Luckily we have an open community PR #351 that will add support for the KEDA temporal trigger, which I think can be merged soon. I just reviewed it.

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,172 @@
...
+
+Temporal Cloud's OpenMetrics endpoint may sometimes return the same embedded timestamps on repeated scrapes for each series across the account simultaneously — backlog series, action counts, error counts, every queue, every namespace. This delay in returning fresh metrics data can impact the speed to which HPA + prometheus adapter scales out or in the replica count for a worker deployment version. This means that HPA + prometheus adapter may not be a good solution if your workload cannot tolerate occasional multi-minute scaling pauses.
+
+> **Note**: This is why `metricsRelistInterval: 5m` is the recommended setting: the discovery window must comfortably exceed the longest expected delay so the metric does not deregister, otherwise re-registration waits up to one more relist cycle after delivery resumes.
+
+HPA cannot scale your Worker Deployment from zero because the signal for scaling does not yet exist. The signal for scaling is the backlog metric for the task queue associated with the workers in the Worker Deployment. This metric will not exist until there is at least one worker polling the task queue.
```
Shivs11, line 100: i took about a second to understand what this really meant - at first, I thought this meant that there won't be a backlog metric emitted if you don't have workers running at all (which is not true since you do have this metric being emitted for the unversioned world without workers being present)

I know you have clearly mentioned versions in the preamble here, but do you think we can be extra clear and mention the backlog count per version is not emitted without a worker being present since that is what creates a version in temporal?

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,172 @@
...
+```
+[tc-openmetrics]: https://docs.temporal.io/cloud/metrics/openmetrics
+
+## HPA strengths
+
+Because HPA uses a single OpenMetrics scrape to gather all series for the namespace in a single HTTP request, the HPA approach scales independently of namespace count. The single HTTP request for OpenMetrics more efficient than KEDA's Temporal API-based approach, and will not run into Temporal API rate limiting problems (see section below on [KEDA limitations](#keda-limitations)).
+
+HPA + prometheus adapter configured to look at both slot util and backlog provides fast scale-up via slot util and a backlog-driven backstop to prevent overly reactive replica count adjustment.
+
+## HPA limitations
```
Shivs11, line 90: this is great! was wondering - should we also mention that our OM endpoint inherently has a 3 minute time lag, which would mean that at time=0, we are seeing a one minute aggregate of time.now() - 3 minutes

Shivs11 (approved):
a couple of nits -- looks g to me otherwise

docs/scaling-recommendations.md
jaypipes: @Shivs11 do we really need to care about the unversioned world in TWC? I mean, TWC doesn't deal with anything that isn't versioned since it automatically creates WorkerDeploymentVersions for new worker image tags...

docs/scaling-recommendations.md
jaypipes: @Shivs11 I think you meant "initially" instead of "inherently"? It's true that there is a 3-minute potential *initial lag* ([ref](https://docs.temporal.io/cloud/metrics/openmetrics#overview)) but this is only relevant for brand new task queues or task queues that have been deregistered from the OM endpoint. This is why there is a note in the doc:

> **Note**: This is why `metricsRelistInterval: 5m` is the recommended setting: the discovery window must comfortably exceed the longest expected delay so the metric does not deregister, otherwise re-registration waits up to one more relist cycle after delivery resumes.

I've added a warning about the initial 3 minute potential export lag.

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,176 @@
...
+
+## HPA scaling signal
+
+This section describes the signal used by HPA + prometheus adapter to adjust the count of workers in a Kubernetes deployment managed by Temporal Worker Controller.
+
+There are two metric data points that are scraped by HPA + prometheus adapter.
+
+`temporal_cloud_v1_approximate_backlog_count` (or just "backlog") is a measurement of the number of pending tasks on a particular task queue that are waiting for a poller (a worker) to pull that task and process it. This is a metric provided by [Temporal Cloud's OpenMetrics aggregation service][tc-openmetrics].
+
+`temporal_slot_utilization` (or just "slot util") is emitted directly by Workers (no Temporal Cloud aggregation), scraped at the Prometheus `ServiceMonitor` interval (~10–30 s), and reflects the current state of a particular Worker. This metric rises *before* backlog accumulates. In other words, slots on the Worker saturate first, then queueing starts.
```
eniko-dif, line 57: Is there a documentation about worker slot setup? In our Datadog environment we ended up setting the default for slot utilization to `1000` and we noticed that we never actually get anywhere close in using them up. It might be worth mentioning for the customers that if this is not properly adjusted for the worker's resources the scaling might not work as expected.

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,176 @@
...
+        metric:
+          name: temporal_cloud_v1_approximate_backlog_count
+          selector:
+            matchLabels:
+              temporal_task_queue: "default_helloworld"
+              task_type: "Activity"
+        target:
+          type: AverageValue
+          averageValue: "1"
+  behavior:
```
eniko-dif, line 171: Isn't it an issue that HPA would scale up if the system is well-sized? (meaning that the backlog count is not building up, but the workers have a high slot utilization)

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,176 @@
...
+```
+backlog appears at T0
+  └─ Temporal Cloud OpenMetrics emission cadence     + ~60s worst-case  (~1 sample/minute)
+       └─ Prometheus scrape interval                 + ~10s
+            └─ HPA poll interval                     + ~15s
+                 └─ scale-up stabilization window    + taken from HPA configuration
+                      └─ first replica added
+```
+[tc-openmetrics]: https://docs.temporal.io/cloud/metrics/openmetrics
+
```
eniko-dif, line 83: nit: i'd mention here that an example configuration is available a bit below (maybe with a link to jump through keda if the user is not interested)

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,176 @@
...
+  metrics:
+    - type: External
+      external:
+        metric:
+          name: temporal_slot_utilization
+          selector:
+            matchLabels:
+              worker_type: "ActivityWorker"
+        target:
+          type: Value
```
eniko-dif: nit: I'd add the comment from the `wrt-hpa-backlog` example why this is a Value, because upon first look I was confused.

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,176 @@
...
+      external:
+        metric:
+          name: temporal_cloud_v1_approximate_backlog_count
+          selector:
+            matchLabels:
+              temporal_task_queue: "default_helloworld"
+              task_type: "Activity"
+        target:
+          type: AverageValue
+          averageValue: "1"
```
eniko-dif, line 170: Isn't 1 a bit too low?

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,176 @@
...
+  behavior:
+    scaleUp:
+      stabilizationWindowSeconds: 30
+      policies:
+        - type: Percent
```
eniko-dif, line 175: I'd personally put quicker scale up values than scale down.

docs/scaling-recommendations.md
jaypipes: @eniko-dif I've added a link to this doc: https://docs.temporal.io/develop/worker-performance

Unfortunately, I don't think there's going to be a one-size-fits-all recommendation for slot utilization values because this is going to be dependent on the rate of the user's workflows (and their composing activities) being executed. Obviously that's going to be highly user-specific. I've added a link to the docs about [choosing an appropriate slot supplier](https://docs.temporal.io/develop/worker-performance#choosing-slot-supplier-types) and highlighted this callout from that doc:

> Scenarios with tasks that have variable, or very high, per-task resource needs should rely on fixed-size suppliers and manual tuning rather than resource-based suppliers.

docs/scaling-recommendations.md
jaypipes: I moved the recommended config up.

docs/scaling-recommendations.md
jaypipes: Yup. Noted. https://github.com/temporalio/temporal-worker-controller/pull/351 looks almost ready to merge.

docs/scaling-recommendations.md
jaypipes: I took this directly from the existing examples in the demo/ directory :) But I agree with you that it should more correctly be `AverageValue` and a value of `"0.75"` not `"750m"`. Updated.

docs/scaling-recommendations.md
jaypipes: Not necessarily... depends on the workload... for a high-request-rate workflow/taskqueue, with workers struggling to process the incoming requests, the backlog count might be much higher than 1, but it all depends on the workload...

docs/scaling-recommendations.md
jaypipes: If this happens, the user would want to adjust the AverageValue target value for temporal_slot_utilization so that it would not trigger a scale up, no? Or adjust the stabilization window...

docs/scaling-recommendations.md
jaypipes: Isn't that what this example shows?

jaypipes (approved):

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,190 @@
...
+The `internal/demo/` example wires the HPA path described here. The KEDA path is mentioned for comparison and as a recommendation for workloads that cannot tolerate the HPA path's limits.
+
+## TL;DR
+
+We recommend choosing a scaler approach that aligns with the workload pattern your application exhibits.
+
+| Workload pattern | Recommendation |
+|------------------|----------------|
+| Continuous traffic (task queue always loaded) | HPA |
+| Idle periods >5 min between work OR needs scale-from-zero | KEDA Temporal scaler |
```
carlydf: I think this is AND and not OR.

If there are idle periods >5 min between work, but pod count is >=1, then the pollers in the one pod will keep the task queues loaded.

Basically:

Task queue always loaded (continuous task traffic OR never scale-from-zero) --> HPA

Task queue may be unloaded (idle periods >5 min between tasks AND needs scale-from-zero) --> KEDA Temporal scaler

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,190 @@
...
+
+The `internal/demo/` example wires the HPA path described here. The KEDA path is mentioned for comparison and as a recommendation for workloads that cannot tolerate the HPA path's limits.
+
+## TL;DR
+
+We recommend choosing a scaler approach that aligns with the workload pattern your application exhibits.
+
+| Workload pattern | Recommendation |
+|------------------|----------------|
+| Continuous traffic (task queue always loaded) | HPA |
```
carlydf: nit: is there a difference between `HPA + prometheus adapter` in lower rows and just `HPA` here? Can we just say `HPA + Temporal Cloud Metrics + metrics adapter` in all the HPA rows? For example, Datadog metrics can be passed to HPA with per-version labels as described [here](https://docs.datadoghq.com/containers/guide/cluster_agent_autoscaling_metrics/?tab=datadogoperator#autoscaling-without-datadogmetric-queries). So someone could use the Temporal OpenMetrics endpoint to get backlog count into their Datadog system, and then use Datadog HPA adapter to use HPA + Datadog metrics. We just have not worked that example because Datadog is paid only.

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,190 @@
...
+| Idle periods >5 min between work OR needs scale-from-zero | KEDA Temporal scaler |
+| Required reactivity < ~60 s from first backlog | KEDA Temporal scaler |
+| Required reactivity ~90 s typical, tolerant of occasional multi-minute stalls | HPA + prometheus-adapter |
+| 1000s of task queues and worker deployment versions  | HPA + prometheus-adapter |
+
+## HPA scaling signal
+
+This section describes the signal used by HPA + prometheus adapter to adjust the count of workers in a Kubernetes deployment managed by Temporal Worker Controller.
```
carlydf, line 51: Maybe here is a place to say

> HPA + metrics adapter.... We will discuss the Prometheus Metrics Adapter in depth because it is free to install so we have tested it with the Worker Controller end to end, but the Temporal Cloud Metrics -> HPA method is expected to work with other metrics providers as long as (1) the metrics provider can ingest Temporal Cloud Metrics from our [OpenMetrics endpoint](https://docs.temporal.io/cloud/metrics/openmetrics), and (2) the metrics provider has a Kubernetes adapter to pipe those metrics to HPA and use native HPA `matchLabels` format to create a per-version metrics query. Using a different aggregation layer could incur additional delays, depending on the configuration of that layer. We will discuss the Prometheus pathway in detail.
>
> For example:
> - Prometheus
>   - [Temporal Cloud Metrics integration](https://docs.temporal.io/cloud/metrics/openmetrics/metrics-integrations#prometheus-grafana) (Temporal Cloud -> Prometheus)
>   - [HPA adapter](https://github.com/kubernetes-sigs/prometheus-adapter) (Prometheus -> HPA)
>
> - Datadog
>   - [Temporal Cloud Metrics integration](https://docs.temporal.io/cloud/metrics/openmetrics/metrics-integrations#datadog) (Temporal Cloud -> DataDog)
>   - [HPA adapter](https://docs.datadoghq.com/containers/guide/cluster_agent_autoscaling_metrics/?tab=datadogoperator#autoscaling-without-datadogmetric-queries) (DataDog -> HPA)
>
> - New Relic
>   - [Temporal Cloud Metrics integration](https://docs.temporal.io/cloud/metrics/openmetrics/metrics-integrations#new-relic) (Temporal Cloud -> New Relic)
>   - [HPA adapter](https://docs.newrelic.com/docs/kubernetes-pixie/kubernetes-integration/advanced-configuration/newrelic-metrics-adapter/) (New Relic -> HPA)
>
> - OpenTelemetry Collector -> system of your choosing
>   - [Temporal Cloud Metrics integration](https://docs.temporal.io/cloud/metrics/openmetrics/metrics-integrations#opentelemetry-collector-configuration) (Temporal Cloud -> OpenTelemetry)
>   - [AWS CloudWatch OpenTelemetry integration](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Agent-OpenTelemetry-metrics.html) (OpenTelemetry -> CloudWatch)
>   - [HPA adapter for AWS CloudWatch](https://aws.amazon.com/blogs/compute/scaling-kubernetes-deployments-with-amazon-cloudwatch-metrics/) (CloudWatch -> HPA)

docs/scaling-recommendations.md
carlydf: Or, maybe we say **HPA + prometheus adapter*** in all those rows, and let the * lead to the paragraph I propose in my other comment: "we tested and are discussing prometheus in detail, but there are other ways to pipe Cloud Metrics -> HPA that have similar caveats with slightly different timing depending on configuration"

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,190 @@
...
+
+## HPA limitations
+
+This section describes two known limitations for HPA + prometheus adapter.
+
+Temporal Cloud's OpenMetrics endpoint may sometimes return the same embedded timestamps on repeated scrapes for each series across the account simultaneously — backlog series, action counts, error counts, every queue, every namespace. This delay in returning fresh metrics data can impact the speed to which HPA + prometheus adapter scales out or in the replica count for a worker deployment version. This means that HPA + prometheus adapter may not be a good solution if your workload cannot tolerate occasional multi-minute scaling pauses.
+
+> **Warning**: There is an [up to 3 minute potential delay][om-delay] before exported metrics are available in the Temporal Cloud OpenMetrics endpoint for new task queues.
+
+> **Note**: This is why `metricsRelistInterval: 5m` is the recommended setting: the discovery window must comfortably exceed the longest expected delay so the metric does not deregister, otherwise re-registration waits up to one more relist cycle after delivery resumes.
```
carlydf, line 98: @dustin-temporal is this correct? / is this the correct way to say it? I know we've discussed in slack, but just want to triple check here also.

docs/scaling-recommendations.md
carlydf: FWIW, I think the "first ever" scale-from-zero is almost not worth mentioning in TWC-land, because the TWC always creates a new worker version with at least 1 replica, thereby creating the queue and metric for that version. I think it could be better to just focus on the queue idleness problem (which I feel you've explained well).

docs/scaling-recommendations.md
carlydf: I think a lot of use cases target zero backlog, so scaling up as soon as any backlog is detected makes sense (especially given detection delays)

internal/demo/k8s/prometheus-stack-values.yaml
```diff
@@ -11,7 +11,9 @@
 #   1. ServiceMonitor — scrapes worker pod metrics (slot gauges) from port 9090
 #   2. Temporal Cloud scrape config — scrapes temporal_cloud_v1_approximate_backlog_count
 #      (Phase 2 only; requires a Temporal Cloud metrics API key)
-#   3. Recording rules — slot utilization ratio (Phase 1) and backlog count by version (Phase 2)
+#   3. Recording rule — slot utilization ratio (Phase 1 only). The backlog count
+#      is consumed directly from the raw Cloud series via prometheus-adapter; see
+#      docs/scaling-recommendations.md for the reasoning.
```
carlydf: should delete "see docs/scaling-recommendations.md for the reasoning."

internal/demo/README.md
```diff
@@ -268,7 +268,7 @@ You'll also need to [opt-in](https://docs.temporal.io/cloud/metrics/openmetrics/
...
 
-> **Note:** This demo ships a Prometheus recording rule that renames `temporal_cloud_v1_approximate_backlog_count` to `temporal_approximate_backlog_count` and reduces it to the labels the HPA cares about. In principle the HPA can consume the raw Cloud metric directly (set `namespaced: false` on the prometheus-adapter rule so it doesn't auto-inject a `namespace` label filter), but this demo uses the recording rule as a known-working path.
+> **Picking a scaling tool for your workload:** This demo uses the HPA + prometheus-adapter path. It works well for continuously-loaded task queues and has a typical end-to-end reactivity of ~85 seconds (dominated by Temporal Cloud's ~1/minute OpenMetrics emission cadence). It cannot do scale-from-zero. For sub-60s reactivity or scale-from-zero, use the KEDA Temporal scaler. See [docs/scaling-recommendations.md](../../docs/scaling-recommendations.md) for the full reactivity model, when to pick which, and a caveat about an account-wide OpenMetrics delivery-delay pattern we observed during testing (retrospectively backfilled, but real for live HPA queries).
```
carlydf: Probably should delete "and a caveat about an account-wide OpenMetrics delivery-delay pattern we observed during testing (retrospectively backfilled, but real for live HPA queries)."

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,190 @@
...
+
+[tc-openmetrics]: https://docs.temporal.io/cloud/metrics/openmetrics
+[worker-perf]: https://docs.temporal.io/develop/worker-performance
+[slot-alloc-strat]: https://docs.temporal.io/develop/worker-performance#choosing-slot-supplier-types
+
+## HPA strengths
+
+Because HPA uses a single OpenMetrics scrape to gather all series for the namespace in a single HTTP request, the HPA approach scales independently of namespace count. The single HTTP request for OpenMetrics more efficient than KEDA's Temporal API-based approach, and will not run into Temporal API rate limiting problems (see section below on [KEDA limitations](#keda-limitations)).
+
+HPA + prometheus adapter configured to look at both slot util and backlog provides fast scale-up via slot util and a backlog-driven backstop to prevent overly reactive replica count adjustment.
```
carlydf: nit: It feels like this sentence is in "note taking" style. Can we say "utilization" instead of "util"? And "can be configured" instead of "configured"? ie:
> HPA + prometheus adapter can be configured to combine slot utilization (or any other other version-labelled metrics) with backlog count. Slot utilization can be used to prevent overly reactive scale-down when backlog is zero but the workers are well-utilized and replica count is right-sized for the workload.

carlydf (commented):
So close! Thank you for all your hard work on this @jaypipes. Let me know what you think of my suggestions :)

docs/scaling-recommendations.md
```diff
@@ -0,0 +1,190 @@
...
+        value: 10
+        periodSeconds: 10
+      selectPolicy: Max
+```
+
+## KEDA strengths
+
+KEDA's Temporal scaler calls `DescribeTaskQueue(stats=true)` (or `DescribeWorkerDeploymentVersion`), which loads the queue synchronously and returns the backlog directly. This allows KEDA to scale Temporal workers from zero.
+
+## KEDA limitations
```
carlydf, line 193: also, IMO we should mention here the fact that KEDA Temporal Scaler uses backlog count _only_, and TWC does not currently template per-version metrics queries in other KEDA scalers.

docs/scaling-recommendations.md
jaypipes: Updated all lines to "HPA + prometheus-adapter" and added a note with your text.

docs/scaling-recommendations.md
jaypipes: ack. changed to AND.

docs/scaling-recommendations.md
jaypipes: Added with slight rewording.

docs/scaling-recommendations.md
jaypipes: done.

docs/scaling-recommendations.md
jaypipes: added.

internal/demo/k8s/prometheus-stack-values.yaml
jaypipes: done.

internal/demo/README.md
jaypipes: updated.

carlydf (commented):
approved! (I can't actually approve because this is technically my PR)

jlegrone (approved):

PR #329, opened 2026-05-21
fix: evict cached Temporal client on failures that indicate possible bad-client

### LINKED ISSUE #328: WorkerDeployment deletion finalizer infinite-loops on transport-level Temporal SDK errors (cached client never evicted)
### Summary

When a `WorkerDeployment` is deleted, `handleDeletion` (`internal/controller/worker_controller.go`) reuses a cached Temporal SDK client from `TemporalClientPool` to drive cleanup. If the cached client is broken at the **transport layer** (e.g. every gRPC call returns `context.DeadlineExceeded`), every requeue of the deletion finalizer re-uses the same poisoned client and the cleanup loop never makes progress. The only way to recover is to restart the controller pod — which drops the in-memory pool.

The main `Reconcile` path already evicts on `serviceerror.PermissionDenied` / `Unauthenticated` (via `isAccessDeniedErr` at `worker_controller.go:324`), but `handleDeletion` has no eviction at all, and transport-layer errors don't surface as access-denied.

### Environment

- Controller image: `ghcr.io/temporalio/temporal-worker-controller:main-…` (in sync with upstream `main` `f9d01cd`)
- Temporal: Temporal Cloud (API-key auth)
- Kubernetes: EKS 1.30
- CRD revision: post-v1.7 rename (`WorkerDeployment` / `Connection`)
- Helm-rendered Secret materialised by ExternalSecrets-operator

### What happened (sanitised production timeline)

Single review env, one `WorkerDeployment` named `example-nestjs-temporal-worker` in namespace `review-review-dev-ylmdqj`:

| Time (UTC) | Pattern | Count |
| --- | --- | --- |
| `11:08` | `Connection.temporal.io "temporal-cloud" not found` | 6 |
| `11:08` | `Reconciler error` (generic event-emit follow-up) | 18 |
| `11:08` | Helm finishes; Secret + Connection + WD all Healthy |  |
| `11:09–11:15` | `WorkerDeployment` is deleted (chart re-render) → finalizer starts |  |
| `11:15–11:54` | `failed to clean up Temporal server-side deployment data, will retry` — underlying error: `unable to describe worker deployment: context deadline exceeded` | ~240 |
| `11:54` | Operator restarts `temporal-worker-controller-manager` pod |  |
| `11:54` | Finalizer succeeds on next reconcile (fresh client dial) |  |

So the failure spans ~40 minutes and 240+ retries of `handleDeletion` with the exact same error, each one re-using the cached SDK client that was poisoned during the brief window when the `Connection` was being created but the API-key Secret was not yet readable.

### Root cause (code refs)

[`internal/controller/worker_controller.go#L544-L600`](https://github.com/temporalio/temporal-worker-controller/blob/f9d01cd/internal/controller/worker_controller.go#L544-L600):

```go
temporalClient, ok := r.TemporalClientPool.GetSDKClient(clientpool.ClientPoolKey{
    HostPort:   temporalConnection.Spec.HostPort,
    Namespace:  workerDeploy.Spec.WorkerOptions.TemporalNamespace,
    SecretName: secretName,
    AuthMode:   authMode,
})
// ... fallback dial path, no eviction on subsequent failures ...

resp, err := deploymentHandler.Describe(ctx, sdkclient.WorkerDeploymentDescribeOptions{})
if err != nil {
    if errors.As(err, &notFound) {
        l.Info("Worker Deployment not found on Temporal server, nothing to clean up")
        return nil
    }
    return fmt.Errorf("unable to describe worker deployment: %w", err)
}
```

Compare to the main `Reconcile` path ([`worker_controller.go#L322-L340`](https://github.com/temporalio/temporal-worker-controller/blob/f9d01cd/internal/controller/worker_controller.go#L322-L340)):

```go
if err != nil {
    if isAccessDeniedErr(err) {
        r.TemporalClientPool.EvictClient(clientPoolKey)
    }
    // ... rate-limit handling, status update ...
}
```

`isAccessDeniedErr` only matches `serviceerror.PermissionDenied` and `serviceerror.Unauthenticated` ([`worker_controller.go#L923-L928`](https://github.com/temporalio/temporal-worker-controller/blob/f9d01cd/internal/controller/worker_controller.go#L923-L928)), so a `context.DeadlineExceeded` against a broken client never triggers eviction even in the main reconcile path.

### Repro (synthetic)

1. Pre-populate the `TemporalClientPool` with a stub SDK client whose `WorkerDeploymentClient().GetHandle(name).Describe(...)` returns `context.DeadlineExceeded`.
2. Create a `WorkerDeployment` + `Connection` referencing the same `(HostPort, TemporalNamespace, SecretName, AuthMode)`.
3. Call `handleDeletion` directly.
4. Observe the error propagated and assert the cached client is **still** in the pool — i.e. the next reconcile will re-use the broken client and re-fail.

A passing unit test for the fix is included in the PR linked below; it inverts step 4 to assert the broken client is evicted and a healthy client is retained on `NotFound`.

### Proposed fix (minimal)

Add a `defer` in `handleDeletion` (immediately after the SDK client is acquired) that evicts the cached client whenever `handleDeletion` returns a non-nil error. `EvictClient` is documented as safe-when-missing, so the defer is also harmless on the dial-but-don't-cache path. `NotFound` on `Describe` is intentionally treated as success and returns `nil`, so the defer skips eviction there and the healthy client stays in the pool.

PR: <will be linked once filed>

### Alternative considered (broader)

Track consecutive Temporal-server failures per `ClientPoolKey` in the main `Reconcile` path and evict after N failures (or wrap every `temporal.GetWorkerDeploymentState` / `Describe` call in an eviction-on-transport-error helper). Rejected for the initial PR because:

- It adds state and changes hot-path behaviour for *every* reconcile.
- The deletion-cleanup path is the one we observed wedged in production; the main reconcile has periodic requeues and the `connectionFinalizerExists` path already self-recovers on `Connection.HostPort` changes.
- The minimal fix in `handleDeletion` is enough to close the recovery gap; a broader change can land separately if maintainers prefer.

Happy to expand the PR scope if reviewers prefer the broader direction.

rupesh-parab-one-app:
## What

Cached Temporal SDK clients are now evicted consistently when the controller observes failures that indicate the cached client may no longer be usable.

This expands the original deletion-cleanup-only fix to cover both places that reuse a cached client:

- the main `Reconcile` path after `GetWorkerDeploymentState` / `DescribeWorkerDeployment`
- `handleDeletion` while cleaning up Temporal server-side Worker Deployment data

The shared `shouldEvictClient(err)` predicate keeps the existing auth behavior and adds bounded recovery for failures that could indicate connectivity issues:

- `serviceerror.PermissionDenied`
- `Unauthenticated`
- `context.DeadlineExceeded`
- `serviceerror.Unavailable`

It intentionally does **not** evict on broader server/application responses such as `ResourceExhausted`, `context.Canceled`, or `NotFound`.

## Why

In #328 we observed the controller repeatedly reusing the same cached SDK client after `DescribeWorkerDeployment` returned `context.DeadlineExceeded`. The controller did not recover until the manager pod restarted and dropped the in-memory pool.

The earlier narrow version of this PR only evicted in `handleDeletion`, but reviewers correctly pointed out that the main `Reconcile` path has the same shape: get a cached client, call `Describe`, then requeue on transport failure without evicting. This PR now closes that recovery gap in both paths.

## Changes

- `internal/controller/worker_controller.go`
  - adds `shouldEvictClient(err)` next to `isAccessDeniedErr`
  - replaces main Reconcile `isAccessDeniedErr` eviction checks with `shouldEvictClient`
  - updates `handleDeletion` to use the same predicate instead of evicting on every non-nil return
- `internal/controller/reconciler_events_test.go`
  - adds `TestShouldEvictClient` to lock down included/excluded error classes
  - adds `TestReconcile_EvictsCachedClientOnTransportFailure`
  - keeps the deletion cleanup regression test and its NotFound-retains-client case
  - keeps the no-op `Close()` on `stubTemporalClient` so `EvictClient` can close test clients safely

## Test plan

- [x] `go test ./internal/controller -run 'Test(ShouldEvictClient|Reconcile_EvictsCachedClientOnTransportFailure|HandleDeletion_EvictsCachedClientOnTemporalFailure|Reconcile_DescribeWorkerDeploymentNotFound)' -count=1`
- [x] `go test ./internal/controller -count=1`

Closes #328.

### DIFF
--- a/internal/controller/worker_controller.go
+++ b/internal/controller/worker_controller.go
@@ -316,7 +316,7 @@ func (r *WorkerDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		getControllerIdentity(),
 	)
 	if err != nil {
-		if isAccessDeniedErr(err) {
+		if shouldEvictClient(err) {
 			r.TemporalClientPool.EvictClient(clientPoolKey)
 		}
 		var rateLimitErr *serviceerror.ResourceExhausted
@@ -362,7 +362,7 @@ func (r *WorkerDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
 
 	// Execute the plan, handling any errors
 	if err := r.executePlan(ctx, l, &workerDeploy, temporalClient, plan); err != nil {
-		if isAccessDeniedErr(err) {
+		if shouldEvictClient(err) {
 			r.TemporalClientPool.EvictClient(clientPoolKey)
 		}
 		r.recordWarningAndSetBlocked(ctx, &workerDeploy,
@@ -540,7 +540,7 @@ func (r *WorkerDeploymentReconciler) handleDeletion(
 	ctx context.Context,
 	l logr.Logger,
 	workerDeploy *temporaliov1alpha1.WorkerDeployment,
-) error {
+) (retErr error) {
 	// Resolve Connection.
 	// The Connection is guaranteed to exist because we hold a finalizer on it
 	// that prevents deletion while any WD references it.
@@ -563,13 +563,15 @@ func (r *WorkerDeploymentReconciler) handleDeletion(
 	authMode := connection.Spec.AuthMode()
 	secretName := connection.Spec.SecretName()
 
-	temporalClient, ok := r.TemporalClientPool.GetSDKClient(clientpool.ClientPoolKey{
+	clientPoolKey := clientpool.ClientPoolKey{
 		HostPort:      connection.Spec.HostPort,
 		TLSServerName: connection.Spec.TLSServerName(),
 		Namespace:     workerDeploy.Spec.WorkerOptions.TemporalNamespace,
 		SecretName:    secretName,
 		AuthMode:      authMode,
-	})
+	}
+
+	temporalClient, ok := r.TemporalClientPool.GetSDKClient(clientPoolKey)
 	if !ok {
 		clientOpts, key, clientAuth, err := r.TemporalClientPool.ParseClientSecret(ctx, secretName, authMode, clientpool.NewClientOptions{
 			K8sNamespace:      workerDeploy.Namespace,
@@ -586,6 +588,15 @@ func (r *WorkerDeploymentReconciler) handleDeletion(
 		}
 		temporalClient = c
 	}
+
+	// Evict cached SDK clients on failures that indicate the cached client may
+	// no longer be usable. Domain responses such as NotFound are handled as
+	// success below and should not churn otherwise healthy clients.
+	defer func() {
+		if shouldEvictClient(retErr) {
+			r.TemporalClientPool.EvictClient(clientPoolKey)
+		}
+	}()
 
 	workerDeploymentName := k8s.ComputeWorkerDeploymentName(workerDeploy)
 	deploymentHandler := temporalClient.WorkerDeploymentClient().GetHandle(workerDeploymentName)
@@ -931,4 +942,18 @@ func isAccessDeniedErr(err error) bool {
 		return true
 	}
 	return grpcstatus.Code(err) == codes.Unauthenticated
+}
+
+func shouldEvictClient(err error) bool {
+	if err == nil {
+		return false
+	}
+	if isAccessDeniedErr(err) {
+		return true
+	}
+	if errors.Is(err, context.DeadlineExceeded) {
+		return true
+	}
+	var unavailable *serviceerror.Unavailable
+	return errors.As(err, &unavailable)
 }
--- a/internal/controller/worker_controller.go
+++ b/internal/controller/worker_controller.go
@@ -318,7 +318,7 @@ func (r *WorkerDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		getControllerIdentity(),
 	)
 	if err != nil {
-		if isAccessDeniedErr(err) {
+		if shouldEvictClient(err) {
 			r.TemporalClientPool.EvictClient(clientPoolKey)
 		}
 		var rateLimitErr *serviceerror.ResourceExhausted
@@ -364,7 +364,7 @@ func (r *WorkerDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
 
 	// Execute the plan, handling any errors
 	if err := r.executePlan(ctx, l, &workerDeploy, temporalClient, plan); err != nil {
-		if isAccessDeniedErr(err) {
+		if shouldEvictClient(err) {
 			r.TemporalClientPool.EvictClient(clientPoolKey)
 		}
 		var rateLimitErr *serviceerror.ResourceExhausted
@@ -551,7 +551,7 @@ func (r *WorkerDeploymentReconciler) handleDeletion(
 	ctx context.Context,
 	l logr.Logger,
 	workerDeploy *temporaliov1alpha1.WorkerDeployment,
-) error {
+) (retErr error) {
 	// Resolve Connection.
 	// The Connection is guaranteed to exist because we hold a finalizer on it
 	// that prevents deletion while any WD references it.
@@ -574,14 +574,16 @@ func (r *WorkerDeploymentReconciler) handleDeletion(
 	authMode := connection.Spec.AuthMode()
 	secretName := connection.Spec.SecretName()
 
-	temporalClient, ok := r.TemporalClientPool.GetSDKClient(clientpool.ClientPoolKey{
+	clientPoolKey := clientpool.ClientPoolKey{
 		HostPort:            connection.Spec.HostPort,
 		TLSServerName:       connection.Spec.TLSServerName(),
 		Namespace:           workerDeploy.Spec.WorkerOptions.TemporalNamespace,
 		SecretName:          secretName,
 		TLSCACertSecretName: connection.Spec.TLSCACertSecretName(),
 		AuthMode:            authMode,
-	})
+	}
+
+	temporalClient, ok := r.TemporalClientPool.GetSDKClient(clientPoolKey)
 	if !ok {
 		clientOpts, key, clientAuth, err := r.TemporalClientPool.ParseClientSecret(ctx, secretName, authMode, clientpool.NewClientOptions{
 			K8sNamespace:      workerDeploy.Namespace,
@@ -598,6 +600,15 @@ func (r *WorkerDeploymentReconciler) handleDeletion(
 		}
 		temporalClient = c
 	}
+
+	// Evict cached SDK clients on failures that indicate the cached client may
+	// no longer be usable. Domain responses such as NotFound are handled as
+	// success below and should not churn otherwise healthy clients.
+	defer func() {
+		if shouldEvictClient(retErr) {
+			r.TemporalClientPool.EvictClient(clientPoolKey)
+		}
+	}()
 
 	workerDeploymentName := k8s.ComputeWorkerDeploymentName(workerDeploy)
 	deploymentHandler := temporalClient.WorkerDeploymentClient().GetHandle(workerDeploymentName)
@@ -961,4 +972,18 @@ func isAccessDeniedErr(err error) bool {
 		return true
 	}
 	return grpcstatus.Code(err) == codes.Unauthenticated
+}
+
+func shouldEvictClient(err error) bool {
+	if err == nil {
+		return false
+	}
+	if isAccessDeniedErr(err) {
+		return true
+	}
+	if errors.Is(err, context.DeadlineExceeded) {
+		return true
+	}
+	var unavailable *serviceerror.Unavailable
+	return errors.As(err, &unavailable)
 }
--- a/internal/controller/reconciler_events_test.go
+++ b/internal/controller/reconciler_events_test.go
@@ -19,6 +19,7 @@ import (
 	"github.com/temporalio/temporal-worker-controller/internal/controller/clientpool"
 	"github.com/temporalio/temporal-worker-controller/internal/planner"
 	deploymentpb "go.temporal.io/api/deployment/v1"
+	enumspb "go.temporal.io/api/enums/v1"
 	"go.temporal.io/api/serviceerror"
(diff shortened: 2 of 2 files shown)


jaypipes:
@rupesh-parab-one-app I'm super embarrassed at the long delay in getting back to this important PR. Apologies. If you wouldn't mind rebasing to main to address the merge conflicts, that would be great. I'll do a review on the revision ASAP.

jaypipes:
@rupesh-parab-one-app thanks for pushing a new rev! FYI, if you'd like to chat about this PR (and any other one!), feel free to find us on the Temporal community Slack \#temporal-workers channel. :)

jaypipes (approved):
👍 good stuff, thank you @rupesh-parab-one-app :)

internal/controller/worker_controller.go
```diff
@@ -932,3 +943,17 @@ func isAccessDeniedErr(err error) bool {
...
+		return false
+	}
+	if isAccessDeniedErr(err) {
+		return true
+	}
+	if errors.Is(err, context.DeadlineExceeded) {
+		return true
+	}
+	var unavailable *serviceerror.Unavailable
+	return errors.As(err, &unavailable)
```
carlydf, line 988: why evict on Unavailable? The [gRPC definition](https://grpc.io/docs/guides/status-codes/) of this status describes it as
> most likely a transient condition, which can be corrected by retrying with a backoff
so I don't think it should be included here

internal/controller/worker_controller.go
```diff
@@ -932,3 +943,17 @@ func isAccessDeniedErr(err error) bool {
...
+func shouldEvictClient(err error) bool {
+	if err == nil {
+		return false
+	}
+	if isAccessDeniedErr(err) {
+		return true
+	}
+	if errors.Is(err, context.DeadlineExceeded) {
+		return true
+	}
```
carlydf, line 986: similarly, I understand that the issue you reported manifested as `context.DeadlineExceeded`, so including `context.DeadlineExceeded` as an eviction condition is the main point of this PR.

But I do have misgivings about labeling a `context.DeadlineExceeded` as a signifier of a "poisoned" connection. Context deadline exceeded could happen due to a very large range of things going on in the server that are unrelated to the client. And, if the SDK is doing something pathologically wrong with a "poisoned" client config, perhaps it would be more appropriate to ask the SDK to return a different error type in that case.

In #328, you call out 

> `ResourceExhausted` may represent throttling where client churn does not help.

as a reason for excluding `ResourceExhausted`, but the same exact reasoning thing can be said about `context.DeadlineExceeded`.

Can we at least put a comment here explaining honestly why we are including `context.DeadlineExceeded` error type in the "evictable" errors? It's ok with me if it says something like "saw this error repeatedly until pod restarted and client cache was cleared. not totally sure how the client became unhealthy, but prefer to err on the side of evicting in the case of this error instead of repeating it indefinitely."

I think we are protected from a perpetual and fast `evict -> reload client -> context deadline error -> evict` loop by the exponential backoff for erroring Reconciles (it could still be perpetual if recreating the client doesnt solve the problem, but at least if would slow over time). That is why I personally am ok with erring on the side of eviction. But I want this PR to be honest about what we do and don't know about why this error happened in the first place.

internal/controller/worker_controller.go
carlydf: As discussed in the other comment, if we evict and it still errors the next time (because the problem was not the client), that is ultimately ok. So.. both of these comments are non-blocking

rupesh-one: You're right, and I don't have the evidence to defend it. I'll drop `Unavailable` unless you'd rather keep it once you've seen the other thread.

Two things I got wrong when I added it. First, I assumed the SDK would have retried a genuinely transient one before we saw it. That is true wherever we go through a deployment handle — all nine handle methods wrap the call in `newGRPCContext(ctx, defaultGrpcRetryParameters(ctx))` (`go.temporal.io/sdk@v1.41.1/internal/internal_worker_deployment_client.go`, e.g. `Describe` at `:181`, `SetCurrentVersion` at `:258`) — so it covers `handleDeletion` and all of `executePlan`'s writes. But `GetWorkerDeploymentState` calls `WorkflowService().DescribeWorkerDeployment(ctx, ...)` directly (`internal/temporal/worker_deployment.go:82`) with the plain reconcile context and no `retry.ConfigKey`, so `NewRetryOptionsInterceptor` takes its else branch and calls `grpc_retry.Disable()` (`go.temporal.io/sdk@v1.41.1/internal/common/retry/interceptor.go:140-143`). No retries at all there. So at that one call site a single transient blip evicts a healthy client, which is your point exactly.

Second, `errors.As(err, &unavailable)` misses `*serviceerror.NamespaceUnavailable`, which `convert.go:80-86` returns when the status carries a `NamespaceUnavailableFailure` detail. So the branch does not even cover the Unavailable family cleanly.

Worth reading alongside the other thread: the deadline branch below it does not match a Temporal-origin deadline at all, so today `Unavailable` is the only transport-failure branch matching anything Temporal-origin. If we drop both, the predicate falls back to the auth behaviour that already exists on `main`, and this PR stops doing anything. That is a fine outcome if it's the right one, but I'd rather say it out loud than have it happen by accident.

internal/controller/worker_controller.go
rupesh-one: I went looking for the explanation you asked for, this check does not match a Temporal deadline.

The SDK's `errorInterceptor` is its outermost unary interceptor and rewrites call errors through `serviceerror.FromStatus(status.Convert(err))` (`go.temporal.io/sdk@v1.41.1/internal/grpc_dialer.go:134`, body at `:201-208`; it passes `GrpcMessageTooLargeError` through unchanged). `codes.DeadlineExceeded` becomes `*serviceerror.DeadlineExceeded` (`go.temporal.io/api@v1.62.8/serviceerror/convert.go:77`), which implements only `Error()` and `Status()` — no `Unwrap`, no `Is`. So `errors.Is(err, context.DeadlineExceeded)` cannot reach the sentinel. The message text survives via `st.Message()`, which is why the logs read `context deadline exceeded` and why this looked correct to me. My own tests pass only because they inject the raw sentinel through a stub and never touch gRPC.

This also answers something I had wrong in #328. The deletion path *does* install the SDK retry config, so I assumed a deadline there had already been retried. It hadn't: `codes.DeadlineExceeded` is deliberately absent from the SDK's retryable set, with the comment "they are coming from go context and 'context errors are not retriable based on user settings' by gRPC library" (`go.temporal.io/sdk@v1.41.1/internal/common/retry/interceptor.go:103-108`). So the SDK forwards a deadline unretried, and then the controller's check doesn't match it. Nothing in the stack reacted.

Reproduction on a branch rather than more code here, tests only, based on this PR's head: https://github.com/rupesh-parab-one-app/temporal-worker-controller/blob/5e80872729d8983673771e9808dd79b1993ab422/REPRO.md

It shows it twice: constructing the error the way the interceptor does, where the predicate returns false, and driving a real SDK client over an in-memory bufconn transport with no Temporal server. A real call whose deadline fires returns `*serviceerror.DeadlineExceeded` with the message `context deadline exceeded`; an unreachable endpoint returns `*serviceerror.Unavailable`. Happy to inline the relevant few lines here if you'd rather not follow a fork link.

On your `ResourceExhausted` parallel: your policy argument stands untouched, and I should be clear I'm not answering it. I'm saying something narrower — that today the question is moot, because the branch is unreachable for the case it was written for. The one deadline it *can* match is a non-Temporal one, such as a Kubernetes write in `executePlan` running out the 5-minute reconcile budget (`worker_controller.go:113`), and evicting the Temporal client there achieves nothing.

One aside on backoff, since you raised it: it isn't present on the cleanup path. `Reconcile` swallows the `handleDeletion` error and returns `ctrl.Result{RequeueAfter: 10 * time.Second}, nil` (`worker_controller.go:157-159`), so controller-runtime calls `Forget` and requeues at a flat 10s rather than rate-limiting — which is the retry cadence I described in #328. The main paths do return real errors, so backoff applies there. I'd rather not change those lines in this PR; I'll file it separately.

Where that leaves the comment you asked for: I don't think it's the right change on its own. My suggestion, if you agree, is to match the concrete type *in addition to* the sentinel — `errors.As(err, &*serviceerror.DeadlineExceeded)` alongside the existing `errors.Is`, since swapping one for the other would drop the Kubernetes case — drop `Unavailable` per your other comment, and then write the honest comment against what's left. The alternative is to drop the deadline case as unproven, but combined with dropping `Unavailable` that reduces the predicate to the auth behaviour already on `main`. Happy either way; tell me which and I'll push it.

internal/controller/worker_controller.go
jaypipes: @rupesh-one @rupesh-parab-one-app @carlydf sorry for the long delayed review.

After reading both your comments, I was struck with the thought: is it really a bad thing for cache eviction to be overly aggressive? If a transient error like `serviceerror.Unavailable` ends up evicting a healthy client and a new client is constructed on the next reconcile loop, is that really a bad thing? To me, getting rid of bad client connections from the cache and keeping the code paths readable and reasonable is more important than getting all complicated trying to figure out whether a transient error is truly transient or just a blip.

I'd be happy to merge this PR as-is since I think the code paths for determining whether a cached client should be evicted are easy to read and reason about, and, again, I don't see a problem with an aggressive eviction policy.

jaypipes (approved):
See inline comment, but I'd be happy to merge this PR as-is.

internal/controller/worker_controller.go
carlydf: sure, I think that makes sense. I definitely agree with your point on code complexity. Before merging, I'd like the title to reflect this decision a bit better (since the PR title is what we will see in the commit history and release notes). I changed it to `fix: evict cached Temporal client on failures that indicate possible bad-client` -- feel free to revert if you don't like that

carlydf (approved):

internal/controller/worker_controller.go
jaypipes: Yup, that new title is great, thank you @carlydf :)

PR #331, opened 2026-05-22
Scope Deployment cache to worker deployments

Shivs11:
## Summary

This scopes the controller-runtime cache for `apps/v1.Deployment` objects to only Deployments labeled with `temporal.io/deployment-name`.

## Why

We investigated a report where one controller manager pod was using several GiB of memory while CPU was almost idle. The cluster only had one `TemporalWorkerDeployment`, but it had thousands of normal Kubernetes `Deployment` objects for unrelated services.

The confusing part is how `Owns(&appsv1.Deployment{})` behaves in controller-runtime.

A useful way to think about this is that there are two separate layers:

1. The manager starts a Kubernetes `Deployment` watcher/cache. This layer asks the API server for Deployment objects, watches for Deployment changes, and keeps the watched objects in memory.
2. The controller event handler then decides which of those Deployment events should enqueue a reconcile. `Owns(&appsv1.Deployment{})` belongs to this second layer: it makes sure only Deployments owned by a `TemporalWorkerDeployment` trigger reconciles for that owner.

Before this PR, layer 2 was already narrow, but layer 1 was still broad. So the controller was only reconciling its own Deployments, but the underlying Deployment watcher/cache could still list, watch, and retain unrelated Deployments from the cluster.

That means memory could grow with all Kubernetes Deployments in the cluster, not just Temporal-managed worker Deployments.

This PR narrows the first layer too. The manager cache now only admits Deployments that have the existing `temporal.io/deployment-name` label, which is the label we already put on worker Deployments generated by this controller.

## Changes

- Export the existing `temporal.io/deployment-name` label constant as `k8s.WorkerDeploymentNameLabel`.
- Configure `cache.Options.ByObject` for `appsv1.Deployment` with an existence selector on `temporal.io/deployment-name`.
- Add unit coverage that the manager cache selector only admits Deployments with the worker label.
- Add unit coverage that generated worker Deployments carry the label the cache selector depends on.
- Update tests to use the exported label constant.

## Notes

This does not introduce a new Kubernetes label. Generated worker Deployments already had `temporal.io/deployment-name`; this change reuses that label to scope the Deployment watcher/cache.

Residual risk: a pre-existing or manually modified controller-owned Deployment missing `temporal.io/deployment-name` will no longer be visible to the cached Deployment reader/watch. Normal upgrade risk should be low because the controller-generated Deployments already used this label before the PR.

## Testing

- `go test ./cmd`
- `go test ./internal/k8s`
- `go test ./...`
- `git diff --check`


internal/controller/cache_test.go
Shivs11, line 1: just made a new file here only because the main changes were also in the "main" file

cmd/main.go
```diff
@@ -41,6 +47,21 @@ func init() {
 	//+kubebuilder:scaffold:scheme
 }
 
+func managerCacheOptions() (cache.Options, error) {
```
jaypipes: Couple things... first, please add the (excellent) explanation about these cache options from the PR summary as a code comment in here. That will help future us understand why we're doing this :) Second, please place this function into the `internal/controller` package and rename it to `NewCacheOptions()`. That way, you can also move the unit test of this code into that package.

jaypipes (changes requested):
great stuff, @Shivs11! couple small requests for you inline, though :)

jaypipes (approved):
muy bueno.

PR #351, opened 2026-05-26
feat(wrt): auto-inject KEDA Temporal trigger metadata for per-version ScaledObjects

nikoul:
## Summary

Extends `WorkerResourceTemplate` auto-injection to populate the `workerDeploymentName` and `workerDeploymentBuildId` fields in KEDA `ScaledObject` `triggers[*].metadata` when the trigger `type` is `"temporal"`. This unblocks using KEDA's Temporal scaler ([kedacore/keda#7672](https://github.com/kedacore/keda/pull/7672)) as the templated resource for per-version backlog scaling, as requested in #286.

Closes (or contributes towards) #286.

## Motivation

For users who already run KEDA cluster-wide for non-Temporal workloads, KEDA owns the `external.metrics.k8s.io` APIService — only one external metrics provider can be installed per cluster. That makes the documented WRT-HPA + prometheus-adapter path unavailable: prometheus-adapter cannot co-exist with KEDA on the same APIService.

KEDA's recent Worker Deployment Versioning support (kedacore/keda#7672) makes it possible for a KEDA `ScaledObject` to scale a single worker version by querying Temporal Cloud directly with `workerDeploymentName` + `workerDeploymentBuildId`. WRT already supports templating arbitrary kinds (verified: `spec.template` is unstructured, SSA-applied, with a kind allow-list in the chart) and `scaleTargetRef: {}` auto-fill works recursively. The only missing piece was injecting the per-version identifiers into the KEDA trigger metadata — same source values the controller already injects into HPA External metric selectors, different JSON path.

## Behaviour

For each templated resource that contains:

```yaml
spec:
  triggers:
    - type: temporal
      metadata:
        workerDeploymentName: ""        # placeholder, any string value
        workerDeploymentBuildId: ""     # placeholder, any string value
        # ...other user-provided KEDA Temporal scaler metadata
```

the controller sets:

- `workerDeploymentName` ← `wrt.Spec.EffectiveWorkerDeploymentName()`
- `workerDeploymentBuildId` ← the per-version build ID

Injection is **opt-in**: the keys are only touched if already present in the template's `metadata` map (mirroring the existing `metrics[*].external.metric.selector.matchLabels` merge pattern). Non-`temporal` triggers in the same `ScaledObject` are left untouched.

## Example

```yaml
kind: WorkerResourceTemplate
metadata:
  name: my-worker-scaler
  namespace: default
spec:
  workerDeploymentRef:
    name: my-worker
  template:
    apiVersion: keda.sh/v1alpha1
    kind: ScaledObject
    spec:
      scaleTargetRef: {}                     # auto-injected (versioned Deployment)
      minReplicaCount: 1
      maxReplicaCount: 10
      pollingInterval: 30
      triggers:
        - type: temporal
          metadata:
            endpoint: us-east-1.aws.api.temporal.io:7233
            namespace: my-temporal-ns
            taskQueue: my-task-queue
            workerDeploymentName: ""         # auto-injected
            workerDeploymentBuildId: ""      # auto-injected
            targetQueueSize: "10"
          authenticationRef:
            kind: ClusterTriggerAuthentication
            name: temporal-cloud-auth
```

The controller will stamp one `ScaledObject` per active worker version, each with the correct `workerDeploymentBuildId`, and delete them on sunset via the existing `DeleteWorkerResources` planner (kind-agnostic).

## Implementation

- `appendTemporalTriggerMetadata(spec, twdName, buildID)` — new helper that walks `spec.triggers[*]`, filters on `type == "temporal"`, and sets the two keys if already present.
- `autoInjectFields` signature now takes `twdName, buildID` so it can pass them through. Behaviour for HPA / matchLabels / scaleTargetRef is unchanged.

Helm-side: adding `ScaledObject` to `workerResourceTemplate.allowedResources` + the controller's `ClusterRole` is left as a follow-up (or a separate commit in this PR if maintainers prefer). For users who want this today, they can add the entry to their chart values without further controller changes.

## Tests

New `TestAutoInjectFields_TemporalTriggerMetadata` covers:

- injects when both keys present (empty string placeholders)
- overwrites stale values
- no-op when keys absent (opt-in semantics)
- non-`temporal` triggers are untouched (heterogeneous trigger array)
- `scaleTargetRef: {}` auto-fill still works on `ScaledObject`
- no panic when `spec.triggers` is absent

All existing tests still pass (`go test ./...`).

### DIFF
--- a/api/v1alpha1/workerresourcetemplate_webhook.go
+++ b/api/v1alpha1/workerresourcetemplate_webhook.go
@@ -342,8 +342,8 @@ func checkTemporalTriggerMetadataNotSet(spec map[string]interface{}, path *field
 		if !ok {
 			continue
 		}
-		triggerType, _ := trigger["type"].(string)
-		if triggerType != "temporal" {
+		triggerType, ok := trigger["type"].(string)
+		if !ok || triggerType != "temporal" {
 			continue
 		}
 		metadata, ok := trigger["metadata"].(map[string]interface{})
--- a/api/v1alpha1/workerresourcetemplate_webhook.go
+++ b/api/v1alpha1/workerresourcetemplate_webhook.go
@@ -313,9 +313,59 @@ func validateWorkerResourceTemplateSpec(spec WorkerResourceTemplateSpec, allowed
 		// the controller generates the correct per-version values at render time.
 		// User labels (e.g. task_type: "Activity") are allowed alongside the controller-owned keys.
 		checkMetricSelectorLabelsNotSet(innerSpec, innerSpecPath, &allErrs)
+
+		// 8. triggers[*].metadata.workerDeploymentName / workerDeploymentBuildId
+		// (KEDA ScaledObject): the controller owns these for triggers of type "temporal" so
+		// each per-version ScaledObject queries the correct Temporal-server worker deployment.
+		// Allow empty-string opt-in ("") and reject any other value.
+		checkTemporalTriggerMetadataNotSet(innerSpec, innerSpecPath, &allErrs)
 	}
 
 	return warnings, allErrs
+}
+
+// checkTemporalTriggerMetadataNotSet validates that, for each KEDA trigger of type "temporal"
+// in spec.triggers, the controller-owned metadata keys (workerDeploymentName,
+// workerDeploymentBuildId) are either absent or set to the empty-string opt-in sentinel.
+// A non-empty value is rejected — the controller fills these per-version at render time
+// and a hardcoded value would point at the wrong worker deployment / build.
+// Non-temporal triggers (prometheus, cron, etc.) are not validated.
+func checkTemporalTriggerMetadataNotSet(spec map[string]interface{}, path *field.Path, allErrs *field.ErrorList) {
+	triggers, ok := spec["triggers"].([]interface{})
+	if !ok {
+		return
+	}
+	triggersPath := path.Child("triggers")
+	for i, t := range triggers {
+		trigger, ok := t.(map[string]interface{})
+		if !ok {
+			continue
+		}
+		triggerType, _ := trigger["type"].(string)
+		if triggerType != "temporal" {
+			continue
+		}
+		metadata, ok := trigger["metadata"].(map[string]interface{})
+		if !ok {
+			continue
+		}
+		triggerPath := triggersPath.Index(i).Child("metadata")
+		for _, key := range []string{"workerDeploymentName", "workerDeploymentBuildId"} {
+			val, present := metadata[key]
+			if !present {
+				continue
+			}
+			s, isString := val.(string)
+			if !isString || s != "" {
+				*allErrs = append(*allErrs, field.Forbidden(
+					triggerPath.Child(key),
+					"if "+key+" is present on a temporal trigger, the controller owns it and "+
+						"will set it to the per-version value; set it to \"\" to opt in to auto-injection, "+
+						"or remove it entirely if you do not need per-version trigger metadata",
+				))
+			}
+		}
+	}
 }
 
 // checkScaleTargetRefNotSet recursively traverses obj looking for any scaleTargetRef that is
--- a/api/v1alpha1/workerresourcetemplate_webhook.go
+++ b/api/v1alpha1/workerresourcetemplate_webhook.go
@@ -326,9 +326,10 @@ func validateWorkerResourceTemplateSpec(spec WorkerResourceTemplateSpec, allowed
 
 // checkTemporalTriggerMetadataNotSet validates that, for each KEDA trigger of type "temporal"
 // in spec.triggers, the controller-owned metadata keys (workerDeploymentName,
-// workerDeploymentBuildId) are either absent or set to the empty-string opt-in sentinel.
-// A non-empty value is rejected — the controller fills these per-version at render time
-// and a hardcoded value would point at the wrong worker deployment / build.
+// workerDeploymentBuildId, namespace) are either absent or set to the empty-string opt-in
+// sentinel. A non-empty value is rejected — the controller fills these at render time from
+// the WorkerDeployment's connection / per-version state, and a hardcoded value would point
+// at the wrong worker deployment / build / Temporal namespace.
 // Non-temporal triggers (prometheus, cron, etc.) are not validated.
 func checkTemporalTriggerMetadataNotSet(spec map[string]interface{}, path *field.Path, allErrs *field.ErrorList) {
 	triggers, ok := spec["triggers"].([]interface{})
@@ -350,7 +351,7 @@ func checkTemporalTriggerMetadataNotSet(spec map[string]interface{}, path *field
 			continue
 		}
 		triggerPath := triggersPath.Index(i).Child("metadata")
-		for _, key := range []string{"workerDeploymentName", "workerDeploymentBuildId"} {
+		for _, key := range []string{"workerDeploymentName", "workerDeploymentBuildId", "namespace"} {
 			val, present := metadata[key]
 			if !present {
 				continue
@@ -360,8 +361,8 @@ func checkTemporalTriggerMetadataNotSet(spec map[string]interface{}, path *field
 				*allErrs = append(*allErrs, field.Forbidden(
 					triggerPath.Child(key),
 					"if "+key+" is present on a temporal trigger, the controller owns it and "+
-						"will set it to the per-version value; set it to \"\" to opt in to auto-injection, "+
-						"or remove it entirely if you do not need per-version trigger metadata",
+						"will set it to the per-version or per-connection value; set it to \"\" to opt in "+
+						"to auto-injection, or remove it entirely if you do not need it",
 				))
 			}
 		}
--- a/api/v1alpha1/workerresourcetemplate_webhook.go
+++ b/api/v1alpha1/workerresourcetemplate_webhook.go
@@ -316,22 +316,23 @@ func validateWorkerResourceTemplateSpec(spec WorkerResourceTemplateSpec, allowed
 
 		// 8. triggers[*].metadata.workerDeploymentName / workerDeploymentBuildId
 		// (KEDA ScaledObject): the controller owns these for triggers of type "temporal" so
-		// each per-version ScaledObject queries the correct Temporal-server worker deployment.
-		// Allow empty-string opt-in ("") and reject any other value.
-		checkTemporalTriggerMetadataNotSet(innerSpec, innerSpecPath, &allErrs)
+		// each per-version ScaledObject queries the correct Temporal Worker Deployment Version
+		// (workerDeploymentName + workerDeploymentBuildId). Allow empty-string opt-in ("") and
+		// reject any other value.
+		checkKEDATriggerMetadata(innerSpec, innerSpecPath, &allErrs)
 	}
 
 	return warnings, allErrs
 }
 
-// checkTemporalTriggerMetadataNotSet validates that, for each KEDA trigger of type "temporal"
+// checkKEDATriggerMetadata validates that, for each KEDA trigger of type "temporal"
 // in spec.triggers, the controller-owned metadata keys (workerDeploymentName,
 // workerDeploymentBuildId, namespace) are either absent or set to the empty-string opt-in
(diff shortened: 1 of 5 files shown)


internal/k8s/workerresourcetemplates.go
```diff
@@ -208,6 +218,43 @@ func appendMetricsMatchLabelSelector(spec map[string]interface{}, metricSelector
...
+		triggerType, _ := trigger["type"].(string)
+		if triggerType != "temporal" {
+			continue
+		}
+		metadata, ok := trigger["metadata"].(map[string]interface{})
+		if !ok {
+			continue
+		}
+		if _, present := metadata["workerDeploymentName"]; present {
+			metadata["workerDeploymentName"] = twdName
```
carlydf: I know this is confusing, but for uniqueness across Kubernetes namespaces, the `workerDeploymentName` that the Temporal server uses is actually `<kubernetes_namespace>/<kubernetes_WorkerDeployment_resource_name>`. The generator function for that is [here](https://github.com/temporalio/temporal-worker-controller/blob/bc58d644b7b9418e45644c47f5189eba865d1f3c/internal/k8s/deployments.go#L139-L143). So here, `twdName` is the Kubernetes `WorkerDeployment` resource name, not the server-side Worker Deployment name.

I'll suggest some changes to this to correct it and try to make it more clear!

carlydf: I checked out your PR and made [this commit](https://github.com/temporalio/temporal-worker-controller/commit/26e0c050b5f2914b79a9e1953c9c6bda88b0308a) on top of it with how I suggest to fix this. Hope that makes sense; let me know what you think.

internal/k8s/workerresourcetemplates_test.go
```diff
@@ -266,20 +266,131 @@ func TestAutoInjectFields_MetricSelector(t *testing.T) {
...
+		autoInjectFields(spec, "my-worker-abc123", twdName, buildID, nil, nil)
+		md := spec["triggers"].([]interface{})[0].(map[string]interface{})["metadata"].(map[string]interface{})
+		assert.Equal(t, twdName, md["workerDeploymentName"])
+		assert.Equal(t, buildID, md["workerDeploymentBuildId"])
+		assert.Equal(t, "us-east-1.aws.api.temporal.io:7233", md["endpoint"])
+		assert.Equal(t, "default", md["namespace"])
+		assert.Equal(t, "my-tq", md["taskQueue"])
+	})
+
+	t.Run("overwrites pre-existing values when keys are present", func(t *testing.T) {
```
carlydf: I think this is valid, but just an idea, in preparation for the auto-injection of `spec.metrics[*].external.metric.selector.matchLabels`, the WorkerResourceTemplate validating webhook rejects objects that have the fields set, to avoid confusion and be very explicit where the fields are coming from. I think it would be consistent to do KEDA templating similarly. See https://github.com/temporalio/temporal-worker-controller/blob/main/docs/worker-resource-templates.md#auto-injection for existing semantics.

carlydf: Also, it would be most consistent if the controller templates the `namespace` field in the Temporal trigger too. The idea being, we know the Temporal namespace that the Worker Deployment is in, so no need for the user to mis-configure it.

internal/k8s/workerresourcetemplates.go
nikoul: It does, thank you

internal/k8s/workerresourcetemplates_test.go
nikoul: Thanks, I think I've implemented something that answers both those suggestions now.

nikoul:
Hello @carlydf @jlegrone , I have tested and deployed this on our internal platform, and it seems to work well with KEDA  2.20.
Could you help move this PR forward? Thanks!

api/v1alpha1/workerresourcetemplate_webhook.go
```diff
@@ -313,11 +313,62 @@ func validateWorkerResourceTemplateSpec(spec WorkerResourceTemplateSpec, allowed
 		// the controller generates the correct per-version values at render time.
 		// User labels (e.g. task_type: "Activity") are allowed alongside the controller-owned keys.
 		checkMetricSelectorLabelsNotSet(innerSpec, innerSpecPath, &allErrs)
+
+		// 8. triggers[*].metadata.workerDeploymentName / workerDeploymentBuildId
+		// (KEDA ScaledObject): the controller owns these for triggers of type "temporal" so
+		// each per-version ScaledObject queries the correct Temporal-server worker deployment.
```
Shivs11: nit: imo, a ScaledObject maps 1:1 a WorkerDeploymentVersion and i think it should be something like:

> ...so each per-version ScaledObject queries the correct Temporal Worker Deployment Version (workerDeploymentName + workerDeploymentBuildId).

internal/k8s/deployments.go
```diff
@@ -139,8 +145,12 @@ func ComputeBuildID(w *temporaliov1alpha1.WorkerDeployment) string {
 
 // ComputeWorkerDeploymentName generates the base worker deployment name
 func ComputeWorkerDeploymentName(w *temporaliov1alpha1.WorkerDeployment) string {
+	return computeWorkerDeploymentName(w.GetNamespace(), w.GetName())
+}
+
+func computeWorkerDeploymentName(k8sNamespace, workerDeploymentResourceName string) string {
```
Shivs11, line 151: i guess we don't need another helper out here for this right? we can just have the computation to be done in the exported ComputeWorkerDeploymentName

internal/k8s/workerresourcetemplates.go
```diff
@@ -86,24 +86,30 @@ func RenderWorkerResourceTemplate(
...
 
-	twdName := wrt.Spec.EffectiveWorkerDeploymentName()
-	selectorLabels := ComputeSelectorLabels(twdName, buildID)
+	// The name of the Worker Deployment Kubernetes resource
+	wdName := wrt.Spec.EffectiveWorkerDeploymentName()
+
+	// The name of the Worker Deployment in Temporal Server, prefixed by the Kubernetes namespace of the resource
+	twdName := computeWorkerDeploymentName(wrt.Namespace, wdName)
+
+	selectorLabels := ComputeSelectorLabels(wdName, buildID)
```
Shivs11, line 95: nit: (not something you have changed but thought of still placing this here) -- seems like the function you are calling here has it's parameter named as twdName, whereas we now only use wdName

maybe we can update that

internal/k8s/workerresourcetemplates.go
```diff
@@ -86,24 +86,30 @@ func RenderWorkerResourceTemplate(
 		return nil, fmt.Errorf("failed to unmarshal spec.template: %w", err)
 	}
 
-	twdName := wrt.Spec.EffectiveWorkerDeploymentName()
-	selectorLabels := ComputeSelectorLabels(twdName, buildID)
+	// The name of the Worker Deployment Kubernetes resource
+	wdName := wrt.Spec.EffectiveWorkerDeploymentName()
+
+	// The name of the Worker Deployment in Temporal Server, prefixed by the Kubernetes namespace of the resource
+	twdName := computeWorkerDeploymentName(wrt.Namespace, wdName)
```
Shivs11: so while twdName makes sense for what this variable should represent, we recently migrated naming one of our Custom Resources (CR) from TWD (TemporalWorkerDeployment) to Worker Deployment (WD) and this took me a second to realize that it did not mean the same thing!

I wonder if we can change this and call it something like serverWD!

Shivs11 (approved):
approved with some comments -- hoping you have a peek at them!

api/v1alpha1/workerresourcetemplate_webhook.go
```diff
@@ -313,11 +313,62 @@ func validateWorkerResourceTemplateSpec(spec WorkerResourceTemplateSpec, allowed
...
+		// Allow empty-string opt-in ("") and reject any other value.
+		checkTemporalTriggerMetadataNotSet(innerSpec, innerSpecPath, &allErrs)
 	}
 
 	return warnings, allErrs
 }
 
+// checkTemporalTriggerMetadataNotSet validates that, for each KEDA trigger of type "temporal"
```
jaypipes: Can we name this `checkKEDATriggerMetadata` please to make it more clear this is a KEDA-specific check?

api/v1alpha1/workerresourcetemplate_webhook.go
```diff
@@ -313,11 +313,62 @@ func validateWorkerResourceTemplateSpec(spec WorkerResourceTemplateSpec, allowed
...
+		for _, key := range []string{"workerDeploymentName", "workerDeploymentBuildId", "namespace"} {
+			val, present := metadata[key]
+			if !present {
+				continue
+			}
+			s, isString := val.(string)
+			if !isString || s != "" {
+				*allErrs = append(*allErrs, field.Forbidden(
+					triggerPath.Child(key),
+					"if "+key+" is present on a temporal trigger, the controller owns it and "+
```
jaypipes:
```suggestion
					"if "+key+" is present on a KEDA ScaledObject temporal trigger, the controller owns it and "+
```

internal/k8s/workerresourcetemplates.go
```diff
@@ -208,6 +225,60 @@ func appendMetricsMatchLabelSelector(spec map[string]interface{}, metricSelector
...
+//
+// twdName must be the fully-qualified Temporal-server Worker Deployment name
+// (computeWorkerDeploymentName(k8sNamespace, wdName)) — that's what workers register
+// with at Temporal Cloud and what KEDA's scaler needs to query backlog for a specific
+// version. temporalNamespace is the Temporal Cloud namespace from the connection.
+//
+// Required by KEDA's Temporal scaler (kedacore/keda#7672) for per-version backlog
+// scaling against Temporal's Worker Deployment Versioning model. Non-temporal
+// triggers in the same ScaledObject are untouched.
+func appendTemporalTriggerMetadata(spec map[string]interface{}, twdName, buildID, temporalNamespace string) {
```
jaypipes: recommend renaming this to `appendKEDATriggerMetadata()` to make it clearer that this is a KEDA-specific thing.

jaypipes (commented):
agree with @Shivs11 's review comments, plus left a couple of my own nits. @nikoul if you can address these shortly, we can approve and merge this PR today. thank you!

api/v1alpha1/workerresourcetemplate_webhook.go
nikoul: Done in 314b4e2.

internal/k8s/workerresourcetemplates.go
nikoul: Done in 314b4e2.

api/v1alpha1/workerresourcetemplate_webhook.go
nikoul: Applied in 314b4e2 (updated the test assertions too).

api/v1alpha1/workerresourcetemplate_webhook.go
nikoul: Reworded in 314b4e2.

internal/k8s/workerresourcetemplates.go
nikoul: Renamed to `serverWDName` in the render path (314b4e2). Left the pre-existing `twdName` elsewhere to keep this scoped.

internal/k8s/workerresourcetemplates.go
nikoul: Renamed to `serverWDName` in 314b4e2.

internal/k8s/deployments.go
nikoul: Kept it: the WRT render path needs the name from `(namespace, wdName)` strings and has no `*WorkerDeployment` to pass to the exported func. Happy to inline in both spots if you'd prefer.

internal/k8s/workerresourcetemplates_test.go
nikoul: Done: the webhook now rejects non-empty values (empty `\"\"` is the opt-in), mirroring matchLabels. See `checkKEDATriggerMetadata`.

jaypipes (approved):
Thanks very much @nikoul :)

Issue #317, opened 2026-05-05
Bug: controller steals ownership of some TemporalWorkerDeployment spec fields when adding finalizer


anatolebeuzon:
After upgrading from v1.5.2 to v1.6.0, `helm upgrade` of TemporalWorkerDeployment resources fails with:

```
Apply failed with 1 conflict: conflict with "temporal-worker-controller"
using temporal.io/v1alpha1: .spec.sunset.deleteDelay
```

After investigating, I think it is because the controller adds a finalizer via `r.Update(ctx, &workerDeploy)`, which is a full-object PUT. During the roundtrip, Go's `metav1.Duration.MarshalJSON()` normalizes duration strings (e.g. `"4h"` → `"4h0m0s"`). The API server sees this as a field modification and transfers ownership from `helm` to `temporal-worker-controller`.

### Fix

One workaround is to do a SSA apply with `--force-conflicts`, but it's manual.

I think a better fix would be to use a `client.MergeFrom` patch instead of `Update` (as is already done elsewhere) to prevent this:

```go
patch := client.MergeFrom(workerDeploy.DeepCopy())
controllerutil.AddFinalizer(&workerDeploy, finalizerName)
if err := r.Patch(ctx, &workerDeploy, patch); err != nil {
```

Same for the `RemoveFinalizer` call.

rhulet333:
I have the same problem, ownership is stolen and I get stuck with a resource that can't be changed and has a delete protection finalizer on it

jaypipes:
> I have the same problem, ownership is stolen and I get stuck with a resource that can't be changed and has a delete protection finalizer on it

@rhulet333 does the workaround described by @anatolebeuzon work for you? I'll try to get a fix for this pushed in the next few days.

rhulet333:
It does not, specifically because we are using an orchestrator that tries to update it whenever a change is pushed, so while we could manually run it every time, it is very disruptive to our rapid testing environment. Our orchestrator also doesn't currently support a force conflicts apply

rhulet333:
@jaypipes is a fix for this still in the works?

jaypipes:
> [@jaypipes](https://github.com/jaypipes) is a fix for this still in the works?

@rhulet333 yessir!

jaypipes:
@rhulet333 @anatolebeuzon sorry for the delay. PR with [fix is here](https://github.com/temporalio/temporal-worker-controller/pull/369).

carlydf:
[...]

PR #369, opened 2026-06-09
use Patch when adding/removing finalizer

jaypipes:
To prevent TWC from mistakenly stealing ownership of an existing WorkerDeployment, we use `Patch()` instead of `Update()` when adding and removing the finalizer on WorkerDeployment.

Issue #317


Shivs11 (approved):
i wonder: could we test this out in any possible way?

jaypipes:
> i wonder: could we test this out in any possible way?

@Shivs11 I had the same thought. :) But I did not see any existing unit or functional tests that I could copy to add this...

PR #378, opened 2026-06-16
feat(connection): add tlsServerName for TLS SNI override

jclinford:
## What

Adds an optional `tlsServerName` field to `ConnectionSpec` (and the `Connection` CRD) that overrides the server name used for TLS certificate verification.

When set, the controller:
- sets `tls.Config.ServerName` on its own Temporal client connections across all auth paths (mTLS, API key, and no-credential), and
- injects `TEMPORAL_TLS_SERVER_NAME` into worker pods so SDK envconfig-based workers verify against the same name.

The field participates in the connection-spec hash and the client-pool cache key, so clients are re-established when it changes.

## Why

When a `Connection`'s `hostPort` resolves to an address whose name differs from the hostname on the server's TLS certificate, verification fails. The concrete case: reaching Temporal Cloud through an internal AWS NLB (TCP passthrough) — the NLB's DNS name differs from the certificate SANs (`*.aws.api.temporal.io`). `tlsServerName` lets the controller (and workers) verify against the correct name while still dialing the load balancer address.

## Testing

- `make manifests generate` (CRD + deepcopy regenerated)
- `go build ./...`, `go vet ./...`
- Added unit tests:
  - clientpool: `ServerName` override on the mTLS, API-key, and no-credential dial paths
  - deployments: `TEMPORAL_TLS_SERVER_NAME` env injection + connection-spec hash sensitivity
- `go test ./...` passes


jaypipes:
@jclinford hey, thank you very much for this contribution! please do see the comment above about signing the contributor license agreement.

api/v1alpha1/connection_types.go
```diff
@@ -26,6 +26,12 @@ type ConnectionSpec struct {
 	// +kubebuilder:validation:Pattern=`^[a-zA-Z0-9.-]+:[0-9]+$`
 	HostPort string `json:"hostPort"`
 
+	// TLSServerName overrides the server name used to verify the Temporal server
+	// certificate when TLS is enabled.
+	// +optional
+	// +kubebuilder:validation:Pattern=`^[a-zA-Z0-9.-]+$`
+	TLSServerName string `json:"tlsServerName,omitempty"`
```
jaypipes: Instead of adding a specific `TLSServerName` field, how about adding a `TLS` field of type `*tls.Config`? That way, we allow configuring more TLS-specific options than just the server name?

jaypipes (changes requested):

api/v1alpha1/connection_types.go
jclinford: Thanks for the review @jaypipes 

I adjusted the PR. `tls.Config` I don't believe is usable directly because there is security sensitive fields that wouldn't map cleanly to a kube schema? So I wrapped it in a new serializable struct that I believe gets at your point.

Does that work or did you have something else in mind? Let me know if I misunderstood here, thanks!

jaypipes: @jclinford I don't see any code changes in your latest PR revision... am I missing something?

jclinford: Sorry didn't notice push failed - it should be set now!

jaypipes (approved):
Thanks @jclinford, lgtm!

Issue #364, opened 2026-05-28
[bug] `make test-unit` fails due to helm dependency on cert-manager


jaypipes:
Just trying to run `make test-unit` in a clean checkout from main (`@c3b1aa2`) and running into a failure trying to load `api/v1alpha1/webhook_test_suite.go`:

```
·> make test-unit
test -s /home/jaypipes/src/github.com/temporalio/temporal-worker-controller/bin/setup-envtest || GOBIN=/home/jaypipes/src/github.com/temporalio/temporal-worker-controller/bin go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
KUBEBUILDER_ASSETS="/home/jaypipes/src/github.com/temporalio/temporal-worker-controller/bin/k8s/1.29.0-linux-amd64" go test ./... -coverprofile cover.out
# github.com/temporalio/temporal-worker-controller/internal/testhelpers/testlogr
go: no such tool "covdata"
# github.com/temporalio/temporal-worker-controller/cmd
go: no such tool "covdata"
# github.com/temporalio/temporal-worker-controller/hack/demohelper
go: no such tool "covdata"
Running Suite: Webhook Suite - /home/jaypipes/src/github.com/temporalio/temporal-worker-controller/api/v1alpha1
===============================================================================================================
Random Seed: 1779991798

Will run 11 of 11 specs
------------------------------
[BeforeSuite] [FAILED] [0.032 seconds]
[BeforeSuite] 
/home/jaypipes/src/github.com/temporalio/temporal-worker-controller/api/v1alpha1/webhook_suite_test.go:56

  Timeline >>
  STEP: bootstrapping test environment @ 05/28/26 14:09:58.042
  [FAILED] in [BeforeSuite] - /home/jaypipes/src/github.com/temporalio/temporal-worker-controller/api/v1alpha1/webhook_suite_test.go:165 @ 05/28/26 14:09:58.074
  << Timeline

  [FAILED] helm template failed — is helm installed?
  Unexpected error:
      <*exec.ExitError | 0x31d14c306920>: 
      exit status 1
      {
          ProcessState: {
              pid: 418028,
              status: 256,
              rusage: {
                  Utime: {Sec: 0, Usec: 32882},
                  Stime: {Sec: 0, Usec: 10334},
                  Maxrss: 89688,
                  Ixrss: 0,
                  Idrss: 0,
                  Isrss: 0,
                  Minflt: 4399,
                  Majflt: 6,
[...]

PR #379, opened 2026-06-16
fix(helm): use dynamic SA name in proxy-rolebinding subject

### LINKED ISSUE #374: proxy-rolebinding hardcodes "controller-manager" SA name instead of chart-generated SA name
## Describe the bug

When `metrics.enabled: true` and `metrics.disableAuth: false` (the defaults), `auth_proxy.yaml` renders a `ClusterRoleBinding` whose subject is hardcoded to `controller-manager`:

```yaml
subjects:
  -	kind: ServiceAccount
    name: controller-manager   # ← hardcoded
    namespace: {{ .Release.Namespace }}
```

However, `serviceaccount.yaml` generates the SA with a dynamic name:

```yaml
name: {{ .Values.serviceAccount.name | default (printf "%s-service-account" .Release.Name) }}
```

Every other binding in `rbac.yaml` (leader-election and manager rolebindings) correctly uses the same dynamic expression. The `proxy-rolebinding` is the only one that still references the old hardcoded `controller-manager` name.

## Impact

With the default release name recommended in the README (`temporal-worker-controller`), the SA is named `temporal-worker-controller-service-account`. The `proxy-rolebinding` points to a non-existent `controller-manager` SA, so the `kube-rbac-proxy` sidecar never receives the `proxy-role` permissions. Metric scraping silently fails.

## Steps to reproduce

```bash
helm template temporal-worker-controller \
  oci://docker.io/temporalio/temporal-worker-controller \
  --version 1.7.0 \
  --set certmanager.install=false | \
  grep -A5 "proxy-rolebinding" -A5 "kind: ServiceAccount"
```

The rendered `proxy-rolebinding` will show `name: controller-manager` while the `ServiceAccount` is `temporal-worker-controller-service-account`.

## Expected behavior

The `proxy-rolebinding` subject should use the same expression as all other bindings:

```yaml
subjects:
  -	kind: ServiceAccount
    name: {{ .Values.serviceAccount.name | default (printf "%s-service-account" .Release.Name) }}
    namespace: {{ .Release.Namespace }}
```

## Workaround

Set `metrics.disableAuth: true` (skips the `auth_proxy.yaml` template entirely), or explicitly set `serviceAccount.name: controller-manager` in values.

## Environment

- Chart version: 1.7.0
- Kubernetes: 1.28+

alonsnir1:
## Summary

Fixes #374.

The `proxy-rolebinding` ClusterRoleBinding in `auth_proxy.yaml` hardcoded `controller-manager` as the ServiceAccount subject. Every other binding in `rbac.yaml` correctly uses the dynamic expression:

\`\`\`yaml
name: {{ .Values.serviceAccount.name | default (printf "%s-service-account" .Release.Name) }}
\`\`\`

With the default release name recommended in the README (`temporal-worker-controller`), the rendered SA name is `temporal-worker-controller-service-account`, so the `proxy-rolebinding` was pointing to a non-existent SA. This caused the `kube-rbac-proxy` sidecar to silently lack the `proxy-role` permissions, breaking authenticated metrics scraping.

## Change

One-line fix — align the `proxy-rolebinding` subject with all other bindings.

## Testing

Verified with `helm template`:
- Before: subject was `controller-manager`
- After: subject is `temporal-worker-controller-service-account` (matching the rendered SA)

Note: a proper Helm unit test for this is blocked on #364 (`make test-unit` fails due to the cert-manager chart dependency issue in the test harness).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/helm/temporal-worker-controller/templates/auth_proxy.yaml
+++ b/helm/temporal-worker-controller/templates/auth_proxy.yaml
@@ -44,7 +44,7 @@ roleRef:
   name: {{ .Release.Name }}-{{ .Release.Namespace }}-proxy-role
 subjects:
   - kind: ServiceAccount
-    name: controller-manager
+    name: {{ .Values.serviceAccount.name | default (printf "%s-service-account" .Release.Name) }}
     namespace: {{ .Release.Namespace }}
 ---
 apiVersion: v1


helm/temporal-worker-controller/templates/auth_proxy.yaml
```diff
@@ -44,7 +44,7 @@ roleRef:
   name: {{ .Release.Name }}-{{ .Release.Namespace }}-proxy-role
 subjects:
   - kind: ServiceAccount
-    name: controller-manager
+    name: {{ .Values.serviceAccount.name | default (printf "%s-service-account" .Release.Name) }}
```
jaypipes, line 47: 👍 matches the [generated service account name](https://github.com/temporalio/temporal-worker-controller/blob/62d72f8cbb7e55ded8b12fa447f186f5356cf86e/helm/temporal-worker-controller/templates/serviceaccount.yaml#L5) from the templates/serviceaccount.yaml file.

jaypipes (approved):
thank you @alonsnir1 !

PR #404, opened 2026-06-25
feat(helm): make pod and container securityContext configurable via values

arnabmaji:
## Summary

Add optional `podSecurityContext` and `containerSecurityContext` value overrides to `manager.yaml`, allowing users to fully customize security settings without vendoring the chart.

**This is a non-breaking, additive change.** Existing `securityContext.seccompProfile.enabled` behavior is preserved unchanged.

## Problem

The current chart hardcodes:
- **Pod level**: `runAsNonRoot: true` + optional `seccompProfile` toggle (boolean)
- **Container level**: `allowPrivilegeEscalation: false` + `capabilities.drop: ALL`

This makes it impossible to satisfy common security policies (Kyverno, OPA Gatekeeper, Pod Security Standards `restricted` profile) that additionally require:
- `readOnlyRootFilesystem: true`
- `runAsUser` / `runAsGroup` / `fsGroup`
- Container-level `seccompProfile`

Users are forced to either create policy exceptions or vendor the chart locally to patch the templates.

## Solution

Add two new **optional** values that override the defaults when set:

- `podSecurityContext: {}` — when non-empty, replaces the default pod securityContext block entirely
- `containerSecurityContext: {}` — when non-empty, replaces the default container securityContext block for both `manager` and `kube-rbac-proxy`

When empty (the default), existing behavior is preserved exactly as-is, including the `securityContext.seccompProfile.enabled` toggle.

### Template logic:

```yaml
{{- if .Values.podSecurityContext }}
      securityContext:
        {{- toYaml .Values.podSecurityContext | nindent 8 }}
{{- else }}
      securityContext:
        runAsNonRoot: true
        {{- if .Values.securityContext.seccompProfile.enabled }}
        seccompProfile:
          type: RuntimeDefault
        {{- end }}
{{- end }}
```

### Example usage (satisfies Kyverno/PSS restricted):

```yaml
podSecurityContext:
  runAsNonRoot: true
  runAsUser: 65532
  runAsGroup: 65532
  fsGroup: 65532
  seccompProfile:
    type: RuntimeDefault

containerSecurityContext:
  allowPrivilegeEscalation: false
  readOnlyRootFilesystem: true
  runAsGroup: 65532
  seccompProfile:
    type: RuntimeDefault
  capabilities:
    drop:
      - "ALL"
```

## Backwards Compatibility

| Scenario | Behavior |
|----------|----------|
| No changes to values | Identical output to current chart |
| `securityContext.seccompProfile.enabled: true` | Still works as before |
| `podSecurityContext` set | Overrides legacy pod securityContext |
| `containerSecurityContext` set | Overrides legacy container securityContext |

## Files Changed

- `helm/temporal-worker-controller/templates/manager.yaml` — conditional template logic
- `helm/temporal-worker-controller/values.yaml` — new optional values with documented examples


jaypipes (approved):
Nice one. thanks @arnabmaji :)

PR #445, opened 2026-07-17
fix(controller): re-apply worker resources when tracked object is missing

frossbeamish:
## What

Guards the WRT apply hash-skip with an existence check on the rendered object. If the object referenced by an unchanged `LastAppliedHash` no longer exists, the controller falls through to the SSA apply (create-or-update) instead of skipping. Also extracts the WRT apply block from `executePlan` into `applyWorkerResourceTemplates` so it can be unit-tested without a Temporal client.

## Why

We hit this in a live cluster running v1.8.0. Sequence:

1. A worker deployment version is sunset: the controller deletes its versioned Deployment and its WRT-rendered ScaledObject (`DeleteWorkerResources`).
2. The WRT `status.versions` entry for that build ID survives the window (the status update that would drop it can hit an optimistic-concurrency conflict, and the build can be re-registered before the next successful status write, in our case the same build ID was re-registered ~4 minutes after the sunset delete).
3. The build ID returns as the target version. The render is byte-identical, so `RenderedHash == LastAppliedHash` and the SSA apply is skipped forever.
4. The returning version's Deployment is created and scaled to 1 replica (the planner's `spec.Replicas == nil` path, expecting the scaler resource to own replicas), but the ScaledObject is never re-created. Result: the current version runs a single unscaled pod, which in our case OOM-looped under the full task-queue load and stalled all workflows on that queue.

The hash match tells the controller the *render* is unchanged; it says nothing about whether the applied object still exists. The existence check is a cache read against an informer the controller already maintains for these GVKs, so the API-server-load rationale for the skip is preserved.

## Testing

- New unit tests in `internal/controller/execplan_wrt_test.go`:
  - hash-unchanged + resource missing → re-applied, status entry re-recorded
  - hash-unchanged + resource present → still skipped (no Patch calls, verified via interceptor)
- `go test ./internal/controller/ ./internal/planner/` green; `go build ./...`, `go vet`, `gofmt` clean.


jaypipes:
@frossbeamish hey, sorry about this, but the code this PR touches was recently changed to [fix a related issue where the sort order of drained versions was causing unintended updates](https://github.com/temporalio/temporal-worker-controller/pull/458) to the WD status. I'm afraid you will need to rebase and see how to fit this PR into that new code path.

jaypipes:
@frossbeamish ping on this. couple requests... a) please sign the CLA so we can move forward on this PR, b) as noted in my comment above, a bunch of this code changed since you submitted this PR and will need to be rebased and c) please create a tracking issue on Github that describes how to reproduce the issue you are seeing (if you are still seeing it!)

Thank you kindly!

Issue #439, opened 2026-07-12
Address scaling oscillations for scale-to-zero scalers


noamyehudai:
## Problem

`WorkerResourceTemplate` renders an **identical** autoscaler for every version — the template cannot express anything conditioned on version state. With the KEDA ScaledObject flavor (v1.8.0, #351) this forces a lose-lose choice for `minReplicaCount`:

- **`minReplicaCount >= 1`** — KEDA enforces the floor on *every* version, including Drained ones (KEDA applies `minReplicaCount` regardless of trigger activity). The controller force-zeroes drained versions at sunset ("Drained versions and inactive versions that are not the rollout target are always scaled to zero during sunset"), so keda-operator and the controller fight a ~30s scale-up/scale-down write war per drained version — observed live, same family as #350. Worse: `getDeleteDeployments` requires `replicas == 0` before deleting, so a floored autoscaler **blocks version Deployment deletion indefinitely**.

- **`minReplicaCount: 0`** — the war disappears, but the **Current** version scales to zero too. The scaler's metric is `ApproximateBacklogCount` only; a keeping-up worker reads backlog ≈ 0 at nearly every 30s poll, so one `cooldownPeriod` of samples that miss the bursts parks the Current version, and the next task pays a ~2min cold start (poll interval + pod boot). Measured on our production queues over 14 days: a bursty queue had 70+ windows of ≥60 min with zero visible backlog (parks ~5×/day at the 3600s cooldown we run); even a queue with ~1.5k tasks/hour parked several times a day purely via sampling-miss. Related failure mode: a version deployed while traffic is idle is held at 0 pollers and can fail promotion (#438).

The semantic actually needed is **state-aware**:

> `minReplicas: N` while the version is **Current** (or Ramping); `0` once it is Draining/Drained.

Verified nothing supports this today: v1.8.0 `RenderWorkerResourceTemplate` passes only (wrt, deployment, buildID, temporalNamespace) — no version state — and the KEDA temporal scaler (`pkg/scalers/temporal_scaler.go`, v2.20.1 == main) contains zero references to version status/drainage.

## The data already exists in the scaler's response

In deployment-version mode the KEDA scaler calls `DescribeWorkerDeploymentVersion` with `ReportTaskQueueStats: true` and reads only `version_task_queues`. The **same response** carries `worker_deployment_version_info` — including `status` (CURRENT / RAMPING / DRAINING / DRAINED), ramp and drainage info — which is currently discarded.

## Options

[...]

PR #446, opened 2026-07-20
Re-apply rendered WRT resources when a sunset build ID is redeployed; retry failed sunset deletes

gibbonjj:
## Problem

Two related lifecycle bugs in the WorkerResourceTemplate apply/skip logic leave versions without their rendered per-version resources (e.g. a KEDA ScaledObject or HPA):

**1. A rolled-back (redeployed) build ID never gets its rendered resource back.**

- At version sunset the controller explicitly deletes the rendered WRT resources (`plan.DeleteWorkerResources`, executed in `executeK8sOperations`).
- The WRT's per-build status entry is **not** pruned at sunset: `executePlan` only rewrites WRT status when at least one apply in the cycle was non-skipped (the `allSkipped` short-circuit), and the sunset build is excluded from applies (`deletingDeployments` skip in `getWorkerResourceApplies`). In the common steady state every other apply is a hash-skip, so no status write happens and the entry survives with its `lastAppliedHash`.
- Rendering is fully deterministic (same template, deployment name, build ID, WRT UID ⇒ same hash). If the same build ID is redeployed later (a rollback), `RenderedHash == LastAppliedHash`, the SSA apply is skipped — and because skipped applies never trigger a status write, the skip persists indefinitely. The rolled-back version runs pinned at ~1 replica with no autoscaler, silently.

**2. A failed rendered-resource delete at sunset is never retried.**

`getDeleteWorkerResources` derived candidates solely from `DeleteDeployments`, which requires the Deployment to still exist in the informer cache. Once the Deployment delete has succeeded, a failed resource delete has nothing to drive a retry (the code comment claiming "retried next reconcile" was wrong), orphaning a ScaledObject that keeps polling the Temporal API until the WRT itself is deleted.

## Fix

The invariant is now: **a WRT status entry lives exactly as long as its rendered resource is believed to exist** — created on apply attempt, pruned only on a *confirmed* delete (NotFound counts as confirmed).

- **planner**: `getDeleteWorkerResources` takes the k8s state and derives delete candidates from the union of (a) Deployments being sunset this cycle and (b) *orphaned status entries* — build IDs with no Deployment. Each `WorkerResourceRef` now carries `WRTName`/`BuildID` so the controller can prune the matching entry. This makes failed deletes retryable forever and self-heals stale entries left behind by earlier controller versions (first reconcile after upgrade: no-op delete → confirmed → pruned).
- **execplan**: `executeK8sOperations` reports which rendered-resource deletes were confirmed. The status writer now triggers on "any non-skipped apply **or** any confirmed delete" (so the sunset cycle prunes the entry even when every apply was hash-skipped), prunes confirmed-deleted builds' entries, and retains entries that were neither applied nor confirmed-deleted so the retry keeps firing. Previously, entries not covered by the current cycle's applies were silently dropped on any status write — which could also orphan a resource with no retry trace if its delete failed in the same cycle another build applied.

No API/CRD changes; `WorkerResourceRef` is internal to the planner. No new RBAC verbs. The no-op skip optimisation is preserved: an all-skipped cycle with nothing deleted still writes nothing.

## Testing

- New regression tests in `internal/controller/execplan_test.go` drive the real `generatePlan` → `executePlan` path across reconcile cycles against the fake client:
  - **sunset → rollback**: the sunset cycle must delete the rendered resource *and* prune its status entry even when all applies are hash-skipped; redeploying the same build ID must re-apply the resource (pre-fix this failed exactly as in production: entry retained, resource `not found` after rollback).
  - **delete-failure retry**: a failed resource delete after a successful Deployment delete must retain the entry, and the next cycle must re-derive and retry the delete from the orphaned entry, then prune (pre-fix: nothing was retried).
  - **skip guard**: an all-skipped, no-delete cycle must not bump the WRT resourceVersion.
- New planner subtests for orphan-ref generation, live-Deployment suppression, sunset/orphan dedup, and the ref identity fields.
- All tests were written first and failed on the pre-fix behaviour, then pass with the fix. Full suite, `go vet`, `gofmt`, and `make lint` are clean.

Related context (not fixed here): #439 discusses state-aware scaler behaviour for the *alive* phases of the version lifecycle; this PR fixes the delete/redeploy bookkeeping of that same lifecycle.


internal/controller/execplan.go
```diff
@@ -31,15 +31,15 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func (r *WorkerDeploymentReconciler) executeK8sOperations(ctx context.Context, l logr.Logger, workerDeploy *temporaliov1alpha1.WorkerDeployment, p *plan) error {
+func (r *WorkerDeploymentReconciler) executeK8sOperations(ctx context.Context, l logr.Logger, workerDeploy *temporaliov1alpha1.WorkerDeployment, p *plan) ([]planner.WorkerResourceRef, error) {
```
jaypipes, line 42: Please add a docstring explaining the new output parameter since this PR really is changing the single-purpose of the method (from executing kubernetes operations contained in the plan to executing kubernetes operations contained in the plan AND returning any rendered WRTs that were deleted)

internal/controller/execplan_test.go
```diff
@@ -0,0 +1,363 @@
...
+	require.NoError(t, r.Create(context.Background(), depA2))
+	rollbackStatus := temporaliov1alpha1.WorkerDeploymentStatus{
+		TargetVersion:  temporaliov1alpha1.TargetWorkerDeploymentVersion{BaseWorkerDeploymentVersion: baseVersion(buildA, depA2, temporaliov1alpha1.VersionStatusInactive)},
+		CurrentVersion: &temporaliov1alpha1.CurrentWorkerDeploymentVersion{BaseWorkerDeploymentVersion: baseVersion(buildB, depB, temporaliov1alpha1.VersionStatusCurrent)},
+	}
+	p2 := runPlanCycle(t, r, twd, connection, rollbackStatus)
+	assert.Empty(t, p2.DeleteWorkerResources, "redeployed build must not have its rendered resource scheduled for deletion")
+
+	// The rendered resource for build-a must exist again: this is the core regression.
+	assert.NoError(t, getRenderedHPA(r, namespace, hpaA.GetName()), "rendered resource for the redeployed build ID must be re-applied after a sunset")
```
jaypipes: for all of the above called to `assert.XXX`, change them to `require.XXX`. these are required conditions. if the condition fails, there's no point continuing the test scenario.

internal/controller/execplan_test.go
```diff
@@ -0,0 +1,363 @@
...
+
+	statuses = getWRTVersionStatuses(t, r, namespace, wrt.Name)
+	if entry, ok := statuses[buildA]; assert.True(t, ok, "WRT status must have an entry for the redeployed build") {
+		assert.Equal(t, hashA, entry.LastAppliedHash, "re-applied entry must record the rendered hash")
+	}
```
jaypipes: This is an awkward way to structure a conditionally-dependent assertion because it's not obvious that `assert.True()` returns a boolean :) Better to use `require.True()` here and get rid of the conditional logic entirely:

```go
entry, ok := statuses[buildA]
require.True(t, ok, "WRT status must have an entry for the redeployed build") 
require.Equal(t, hashA, entry.LastAppliedHash, "re-applied entry must record the rendered hash")
```

internal/controller/execplan_test.go
```diff
@@ -0,0 +1,363 @@
...
+	p1 := runPlanCycle(t, r, twd, connection, sunsetStatus)
+	require.NotEmpty(t, p1.DeleteWorkerResources, "cycle 1 must attempt to delete the sunset version's rendered resource")
+
+	// Deployment is gone, but the rendered resource survived the failed delete.
+	require.True(t, apierrors.IsNotFound(r.Get(context.Background(), types.NamespacedName{Namespace: namespace, Name: depA.Name}, &appsv1.Deployment{})), "deployment for build-a should be deleted")
+	require.NoError(t, getRenderedHPA(r, namespace, hpaA.GetName()), "rendered resource for build-a should still exist after the failed delete")
+
+	// The status entry must NOT be pruned on a failed delete — it is what drives the retry.
+	statuses := getWRTVersionStatuses(t, r, namespace, wrt.Name)
+	require.Contains(t, statuses, buildA, "WRT status entry must be retained while its rendered resource still exists")
```
jaypipes, line 310: 👍

internal/controller/execplan_test.go
```diff
@@ -0,0 +1,363 @@
...
+		CurrentVersion: &temporaliov1alpha1.CurrentWorkerDeploymentVersion{BaseWorkerDeploymentVersion: baseVersion(buildB, depB, temporaliov1alpha1.VersionStatusCurrent)},
+	}
+	p2 := runPlanCycle(t, r, twd, connection, steadyStatus)
+	assert.NotEmpty(t, p2.DeleteWorkerResources, "delete of the orphaned rendered resource must be retried even though its Deployment no longer exists")
+
+	assert.True(t, apierrors.IsNotFound(getRenderedHPA(r, namespace, hpaA.GetName())), "orphaned rendered resource must be deleted on retry")
+
+	statuses = getWRTVersionStatuses(t, r, namespace, wrt.Name)
+	assert.NotContains(t, statuses, buildA, "WRT status entry must be pruned once the rendered resource is deleted")
+	assert.Contains(t, statuses, buildB, "WRT status entry for the live build must be retained")
```
jaypipes: Change `assert`s to `require`s

internal/controller/execplan_test.go
```diff
@@ -0,0 +1,363 @@
...
+		CurrentVersion: &temporaliov1alpha1.CurrentWorkerDeploymentVersion{BaseWorkerDeploymentVersion: baseVersion(buildB, depB, temporaliov1alpha1.VersionStatusCurrent)},
+	}
+	p := runPlanCycle(t, r, twd, connection, steadyStatus)
+	require.Len(t, p.ApplyWorkerResources, 1)
+	require.Empty(t, p.DeleteWorkerResources)
+
+	after := &temporaliov1alpha1.WorkerResourceTemplate{}
+	require.NoError(t, r.Get(context.Background(), types.NamespacedName{Namespace: namespace, Name: wrt.Name}, after))
+	assert.Equal(t, before.ResourceVersion, after.ResourceVersion, "no-op cycle must not write WRT status")
+}
```
jaypipes, line 362: 👍 good test.

internal/planner/planner.go
```diff
@@ -319,7 +337,34 @@ func getDeleteWorkerResources(
...
+		// deduplicated (a build sunset this cycle usually still has a status entry).
+		buildIDs := make([]string, 0, len(deletingBuildIDs)+len(wrt.Status.Versions))
+		seen := make(map[string]struct{}, len(deletingBuildIDs)+len(wrt.Status.Versions))
 		for _, buildID := range deletingBuildIDs {
+			if _, dup := seen[buildID]; dup {
+				continue
+			}
+			seen[buildID] = struct{}{}
+			buildIDs = append(buildIDs, buildID)
+		}
```
jaypipes: This is an awfully verbose way of doing [`slices.Compact`](https://pkg.go.dev/slices#Compact):

```go
buildsIDs := slices.Compact(deletingBuildIDs)
```

internal/planner/planner.go
```diff
@@ -319,7 +337,34 @@ func getDeleteWorkerResources(
...
+			if _, dup := seen[v.BuildID]; dup {
+				continue
+			}
+			if k8sState != nil {
+				if _, live := k8sState.Deployments[v.BuildID]; live {
+					continue // build still has a Deployment; its resource is managed by applies
+				}
+			}
+			seen[v.BuildID] = struct{}{}
+			buildIDs = append(buildIDs, v.BuildID)
```
jaypipes: and here, just do:

```go
buildIDs = append(buildIDs, v.BuildID)
```

and then de-dupe after the for loop:

```go
buildIDs = slices.Compact(buildIDs)
```

much easier to read...

jaypipes (changes requested):
@gibbonjj thanks very much for this PR. solves a real problem and has good test cases included. ++

I have some minor requests inline to take a look at. Fix those up and I'll approve and merge this tomorrow.

internal/controller/execplan.go
gibbonjj: Added in cb2df07: the docstring covers what the returned refs are (deletes confirmed this cycle, NotFound included), how executePlan uses them to prune the matching WRT status entries, and that rendered-resource delete failures are logged rather than returned — so the returned slice can be partial even with a nil error.

internal/controller/execplan_test.go
gibbonjj: Converted all of these to `require.*` in cb2df07 — the file no longer uses `assert` at all, so the import is gone too.

internal/controller/execplan_test.go
gibbonjj: Restructured to the flat `require.True` + `require.Equal` form in cb2df07.

internal/controller/execplan_test.go
gibbonjj: Converted to `require.*` in cb2df07.

internal/planner/planner.go
gibbonjj: Switched to the `slices` package in cb2df07, with one adjustment to the suggestion: `slices.Compact` only collapses *adjacent* equal elements, and the duplicates in this union are typically non-adjacent — a sunset build enters via `deletingBuildIDs` first and its surviving status entry is appended later in the second loop. So both sources plain-append into one slice and a single `slices.Sort` + `slices.Compact` dedups it at the end. Ref order carries no meaning (the deletes are independent), and the existing `status entry for deployment being deleted produces exactly one ref` subtest covers the dedup behaviour.

internal/planner/planner.go
gibbonjj: Done in cb2df07 — plain `append` in both loops, then one `slices.Sort` + `slices.Compact` after (Sort first because Compact only removes adjacent duplicates; details in the thread above).

jaypipes (approved):
Thanks @gibbonjj (and Claude) :)

Issue #61, opened 2025-06-25
add URL field in the version metadata


Shivs11:
- This can be done so that we can encode workflow URLs as part of the version metadata which, in turn, can be rendered on the UI so that users can easily view the workflows on which their version is failing!

carlydf:
We also want to consider a Version URL field templating approach for the controller when we do this (#51 )

Issue #50, opened 2025-06-25
Surface "Unhealthy" version in UI / k8s resource info after rollout acceptance test or worker healthcheck fails


carlydf:
Open questions:
- How to differentiate between pre-ramp unhealth, ramp unhealth, and general worker health during any version stage.

carlydf:
We should surface worker healthcheck error and Gate workflow failure in the TemporalWorkerDeployment events for sure. Rollout acceptance testing (during ramp phase) is post-GA, so I'm putting this as post-GA

PR #448, opened 2026-07-21
feat(controller): surface poller and gate-workflow health as conditions and events

### LINKED ISSUE #447: Surface Temporal-native poller health as Conditions + Events (extends #50)
**Is your feature request related to a problem? Please describe.**

Investigating #50 (surface "Unhealthy" version after rollout acceptance test / worker healthcheck fails), I found that today's per-version `HealthySince` field (`api/v1alpha1/workerdeployment_types.go`) is computed only from Kubernetes Deployment readiness — `internal/controller/state_mapper.go` calls `k8s.IsDeploymentHealthy(deployment)`, which checks pod/replica status. It has no visibility into whether Temporal is actually receiving polls from those pods. A Deployment can look perfectly healthy to Kubernetes (all replicas Ready) while its workers are misconfigured, stuck, or unable to reach Temporal — and today the controller has no way to surface that.

Separately, the controller already fetches real poller data from Temporal: `internal/temporal/worker_deployment.go`'s `getPollers()` calls `client.DescribeTaskQueue(...)` and gets back the actual poller list per task queue. Today this is used only to compute one internal boolean (`AllTaskQueuesHaveUnversionedPoller`, a versioning-migration heuristic) — the poller data itself is discarded, never surfaced to the user.

@carlydf's comment on #50 confirms this is in scope now (not gated on the post-GA rollout-acceptance-testing work):

> We should surface worker healthcheck error and Gate workflow failure in the TemporalWorkerDeployment events for sure.

**Describe the solution you'd like**

- Extend `VersionInfo` with a poller-health summary derived from the poller data already fetched in `getPollers()` (no new Temporal API calls needed for the base case).
- Add a new condition type on `WorkerDeploymentStatus` (following the existing `Reason*` constant pattern already in `workerdeployment_types.go`), e.g. `ConditionWorkersHealthy`, with reasons:
  - `ReasonPollersHealthy` — all expected task queues have active pollers
  - `ReasonNoActivePollers` — a version has zero pollers on one or more task queues
  - `ReasonPollerStatusUnknown` — couldn't determine (e.g. transient API error) — should not regress to "unhealthy" on a fetch failure
- Emit K8s Events on health-state transitions (not every reconcile, to avoid event spam) using the existing recorder pattern (`r.Recorder.Eventf(workerDeploy, corev1.EventTypeWarning/Normal, reason, "%s", message)`, already used in `worker_controller.go`/`execplan.go`):
  - `Warning` when a version loses all pollers on a task queue
  - `Normal` when a version's pollers become healthy again
- Gate workflow failure surfacing (explicitly called out by @carlydf above): the controller already tracks `TestWorkflows []WorkflowExecution` with status including `Failed`/`Canceled`/`Terminated`/`TimedOut`. Emit a `Warning` event referencing the workflow ID when a test/gate workflow transitions into one of these terminal-failure states.

This is purely additive — new condition type, new event emissions, one new struct field — with no change to rollout/reconciliation decision logic (`internal/planner/planner.go`). Health reporting, not health gating; the post-GA rollout-acceptance-testing work in #50 is a separate, later step that could consume this data but isn't required for it.

**Describe alternatives you've considered**

- Waiting for the post-GA rollout-acceptance-testing feature in #50 to land first — but @carlydf's comment explicitly separates worker-healthcheck/Gate-failure surfacing from that post-GA work, so this can land independently now.
- Relying only on the existing k8s-native `HealthySince` — insufficient, since it can't detect a worker that's running but not actually polling Temporal.

**Additional context**

Also sets up naturally for #61 (add URL field to version metadata) — once real poller/Gate-failure diagnostics exist, encoding workflow URLs to link directly to the failing execution becomes a much smaller follow-up. Not attempting #61's CRD/metadata change in this issue, since it's separately milestoned Post-GA.

Happy to contribute a PR for this if the team is open to the approach.

wankhede04:
## What was changed?

- `internal/temporal/worker_deployment.go`: extend `VersionInfo` with `PollerHealth map[string]bool` (per-task-queue: has at least one poller) and `PollerHealthUnknown bool`, populated from poller data already fetched via `DescribeTaskQueue` (no new round trips for the target version's task queues; one additional `DescribeVersion` call for the current version, since nothing previously fetched its task queues).
- `api/v1alpha1/workerdeployment_types.go`: add `ReasonActivePollers` and `ReasonPollerStatusUnknown`, and broaden the existing `ReasonWaitingForPollers` doc comment to also cover "current version has become current but has task queues with no active poller" (not just "not yet registered with Temporal").
- `internal/controller/workers_healthy.go` (new): a pure `computePollerHealthCondition(pollerHealth, unknown) (status, reason, affectedQueues)` function that decides, from a version's poller presence data, whether `ConditionProgressing` should report `ReasonActivePollers`, `ReasonWaitingForPollers` (naming the affected queues), or `ReasonPollerStatusUnknown`. A confirmed missing-poller queue always takes precedence over an unrelated fetch error.
- `internal/controller/worker_controller.go`: `syncConditions` now takes the fetched `temporalState` and, when the target version is `Current`, factors poller presence for the serving version (current version if set, else target) into `ConditionProgressing` using `computePollerHealthCondition`. `ConditionReady` is unaffected — it continues to reflect rollout completion only, matching its pre-existing (Kubernetes-readiness-only) semantics for backward compatibility. `setCondition` now returns whether the condition actually changed, used to emit a Warning/Normal event only on a poller-status transition (not every reconcile); no event is emitted for the `Unknown` reason since it isn't itself a state transition worth alerting on.

This is a smaller, more conservative shape than earlier revisions of this PR: poller health is folded into the existing `ConditionProgressing`/`ConditionReady` conditions rather than introduced as a new `ConditionWorkersHealthy` condition type, and the gate/test-workflow-failure event work has been split out into a separate PR (#457, already merged) to keep this change focused on poller health alone.

## Why?

Closes #447.

Today `ConditionReady`/`ConditionProgressing` are derived only from Kubernetes Deployment/pod readiness and Temporal rollout status — they have no visibility into whether Temporal is actually receiving polls. A Deployment can be fully Ready and its version fully rolled out (Current) while its workers are misconfigured, stuck, or unable to reach Temporal, and today the controller has no way to surface that. Meanwhile `internal/temporal/worker_deployment.go`'s `getPollers()` already fetches the real poller list via `DescribeTaskQueue`, but today it's reduced to a single internal bool (`AllTaskQueuesHaveUnversionedPoller`) used only for a rollout-strategy heuristic — the poller data itself is otherwise discarded.

This directly answers the non-post-GA part of #50: @carlydf's comment there states "We should surface worker healthcheck error and Gate workflow failure in the TemporalWorkerDeployment events for sure," explicitly separating this from the post-GA rollout-acceptance-testing work. The gate-workflow-failure half of that has already landed via #457; this PR covers the poller-health half.

#61 (add URL field to version metadata) is intentionally **out of scope** here — it's a separately milestoned Post-GA follow-up that this work sets up for but doesn't attempt.

## How was this tested?

- `gofmt -l .` — clean
- `go vet ./...` — clean
- `go build ./...` — succeeds
- `go test ./internal/controller/... ./internal/temporal/... ./internal/planner/... ./internal/k8s/...` — all pass, including:
  - A new table-driven unit test (`workers_healthy_test.go`) covering `computePollerHealthCondition`'s decision logic (nil/never-checked, all-healthy, one-queue-unhealthy, empty-map/no-queues-to-check, fetch-error-treated-as-unknown-not-unhealthy, and a confirmed-unhealthy-queue winning over an unrelated fetch error).
  - Updated `reconciler_events_test.go` cases covering `ConditionProgressing` transitions for a Current version with active pollers, no active pollers, and unknown poller status.
- `make test-unit`'s webhook (Ginkgo) suite could not be run in my environment because `helm` isn't installed there; confirmed this failure is pre-existing on unmodified `main` too (unrelated to this change) and is orthogonal to the reconciler/controller logic touched here.

## Risks

This is reporting-only: new condition reasons and new `VersionInfo` fields, with **no changes to rollout/reconciliation decision logic** (`internal/planner/planner.go` is untouched by this PR). The one behavior-affecting addition is a new `DescribeVersion` call for the current version each reconcile (needed since nothing previously fetched its task queues); a `DescribeTaskQueue` failure is treated as `Unknown`, never surfaced as an active poller problem, to avoid false alarms on transient errors.


api/v1alpha1/conditions.go
```diff
@@ -11,6 +11,12 @@ const (
 	// ConditionProgressing is True while a rollout is actively in-flight —
 	// i.e., the target version has not yet been promoted to current.
 	ConditionProgressing = "Progressing"
+
+	// ConditionWorkersHealthy indicates whether workers for the version currently
+	// serving production traffic (or, before the first rollout completes, the target
+	// version) are actively polling Temporal -- as opposed to merely being Ready at
+	// the Kubernetes level.
+	ConditionWorkersHealthy = "WorkersHealthy"
```
jaypipes: Is it necessary to create a new Condition type for this? How about adding `ReasonNoActivePollers` and `ReasonPollerStatusUnknown` but using the existing `ConditionReady` and adding the poller health check as part of the calculation of `ConditionReady` status of `True` for `WorkerDeployment`?

internal/controller/genstatus.go
```diff
@@ -42,6 +57,25 @@ func (r *WorkerDeploymentReconciler) generateStatus(
...
+			if !isGateWorkflowTerminalFailure(wf.Status) {
+				continue
+			}
+			if prevStatusByWorkflowID[wf.WorkflowID] == wf.Status {
+				continue
+			}
+			r.Recorder.Eventf(workerDeploy, corev1.EventTypeWarning, ReasonGateWorkflowFailed,
+				"Gate/test workflow %s for version %s ended with status %s", wf.WorkflowID, targetBuildID, wf.Status)
+		}
+
```
jaypipes, line 78: I would support adding this code in a separate PR. It's technically orthogonal to the poller health check work.

internal/temporal/worker_deployment.go
```diff
@@ -333,6 +365,27 @@ func getPollers(ctx context.Context,
...
+	client temporalClient.Client,
+	tqs []temporalClient.WorkerDeploymentTaskQueueInfo,
+) (health map[string]bool, unknown bool) {
+	health = make(map[string]bool, len(tqs))
+	for _, tqInfo := range tqs {
+		pollers, err := getPollers(ctx, client, tqInfo)
+		if err != nil { //nolint:revive // TODO(carlydf): consider logging this error
+			unknown = true
+			continue
+		}
```
jaypipes: I suspect it would be wiser to just return the error here instead of continuing on to the next task queue. Considering that the error is most likely either going to be hitting a rate limit or a network partition/connectivity failure, making a call to DescribeTaskQueue directly after receiving either one of those errors is just going to exacerbate things.

jaypipes (changes requested):
Some good stuff in here, thank you @wankhede04 :) However, I'd love to see this split into two PRs, one that adds the warning event emission for gate workflow failures and another that adds the poller health checking.

wankhede04:
Thanks for the review @jaypipes! Addressed all three points:

1. **Split into two PRs** — split out the gate/test-workflow-failure event emission into #457. This PR (#448) now contains only the poller-health-checking work.
2. **Folded poller health into `ConditionReady` instead of a new condition type** — removed `ConditionWorkersHealthy`. `ReasonNoActivePollers` and `ReasonPollerStatusUnknown` are now set directly on `ConditionReady` (False/Unknown respectively) as part of its calculation for the `Current` version status, instead of a separate condition.
3. **Fail fast on `DescribeTaskQueue` errors** — `computePollerHealth` now returns immediately on the first error instead of continuing to the next task queue, so a rate-limit/connectivity issue doesn't get compounded by hitting every remaining task queue.

Re-verified `gofmt`, `go vet`, `go build`, `golangci-lint` (0 issues against main), and the full unit test suite (including updated/added cases in `TestSyncConditions`) after these changes.

api/v1alpha1/conditions.go
wankhede04: Good call — done. Removed `ConditionWorkersHealthy` entirely. `ReasonNoActivePollers` and `ReasonPollerStatusUnknown` are now set directly on the existing `ConditionReady` (False/Unknown respectively) as part of its calculation when the target version's rollout has otherwise completed (`VersionStatusCurrent`), instead of a separate condition type. See the updated `syncConditions` in `internal/controller/worker_controller.go`.

internal/controller/genstatus.go
wankhede04: Agreed, done. Split this out into a standalone PR: #457. This PR (#448) now contains only the poller-health-checking work — `isGateWorkflowTerminalFailure` and the associated event-emission code have been removed from `genstatus.go` here.

internal/temporal/worker_deployment.go
wankhede04: Makes sense, done. `computePollerHealth` now returns immediately on the first `getPollers`/`DescribeTaskQueue` error instead of continuing on to the remaining task queues. Task queues already checked before the error are still returned in the health map, and callers set `PollerHealthUnknown=true` on a non-nil error, so the 'unknown, not unhealthy' semantics are preserved.

api/v1alpha1/workerdeployment_types.go
```diff
@@ -143,6 +143,26 @@ const (
 
 	// Deprecated: Use ReasonRolloutComplete on ConditionReady instead.
 	ReasonConnectionHealthy = "ConnectionHealthy"
+
+	// ReasonPollersHealthy is used on ConditionReady=True (in place of
+	// ReasonRolloutComplete) and on the Normal Event emitted when a version's workers
+	// transition from having no active pollers (or unknown status) back to actively
+	// polling every known task queue.
+	ReasonPollersHealthy = "PollersHealthy"
```
jaypipes: Please change this to `ReasonActivePollers`. The actual health of the poller isn't known :) Just that it's actively polling the server.

Also, I have changed my mind about having this reason affect `ConditionReady`. Instead, let's make this reason only impact `ConditionProgressing=False` so that it matches the existing `ReasonWaitingForPollers` that is set on `ConditionProgressing=True`.

api/v1alpha1/workerdeployment_types.go
```diff
@@ -143,6 +143,26 @@ const (
...
+
+	// ReasonNoActivePollers is set on ConditionReady=False -- even though the target
+	// version has otherwise completed rollout and become current -- when one or more
+	// of the version's task queues have no active poller. A Deployment can be fully
+	// Ready at the Kubernetes level while its workers are misconfigured, stuck, or
+	// unable to reach Temporal; this reason distinguishes that case from a healthy
+	// rollout.
+	ReasonNoActivePollers = "NoActivePollers"
```
jaypipes: I'm wondering if we need `ReasonNoActivePollers` at all since we have the existing `ReasonWaitingForPollers`. thoughts?

api/v1alpha1/workerdeployment_types.go
wankhede04: Good call — done. Renamed `ReasonPollersHealthy` to `ReasonActivePollers`. Also moved this signal off `ConditionReady` entirely: poller presence for the current version is now surfaced only on `ConditionProgressing` (`True`/`ReasonWaitingForPollers` when a task queue has no active poller, `False`/`ReasonActivePollers` when all do), mirroring the existing `ReasonWaitingForPollers` pair rather than affecting `ConditionReady`. `ConditionReady=True/ReasonRolloutComplete` no longer depends on poller presence.

api/v1alpha1/workerdeployment_types.go
wankhede04: Agreed — removed `ReasonNoActivePollers` entirely. The missing-poller case now reuses the existing `ReasonWaitingForPollers` (set on `ConditionProgressing=True`), same as you suggested.

wankhede04:
Thanks @jaypipes — addressed both remaining points from your last review:

1. Renamed `ReasonPollersHealthy` → `ReasonActivePollers`, since (as you noted) we only know pollers are actively polling, not that they're otherwise "healthy".
2. Moved poller presence off `ConditionReady` entirely and onto `ConditionProgressing`: `ReasonWaitingForPollers` (Progressing=True) is now reused for the missing-poller case, and `ReasonActivePollers` (Progressing=False) mirrors it when all task queues have a poller. `ReasonNoActivePollers` has been removed, and `ConditionReady=True/ReasonRolloutComplete` no longer depends on poller presence.

Re-verified `gofmt`, `go vet`, `go build`, and the full `internal/controller`/`internal/temporal` test suite after these changes — all green.

This should be the last of the open feedback on this PR. Would appreciate a merge at your earliest convenience — happy to make any further adjustments if needed.

wankhede04:
@jaypipes Can u please merge this PR

jaypipes:
> @jaypipes Can u please merge this PR

@wankhede04 this PR is going into the v1.9.0 release series. I am working to cut the v1.8.1 release series today so I will merge this PR once that release is cut.

wankhede04:
Sounds good, thanks @jaypipes! No rush — appreciate you keeping this queued for right after the v1.8.1 cut. Let me know if there's anything else needed on my end in the meantime.

jaypipes:
> Sounds good, thanks @jaypipes! No rush — appreciate you keeping this queued for right after the v1.8.1 cut. Let me know if there's anything else needed on my end in the meantime.

@wankhede04 please do update the PR summary to make it accurate with all the changes made! :)

wankhede04:
@jaypipes done — updated the PR summary to accurately reflect the current diff (poller health folded into the existing `ConditionProgressing`/`ConditionReady` conditions rather than a new `ConditionWorkersHealthy` type, and noted that the gate/test-workflow-failure event work was split out into #457, already merged). Let me know if anything else needs adjusting!

internal/temporal/worker_deployment.go
```diff
@@ -228,6 +257,12 @@ func GetTestWorkflowStatus(
 		return nil, fmt.Errorf("unable to describe worker deployment version for buildID %q: %w", buildID, err)
 	}
 
+	// Poller health for the target version, computed from the task queues already
+	// fetched above via DescribeVersion (no additional per-version round trip).
+	var pollerErr error
+	temporalState.Versions[buildID].PollerHealth, pollerErr = computePollerHealth(ctx, client, versionResp.Info.TaskQueuesInfos) //nolint:revive // TODO(carlydf): consider logging this error
```
eniko-dif, line 263: I might be wrong, but I think this is not needed, since you only use the PollerHealth information when the version status is current, and in that case, you compute the poller health above.

internal/temporal/worker_deployment.go
```diff
@@ -38,6 +38,17 @@ type VersionInfo struct {
 	//   - Strategy is Progressive, and
 	//   - Presence of unversioned pollers in all task queues of target version cannot be confirmed.
 	AllTaskQueuesHaveUnversionedPoller bool
+
+	// PollerHealth summarizes, per task queue, whether at least one poller was observed.
+	// Keyed by task queue name; true = has at least one poller. A task queue is absent
+	// from the map if its poller status could not be determined (e.g. a transient
+	// DescribeTaskQueue error) -- see PollerHealthUnknown.
+	PollerHealth map[string]bool
```
eniko-dif, line 46: I think the name `PollerHealth` is not that descriptive as a name as it does not refer to the health of the workers. Based on the `computePollerHealthCondition` function, you are looking for task queues without pollers, so I'd simplify this to `TaskQueuesWithoutPollers []string` and only gather task queues of interest in `computePollerHealth`. (The functions could be renamed accordingly.)

internal/temporal/worker_deployment.go
```diff
@@ -38,6 +38,17 @@ type VersionInfo struct {
...
+
+	// PollerHealthUnknown is true if poller status could not be determined for one or
+	// more of this version's task queues. Callers must not interpret this as unhealthy;
+	// it means "don't know", not "broken".
+	PollerHealthUnknown bool
```
eniko-dif, line 51: Maybe this could be `TaskQueueDescribeError error` and either be nil or the error directly?

internal/temporal/worker_deployment.go
```diff
@@ -177,6 +188,24 @@ func GetWorkerDeploymentState(
...
+		// merely being Ready at the Kubernetes level. Only checked for the current
+		// version to avoid an extra DescribeVersion/DescribeTaskQueue round trip per
+		// version on every reconcile.
+		if versionInfo.Status == temporaliov1alpha1.VersionStatusCurrent {
+			currentDesc, descErr := deploymentHandler.DescribeVersion(ctx, temporalClient.WorkerDeploymentDescribeVersionOptions{
+				BuildID: version.DeploymentVersion.BuildId,
+			})
+			if descErr == nil {
+				var pollerErr error
+				versionInfo.PollerHealth, pollerErr = computePollerHealth(ctx, client, currentDesc.Info.TaskQueuesInfos) //nolint:revive // TODO(carlydf): consider logging this error
```
eniko-dif, line 202: Since this is a heavy operation, could this be checked less frequently by tracking the last task queue pollers check time? I think once every 5-10 minutes would be sufficient.

internal/temporal/worker_deployment.go
```diff
@@ -333,6 +368,31 @@ func getPollers(ctx context.Context,
...
+// (or a task queue missing from health) as unhealthy -- it means "don't know", not
+// "broken".
+func computePollerHealth(
+	ctx context.Context,
+	client temporalClient.Client,
+	tqs []temporalClient.WorkerDeploymentTaskQueueInfo,
+) (health map[string]bool, err error) {
+	health = make(map[string]bool, len(tqs))
+	for _, tqInfo := range tqs {
+		pollers, err := getPollers(ctx, client, tqInfo)
```
eniko-dif, line 387: Based on the comment, if the issue is potentially hitting rate-limits, wouldn't it make sense to pause for a short time between task queue describe calls?

jaypipes:
@wankhede04 just a friendly ping to respond to @eniko-dif 's review comments when you have a moment. I'm going to need to bump this into the milestone after tomorrow's v1.9.0 release.

jaypipes:
@wankhede04 ping on this one. Any chance you might be able to resolve the merge conflicts for this PR and re-push? I'd still love to get this into the next release.

jaypipes:
@wankhede04 Hi again! We've made a number of changes to internal/temporal/worker_deployment.go in the last couple releases that require this PR to be rebased and have the merge conflicts resolved. Would you like me to do that for you?

wankhede04:
> @wankhede04 Hi again! We've made a number of changes to internal/temporal/worker_deployment.go in the last couple releases that require this PR to be rebased and have the merge conflicts resolved. Would you like me to do that for you?

Rebased and force-pushed — thanks for flagging it. One thing worth calling out: the rebase initially dropped the gate-workflow-failure code from `genstatus.go`/`util.go`, because my old "split this out into #457" commit replayed cleanly against the new `main` and matched/deleted the copy #457 already merged there. Caught it before pushing and restored those two files to `main`'s version, so this PR is back to touching only the poller-health files. Verified `gofmt`/`go vet`/`go build`/tests all pass post-rebase.

internal/temporal/worker_deployment.go
wankhede04: Agreed this is worth throttling. Leaning toward a small in-memory cache in the reconciler (buildID → last-checked time), rather than persisting anything on the CR status — keeps this contained to `internal/controller` with no CRD schema change. Downside is it resets on controller restart/leader failover, but given the check is best-effort health signal rather than a correctness-critical value, that seems acceptable. Does 5 min sound like the right default, or would you rather it be configurable?

internal/temporal/worker_deployment.go
wankhede04: Given the fail-fast change (stop on first error rather than continuing) and now the throttling from your other comment, I think a proactive sleep between calls becomes redundant with reduced value — the throttling cuts total call volume, and fail-fast already limits the blast radius of a rate-limit hit. Would rather not add per-call latency to every reconcile unless you still see a gap this doesn't cover — let me know if I'm missing a scenario.

internal/temporal/worker_deployment.go
wankhede04: > Good catch, you're right. `GetTestWorkflowStatus` only runs when `targetBuildID != CurrentBuildID`, so the `PollerHealth` it computed here was never the version `syncConditions` reads back (that only looks at the Current version's `PollerHealth`). [Removed it](https://github.com/temporalio/temporal-worker-controller/pull/448/changes/3b3b72a0f46b9938c4f557bdaa54c74ff724120d) — no functional change, just drops a wasted `DescribeTaskQueue` round trip per task queue on every reconcile for a non-current target version.

internal/temporal/worker_deployment.go
wankhede04: > Agreed on both — [renamed](https://github.com/temporalio/temporal-worker-controller/pull/448/commits/5096b05835926a15f16bfe41d4edc184183e71c9) `PollerHealth map[string]bool` to `TaskQueuesWithoutPollers []string` and `PollerHealthUnknown bool` to `TaskQueueDescribeError error`, and simplified `computePollerHealth` to only collect the queues actually missing a poller instead of building a full per-queue map. One thing I kept from the original design: the old code distinguished "never checked" from "checked, all healthy" via nil-map vs. non-nil-empty-map. I preserved that on the slice (`computePollerHealth` always returns a non-nil slice on success, even for zero task queues), so that distinction doesn't get lost in the rename.

internal/controller/worker_controller.go
```diff
@@ -769,21 +774,68 @@ func (r *WorkerDeploymentReconciler) setCondition(
...
+		if twd.Status.CurrentVersion != nil {
+			buildID = twd.Status.CurrentVersion.BuildID
+		}
+		var taskQueuesWithoutPollers []string
+		var describeErr error
+		if temporalState != nil {
+			if versionInfo, exists := temporalState.Versions[buildID]; exists {
+				taskQueuesWithoutPollers = versionInfo.TaskQueuesWithoutPollers
+				describeErr = versionInfo.TaskQueueDescribeError
+			}
```
jaypipes: what happens if `buildID` is not found in `temporalState.Versions`?

internal/controller/worker_controller.go
```diff
@@ -769,21 +774,68 @@ func (r *WorkerDeploymentReconciler) setCondition(
...
+		}
+		progressingStatus, progressingReason, affectedQueues := computePollerHealthCondition(taskQueuesWithoutPollers, describeErr)
```
jaypipes: Are we computing the poller health condition? or are we computing the ConditionProgressing for the current/target version?

internal/controller/worker_controller.go
```diff
@@ -769,21 +774,68 @@ func (r *WorkerDeploymentReconciler) setCondition(
...
+				r.Recorder.Eventf(twd, corev1.EventTypeWarning, temporaliov1alpha1.ReasonWaitingForPollers,
+					"Version %s has no active pollers on task queue(s): %s", buildID, strings.Join(affectedQueues, ", "))
+			case temporaliov1alpha1.ReasonActivePollers:
+				r.Recorder.Eventf(twd, corev1.EventTypeNormal, temporaliov1alpha1.ReasonActivePollers,
+					"Version %s has active pollers on all known task queues", buildID)
+			case temporaliov1alpha1.ReasonPollerStatusUnknown:
+				// Don't emit an event for Unknown -- it just means "don't know yet",
+				// not a state transition worth alerting on.
+			}
+		}
```
jaypipes: This block of code from line 796 to 833 should probably be placed into a `setConditionProgressingWhenCurrent` helper function. That would make it easier to read this code and maintain long-term.

internal/controller/workers_healthy.go
```diff
@@ -0,0 +1,47 @@
...
+
+	temporaliov1alpha1 "github.com/temporalio/temporal-worker-controller/api/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// computePollerHealthCondition derives a ConditionProgressing status/reason from a
+// version's poller presence data, for use as part of the ConditionProgressing
+// calculation when the target version is Current (see syncConditions). It is a
+// pure function so the decision logic can be unit tested without an envtest
+// environment.
```
jaypipes: This docstring makes me think you should rename this function to `setConditionProgressingForCurrent` and move the code I highlighted above from `syncConditions` to this function.

internal/temporal/worker_deployment.go
```diff
@@ -134,6 +148,23 @@ func GetWorkerDeploymentState(
 			ctx, l, client, targetBuildID, strategy,
 			depHandle, routingConfig, version,
 		)
+
+		// Poller health for the current version reflects whether workers are actively
```
jaypipes: It's not really poller health that we are calculating here is it? We're just calculating the number of task queues that have no poller. Not sure that qualifies as "poller health"...

internal/temporal/worker_deployment.go
```diff
@@ -452,6 +483,35 @@ func getPollers(ctx context.Context,
 	return resp.GetPollers(), nil
 }
 
+// computePollerHealth reports the names of task queues, among tqs, that were
```
jaypipes: Recommend renaming this function to `getTaskQueuesWithNoPollers` since that's what it's doing...

internal/controller/worker_controller.go
wankhede04: Good question — if `buildID` isn't in `temporalState.Versions` (nil `temporalState`, or this version wasn't described this reconcile), `taskQueuesWithoutPollers`/`describeErr` stay at their zero values, which `progressingConditionFromTaskQueues` treats the same as "never checked" → `ReasonPollerStatusUnknown` — not healthy, not unhealthy, just unknown. Added an explicit comment at the call site so this isn't left to be inferred.

internal/controller/workers_healthy.go
wankhede04: 2-4 Done, with one adjustment: renamed to `setConditionProgressingForCurrent` and moved the `syncConditions` block there as you said, but kept a small private pure function underneath it (`progressingConditionFromTaskQueues`) for the actual status/reason decision — that's what the unit test (renamed `TestProgressingConditionFromTaskQueues`) exercises without needing a reconciler/recorder. Let me know if you'd rather it all live in one function.

internal/temporal/worker_deployment.go
wankhede04: Agreed, and these two go together — reworded the comment to talk about "task queues without a poller" instead of "poller health", and renamed `computePollerHealth` to `getTaskQueuesWithNoPollers` as suggested.

jaypipes (approved):
I'm happy with this. @carlydf @eniko-dif @jlegrone @Shivs11 would love a review on this since it's a significant addition.

jaypipes:
@wankhede04 we've had some issues with the CLAAssistant bot. :( Would you mind re-signing the CLA please? Thank you!

wankhede04:
> @wankhede04 we've had some issues with the CLAAssistant bot. :( Would you mind re-signing the CLA please? Thank you!

Hey @jaypipes, I have re-signed the CLA. thanks for the review!

internal/temporal/worker_deployment.go
eniko-dif: Sorry for the late reply, but yes, 5 minutes sounds good.

internal/temporal/worker_deployment.go
eniko-dif: What I've meant is that if the issue may come from calling getPollers many times in succession, a small wait in-between could help. Throttling the overall check would still create a burst here.

internal/temporal/worker_deployment.go
```diff
@@ -452,6 +483,35 @@ func getPollers(ctx context.Context,
...
+// must not interpret a non-nil error as unhealthy -- it means "don't know", not
+// "broken".
+func getTaskQueuesWithNoPollers(
+	ctx context.Context,
+	client temporalClient.Client,
+	tqs []temporalClient.WorkerDeploymentTaskQueueInfo,
+) (taskQueuesWithoutPollers []string, err error) {
+	taskQueuesWithoutPollers = make([]string, 0, len(tqs))
+	for _, tqInfo := range tqs {
+		pollers, err := getPollers(ctx, client, tqInfo)
```
eniko-dif, line 506: I've just noticed this, but shouldn't `getPollers` include Nexus task queues? @jaypipes was that intentionally left out before?

jaypipes: @eniko-dif I don't think it was intentionally left out, no... but we can fix in a future PR I think, since this one did not touch the `getPollers()` function.

internal/temporal/worker_deployment.go
jaypipes: @eniko-dif I'm not too concerned about this. as @wankhede04 points out, it fails fast if it hits a rate limiting error from Temporal API server. I'm not convinced adding additional code to do a retry with backoff just for this particular API call is beneficial.

jaypipes (approved):

internal/temporal/worker_deployment.go
eniko-dif: I don't mind a follow-up PR, but if the users have nexus, wouldn't this new code create warnings?

jaypipes: @eniko-dif we ended up making a decision to not merge this PR into yesterday's release (v1.11.0). @wankhede04 if you could address this issue (add Nexus operation to the getPollers() implementation) that would be great. We'll merge this as soon as we see that and it will go into the v1.12.0 release on Sept 30. Again, thanks so much for your patience!

wankhede04: @jaypipes Added the Nexus task queue case to `getPollers()` — it was falling through with a nil `DescribeTaskQueueResponse`, and `GetPollers()` on nil returns an empty slice, so any Nexus task queue was unconditionally reported as having no active poller regardless of its real state.

internal/temporal/worker_deployment_test.go
```diff
@@ -12,9 +12,13 @@ import (
 
 	"github.com/go-logr/logr/funcr"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	temporaliov1alpha1 "github.com/temporalio/temporal-worker-controller/api/v1alpha1"
 	deploymentpb "go.temporal.io/api/deployment/v1"
 	enumspb "go.temporal.io/api/enums/v1"
+	taskqueuepb "go.temporal.io/api/taskqueue/v1"
+	"go.temporal.io/api/workflowservice/v1"
+	temporalClient "go.temporal.io/sdk/client"
```
jaypipes: @wankhede04 golang-ci linter is saying this should be `temporalclient` instead of `temporalClient`...

wankhede04: @jaypipes fixed — renamed to `temporalclient` throughout the test file.

jaypipes (approved):
👍

wankhede04:
> 👍

@jaypipes Should we merge ?

jaypipes:
@wankhede04 merged :) thanks again for your patience and persistence in getting this over the line.

wankhede04:
> @wankhede04 merged :) thanks again for your patience and persistence in getting this over the line.

Glad to have a meaningful contribution done @jaypipes, thanks for the constant support!

PR #450, opened 2026-07-23
helm: use namespaced Role for manager-role when restrictWatchNamespaces is set

aldredb:
## Summary

When `rbac.restrictWatchNamespaces` is configured, the manager-role is currently emitted as a
`ClusterRole` — only the binding is namespace-scoped (via `RoleBinding`). Organizations with strict
RBAC policies push back on ClusterRoles even when they're only bound via RoleBinding, because
ClusterRoles are visible cluster-wide and can be bound by anyone with bind permissions.

This PR changes the chart so that when `restrictWatchNamespaces` is set:

- The manager-role is emitted as a **namespaced `Role`** (one per watched namespace) instead of a `ClusterRole`
- The per-namespace `RoleBinding` references `kind: Role` instead of `kind: ClusterRole`
- The two inherently cluster-scoped rules (`namespaces get`, `subjectaccessreviews create`) remain in the separate `manager-cluster-role` ClusterRole, which is unchanged

Default behavior (empty `restrictWatchNamespaces`) is unchanged — still `ClusterRole` + `ClusterRoleBinding`.

## What was verified

- `helm template` renders correctly for both default and restricted modes
- `helm lint` passes
- Multi-namespace `restrictWatchNamespaces` produces one Role + one RoleBinding per namespace
- Controller source code never references RBAC resources by name/kind at runtime
- All temporal.io CRDs are namespace-scoped, so namespaced Roles can grant permissions on them
- Rules completeness: only `namespaces get` and `subjectaccessreviews create` excluded from the namespaced Role (both covered by `manager-cluster-role`)
- Also tested that it works in my local environment


helm/temporal-worker-controller/templates/rbac.yaml
```diff
@@ -58,6 +58,117 @@ subjects:
...
+# create) are handled by the separate manager-cluster-role below.
+{{- range .Values.rbac.restrictWatchNamespaces }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  labels:
+    app.kubernetes.io/component: rbac
+    {{- include "temporal-worker-controller.labels" $ | nindent 4 }}
+  name: {{ $.Release.Name }}-{{ $.Release.Namespace }}-manager-role
```
jaypipes: Why is `{{ $.Release.Namespace }}-` included in the `name` field template? Shouldn't the name of the namespaced manager role always be `{{ $.Release.Name}}-manager-role` since the `namespace` field will be differentiating due to the range loop over `restrictWatchNamespaces`?

helm/temporal-worker-controller/templates/rbac.yaml
```diff
@@ -230,7 +342,7 @@ metadata:
   namespace: {{ . }}
 roleRef:
   apiGroup: rbac.authorization.k8s.io
-  kind: ClusterRole
+  kind: Role
   name: {{ $.Release.Name }}-{{ $.Release.Namespace }}-manager-role
```
jaypipes: And obviously you'd want to remove `{{ $.Release.Namespace }}-` from here...

jaypipes (changes requested):
Mostly excellent, but see inline for one issue to address. Thank you for the submission @aldredb!

aldredb:
@jaypipes I addressed your comments

jaypipes (approved):
👍 thank you @aldredb! changes look good to me. I'd like one more contributor to review before merging, though.

jaypipes (changes requested):
@aldredb OK, so there's an issue uncovered by the excellent @Shivs11.

We have this [hack/sync-rbac-rules.py](https://github.com/temporalio/temporal-worker-controller/blob/main/hack/sync-rbac-rules.py) script that is intended to keep the different RBAC rules that are [specified with kubebuilder code markers](https://github.com/temporalio/temporal-worker-controller/blob/7e13417f0f7009f637b0dd09e7c8ae3652729ac5/internal/controller/worker_controller.go#L110-L127) in sync with the `helm/temporal-worker-controller/templates/rbac.yaml` file's contents. As far as we can tell, that script will be broken by this PR -- or the script may end up just overwriting the changes manually introduced in this PR.

So, I'm going to put a hold on merging this particular PR until we can figure out a way forward on this. I suspect we may want to have this PR update the `hack/sync-rbac-roles.py` script to generate the Role (not ClusterRole) rules entries. For now, though, unfortunately this PR isn't going to make it into the v1.8.1 (chart 0.28.0) release series, sorry!

aldredb:
> @aldredb OK, so there's an issue uncovered by the excellent @Shivs11.
> 
> We have this [hack/sync-rbac-rules.py](https://github.com/temporalio/temporal-worker-controller/blob/main/hack/sync-rbac-rules.py) script that is intended to keep the different RBAC rules that are [specified with kubebuilder code markers](https://github.com/temporalio/temporal-worker-controller/blob/7e13417f0f7009f637b0dd09e7c8ae3652729ac5/internal/controller/worker_controller.go#L110-L127) in sync with the `helm/temporal-worker-controller/templates/rbac.yaml` file's contents. As far as we can tell, that script will be broken by this PR -- or the script may end up just overwriting the changes manually introduced in this PR.
> 
> So, I'm going to put a hold on merging this particular PR until we can figure out a way forward on this. I suspect we may want to have this PR update the `hack/sync-rbac-roles.py` script to generate the Role (not ClusterRole) rules entries. For now, though, unfortunately this PR isn't going to make it into the v1.8.1 (chart 0.28.0) release series, sorry!

`hack/sync-rbac-roles.py` is adjusted accordingly. 

Tested with
```
make manifests
git diff helm/temporal-worker-controller/templates/rbac.yaml # no drift
```

Kindly need your review. Hope this PR can still make it to the 0.28.0 chart release

helm/temporal-worker-controller/templates/rbac.yaml
```diff
@@ -226,12 +340,12 @@ metadata:
   labels:
     app.kubernetes.io/component: rbac
     {{- include "temporal-worker-controller.labels" $ | nindent 4 }}
-  name: {{ $.Release.Name }}-{{ $.Release.Namespace }}-manager-rolebinding
```
eniko-dif, line 229: I don't think the namespace should be dropped here. It might be rare, but two controller installations could act on the same namespace with different access rights. Also, manager-cluster-role, manager-cluster-rolebinding, connection-editor-role, workerdeployment-viewer-role all use the namespace in the naming.

I've seen that @jaypipes wanted to remove this, but I don't understand why?

helm/temporal-worker-controller/templates/rbac.yaml
```diff
@@ -226,12 +340,12 @@ metadata:
...
+  name: {{ $.Release.Name }}-manager-rolebinding
   namespace: {{ . }}
 roleRef:
   apiGroup: rbac.authorization.k8s.io
-  kind: ClusterRole
-  name: {{ $.Release.Name }}-{{ $.Release.Namespace }}-manager-role
```
eniko-dif, line 234: Same as above.

helm/temporal-worker-controller/templates/rbac.yaml
```diff
@@ -58,6 +58,119 @@ subjects:
...
+{{- if .Values.rbac.restrictWatchNamespaces }}
+# Namespace-scoped mode: the manager role is emitted as a namespaced Role in each
+# watched namespace. Cluster-scoped rules (namespaces get, subjectaccessreviews
+# create) are handled by the separate manager-cluster-role below.
+{{- range .Values.rbac.restrictWatchNamespaces }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  labels:
```
eniko-dif, line 70: nit: could you please add the labels to the ClusterRole too for consistency?

helm/temporal-worker-controller/values.yaml
```diff
@@ -42,10 +42,10 @@ rbac:
   create: true
   # restrictWatchNamespaces keeps the namespaces the controller watches and the
   # namespaces its RBAC grants access to in sync. Empty (the default) watches all
-  # namespaces with cluster-wide RBAC. When set, the controller watches only these
-  # namespaces (via WATCH_NAMESPACES) and is granted the manager role through a
-  # RoleBinding in each listed namespace, plus a minimal cluster-scoped role for the
-  # two grants that cannot be namespaced (namespaces get, subjectaccessreviews create).
+  # namespaces with cluster-wide RBAC (ClusterRole + ClusterRoleBinding). When set,
```
eniko-dif, line 45: nit: could you please update the `values.schema.json` file's description too? (line 118)

jaypipes (approved):
Awesome. Thank you so much @aldredb this looks great!

hack/sync-rbac-rules.py
```diff
@@ -1,65 +1,114 @@
...
-    updated = pattern.sub(r"\g<1>" + rules_text + r"\g<3>", content)
-    with open(path, "w") as f:
-        f.write(updated)
+    return pattern.sub(r"\g<1>" + replacement + r"\g<3>", content)
+
+
+def main():
+    rules_text = extract_rules_text(ROLE_YAML)
+
+    with open(HELM_RBAC) as f:
```
eniko-dif: I'd open the file with `r+` and keep the file handler until we write back the changes. It's probably really low chance, but technically this way we might write back changes to an already changed file.

eniko-dif (commented):
Could you please add a CI test in `.github/workflows/helm-validate.yml` with a `restrictWatchNamespaces` scenario?

helm/temporal-worker-controller/templates/rbac.yaml
jaypipes: @eniko-dif I think [they are already](https://github.com/temporalio/temporal-worker-controller/blob/7e13417f0f7009f637b0dd09e7c8ae3652729ac5/helm/temporal-worker-controller/templates/rbac.yaml#L341-L344).

helm/temporal-worker-controller/templates/rbac.yaml
jaypipes: > I don't think the namespace should be dropped here. It might be rare, but two controller installations could act on the same namespace with different access rights.

I don't actually think that would work :) Consider that both controllers would spin up and watch the exact same CRs in the same Kubernetes Namespace. How would we keep each controller from stepping on each other's toes when processing those CRs?

helm/temporal-worker-controller/templates/rbac.yaml
eniko-dif: I don't see it around lines 176 here though.

hack/sync-rbac-rules.py
aldredb: Done

helm/temporal-worker-controller/values.yaml
aldredb: Done

helm/temporal-worker-controller/templates/rbac.yaml
eniko-dif: They could manage different CRs in the same namespace if i'm not mistaken? In any case, if this is not something that we want to support, it should be blocked somewhere explicitly. In the meantime, I think it makes more sense to have the same pattern as other cluster resources.

helm/temporal-worker-controller/templates/rbac.yaml
aldredb: @jaypipes  I think that link points at the manager-rolebinding block — that one does have the labels. The one @eniko-dif flagged is the manager-role ClusterRole in the {{- else }} branch (rbac.yaml L174), which only has metadata.name:

```
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: {{ .Release.Name }}-{{ .Release.Namespace }}-manager-role
rules:
```

Worth noting this is pre-existing on main, not something this PR introduced — it's the only RBAC object in the template without the labels. Added them here since it's a one-liner.

helm/temporal-worker-controller/templates/rbac.yaml
jaypipes: @eniko-dif ack. ok, no need to let the perfect be the enemy of the good on this :) @aldredb can you restore the `{{ $Release.Namespace }}-` to the name please?

eniko-dif (approved):

jaypipes (approved):
👍 thank you for your patience on these revisions @aldredb

helm/temporal-worker-controller/templates/rbac.yaml
aldredb: Done 👍

helm/temporal-worker-controller/templates/rbac.yaml
aldredb: Restored

PR #452, opened 2026-07-23
clean up auth mode and secret name handling

jaypipes:
This commit is the first of a series that attempts to clean up the Connection CRD and handling of Connection CRs in the client pool and controller packages.

This moves the `AuthMode` type out of `internal/clientpool` and into the `api/v1alpha1` package but *does not* add a field to `ConnectionSpec` for auth mode. Instead, we add a `ConnectionSpec.AuthMode()` method that returns the calculate authentication mode after looking at the different fields of `ConnectionSpec`. Similarly, we add a
`ConnectionSpec.SecretName()` method that returns the name of the Kubernetes Secret that holds the authentication information for the Connection. We clean up code in clientpool and controller by now calling these methods on `ConnectionSpec`.

Finally, we split out a `ConnectionSpec.Validate()` method that returns an error if the spec isn't valid. This splitting of validation from getter logic is more future-proof.

### DIFF
--- a/api/v1alpha1/connection_types.go
+++ b/api/v1alpha1/connection_types.go
@@ -5,8 +5,20 @@
 package v1alpha1
 
 import (
+	"errors"
+
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// AuthMode is the mode for authenticating to a Temporal server.
+type AuthMode string
+
+const (
+	AuthModeTLS           AuthMode = "TLS"
+	AuthModeAPIKey        AuthMode = "API_KEY"
+	AuthModeNoCredentials AuthMode = "NO_CREDENTIALS"
+	// Add more auth modes here as they are supported
 )
 
 // NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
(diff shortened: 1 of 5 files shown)


api/v1alpha1/connection_types.go
```diff
@@ -59,6 +71,38 @@ type ConnectionSpec struct {
...
+	default:
+		return AuthModeNoCredentials
+	}
+}
+
+// SecretName extracts the secret name from the ConnectionSpec, returning an
+// error if the secret name is missing and the authentication mode requires it.
+func (s ConnectionSpec) SecretName() (string, error) {
+	authMode := s.AuthMode()
+	switch authMode {
```
eniko-dif: nit: this can be simplified by `switch mode := s.AuthMode(); mode`

eniko-dif (approved):

api/v1alpha1/connection_types.go
jaypipes: fixed :)

eniko-dif (approved):
Trying to approve again.

api/v1alpha1/connection_types.go
```diff
@@ -59,6 +71,37 @@ type ConnectionSpec struct {
...
+		return s.MutualTLSSecretRef.Name, nil
+	case AuthModeAPIKey:
+		if s.APIKeySecretRef == nil || s.APIKeySecretRef.Name == "" {
+			return "", errors.New("API key secret name is not set")
+		}
+		return s.APIKeySecretRef.Name, nil
+	default:
+		return "", nil
+	}
+}
```
jlegrone, line 118: Nit: I think it would be nicer to remove the error return value from this function and instead add a `Validate() error` method to `ConnectionSpec`. That way there is a clear separation of concerns between validation (eg. when the CR is created/updated) and the process of actually establishing a connection (there may also be validations we want to perform on `ConnectionSpec` that don't involve the secret name in the future). This method can then be refactored to `SecretName() string` or `SecretName() (string, bool)`.

jaypipes: @jlegrone agreed 100%. I'm going to work on that in following PRs in this series :)

jaypipes: @jlegrone alrighty, split out Validate() from SecretName(). Please let me know if that's what you're looking for! :)

jlegrone: Last change request, I promise 🫡 

Please update the validation function here as well: https://github.com/temporalio/temporal-worker-controller/blob/9b0ff0d313ea9d0e91c8f38455b7f022fcb62268/api/v1alpha1/workerdeployment_webhook.go#L74

jlegrone: Never mind, I forgot it's a separate CRD so we need another webhook. That makes sense for a followup PR. Thanks!

jlegrone (approved):

PR #457, opened 2026-07-24
feat(controller): emit a Warning event on gate/test workflow terminal failure

wankhede04:
## What was changed?

- `internal/controller/genstatus.go`: add `isGateWorkflowTerminalFailure` and emit a Warning event (`GateWorkflowFailed`) the first time a gate/test workflow for the target version is observed to have ended in a non-successful terminal state (`Failed`, `Canceled`, `Terminated`, `TimedOut`). The previous reconcile's recorded status is compared against the freshly-fetched status so the event only fires once per transition, not on every reconcile loop.
- `internal/controller/util.go`: add the `ReasonGateWorkflowFailed` event reason constant.

## Why?

Split out of #448 per [review feedback from @jaypipes](https://github.com/temporalio/temporal-worker-controller/pull/448#pullrequestreview-3255588939): *"I'd love to see this split into two PRs, one that adds the warning event emission for gate workflow failures and another that adds the poller health checking."* This PR is the gate-workflow-failure half; #448 now contains only the poller-health-checking work.

Also directly relates to #447 / #50: @carlydf's comment on #50 states *"We should surface worker healthcheck error and Gate workflow failure in the TemporalWorkerDeployment events for sure,"* calling out Gate workflow failure surfacing as in-scope independent of the post-GA rollout-acceptance-testing work.

## How was this tested?

- `gofmt -l .` — clean
- `go vet ./...` — clean
- `go build ./...` — succeeds
- `make GOLANGCI_LINT_FIX=false GOLANGCI_LINT_BASE_REV=main lint-code` — 0 issues
- `go test ./internal/controller/...` — all pass

## Risks

This is purely additive: one new event-reason constant and one new Warning event emission, with no changes to rollout/reconciliation decision logic or existing status fields.


jaypipes (approved):
👍 thanks @wankhede04 :)

PR #462, opened 2026-07-26
Fix/register workerdeployment webhook

### LINKED ISSUE #461: webhook.enabled=true rejects every WorkerDeployment with 404 — chart ships vworkerdeployment.kb.io but the controller binary never registers the handler

Roll out Temporal Worker Deployment Versioning for our .NET workers on AKS using the Worker Controller. We installed the controller + CRDs Helm charts (v0.27.0 / appVersion 1.8.0, CRDs-first) and, since the chart's values.yaml describes `webhook.enabled` as enabling "the optional WorkerDeployment validating webhook", we enabled it to get apply-time validation of our `WorkerDeployment` CRs.

### Describe the bug

With `webhook.enabled: true`, **every `WorkerDeployment` create/update is rejected** by admission, regardless of the CR's contents:

```
Error: UPGRADE FAILED: failed to create resource: Internal error occurred:
failed calling webhook "vworkerdeployment.kb.io": failed to call webhook:
the server could not find the requested resource
```

The chart installs a `ValidatingWebhookConfiguration` (`vworkerdeployment.kb.io`, `failurePolicy: Fail`) pointing at `/validate-temporal-io-v1alpha1-workerdeployment` — but the controller binary never registers a handler at that path, so the webhook server answers HTTP 404 and the API server rejects the resource. TLS/cert-manager is healthy: the WorkerResourceTemplate webhook on the **same** service, cert, and port works fine.

Root cause (from reading the source):

- `api/v1alpha1/workerdeployment_webhook.go` implements the validator (Progressive `rampPercentage` strictly increasing; `gate.input`/`gate.inputFrom` mutual exclusivity) and defaulter (sunset delays), with tests — and defines `SetupWebhookWithManager`.
- `cmd/main.go` registers only `NewWorkerResourceTemplateValidator(mgr).SetupWebhookWithManager(mgr)`. No non-test code calls the WorkerDeployment `SetupWebhookWithManager` on any tag from v1.0.0 through v1.8.0, nor on current `main`.
- History: in v1.4.0 the registration existed but was commented out (`// TODO(jlegrone): Enable the webhook after fixing TLS`); by v1.5.2 the commented block was deleted when the WRT webhook was wired in. The chart's `webhook.enabled` flag and the `vworkerdeployment.kb.io` template kept shipping. PR #293's description says "The optional TWD webhook remains to provide apply-time feedback to those who enable it" — but the released binaries never serve it.

Suggested fix: register the webhook in `cmd/main.go` next to the WRT validator, e.g.

```go
if err = (&temporaliov1alpha1.WorkerDeployment{}).SetupWebhookWithManager(mgr); err != nil {
    setupLog.Error(err, "unable to create webhook", "webhook", "WorkerDeployment")
    os.Exit(1)
}
```

(or remove the `webhook.enabled` flag + template from the chart until the handler ships, so enabling it can't brick all WorkerDeployment applies).

### Minimal Reproduction

1. Install cert-manager, then the two charts:
   ```bash
   helm install temporal-worker-controller-crds oci://docker.io/temporalio/temporal-worker-controller-crds --version 0.27.0 -n temporal-system --create-namespace
   helm install temporal-worker-controller oci://docker.io/temporalio/temporal-worker-controller --version 0.27.0 -n temporal-system \
     --set certmanager.install=false --set certmanager.enabled=true \
     --set webhook.enabled=true
   ```
2. Apply any valid `WorkerDeployment` CR (e.g. the one from the README quick start).
3. Admission fails deterministically with the 404 error above. `kubectl logs` on the manager shows no "Registering webhook" line for the workerdeployment path — only the workerresourcetemplate one.

### Environment/Versions

- OS and processor: AKS (Linux, x86-64 nodes)
- Temporal Version: server 1.29.7 (self-hosted); Temporal Worker Controller chart 0.27.0 / appVersion 1.8.0; .NET SDK (Temporalio) 1.13.0
- Are you using Docker or Kubernetes or building Temporal from source? Kubernetes (AKS), official OCI Helm charts, cert-manager for webhook TLS

### Additional context

- Reproduced identically on three clusters (dev/stage/prod) — deterministic, not a startup/TLS race: the failure mode is an HTTP 404 from a healthy, TLS-serving webhook server, not `connection refused` / `no endpoints` / x509 errors.
- Workaround: set `webhook.enabled=false` (the default). Spec validation added in #293 via CRD CEL rules still applies; only the two webhook-exclusive checks (ramp ordering, gate input exclusivity) are lost.
- Related: #62, #293, and the v1.4.0 `cmd/main.go` TODO comment referenced above

MalithaPerera90:
## What was changed

- Registered the `WorkerDeployment` admission webhook with the controller manager in `cmd/main.go`, next to the existing `WorkerResourceTemplateValidator` registration.
- Added integration tests that exercise the WorkerDeployment webhook through the real envtest admission path, using the `ValidatingWebhookConfiguration` rendered from the Helm chart with `--set webhook.enabled=true` (`api/v1alpha1/workerdeployment_webhook_integration_test.go`).
- Generalized the suite's chart-webhook loader (`loadWRTWebhookFromHelmChart` → `loadWebhookFromHelmChart(chartPath, webhookName, extraArgs...)`) so both webhook configurations are loaded from the chart.

## Why?

The Helm chart's `webhook.enabled` flag installs a `ValidatingWebhookConfiguration` (`vworkerdeployment.kb.io`) pointing at `/validate-temporal-io-v1alpha1-workerdeployment`, but no released binary (v1.0.0–v1.8.0, nor `main`) ever registered that handler. With the flag enabled, the webhook server answers HTTP 404 and — because the config uses `failurePolicy: Fail` — **every `WorkerDeployment` create/update is rejected**:

```
Internal error occurred: failed calling webhook "vworkerdeployment.kb.io":
failed to call webhook: the server could not find the requested resource
```

The handler itself has existed since v1.4.0 (`api/v1alpha1/workerdeployment_webhook.go`, with unit tests) — its registration was commented out pending TLS work (`// TODO(jlegrone): Enable the webhook after fixing TLS`) and the commented block was removed during the v1.5.x WorkerResourceTemplate work, while the chart kept shipping the flag and webhook template.

This change is low-risk:
- `webhook_suite_test.go` already performs this exact registration in the envtest manager, so the wired behavior was already tested — this aligns the production wiring with the tested path.
- With `webhook.enabled=false` (the default), no WorkerDeployment webhook config is installed, so the newly served path is never called — zero behavior change for default installs.
- The webhook server, port, and TLS serving cert already exist unconditionally for the always-on WorkerResourceTemplate webhook.

The new integration tests also close the gap that let this ship unnoticed: the suite previously installed only the WRT webhook config into envtest, so no test ever sent a `WorkerDeployment` through the kube-apiserver admission chain. (Note: they validate the chart-rendered config against the served path end-to-end, but cannot guard the `cmd/main.go` wiring itself — that would need an e2e test against the built binary.)

## Checklist

1. Closes #461 

2. How was this tested:

- New envtest integration specs (real kube-apiserver → webhook server admission requests, webhook config rendered from the Helm chart with `--set webhook.enabled=true`):
  - a valid `WorkerDeployment` is admitted
  - Progressive rollout steps with non-increasing `rampPercentage` are rejected on create and on update
  - a gate setting both `input` and `inputFrom` is rejected
- Full webhook suite passes locally: 15/15 specs (11 existing + 4 new).
- `go build ./...`, `go vet`, and the existing unit tests (`TestWorkerDeployment_*`) all pass.

3. Any docs updates needed?

No docs changes needed — this makes the behavior match what the chart's `values.yaml` already documents for `webhook.enabled` ("controls the optional WorkerDeployment validating webhook").

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -138,6 +138,10 @@ func main() {
 		setupLog.Error(err, "unable to create webhook", "webhook", "WorkerResourceTemplate")
 		os.Exit(1)
 	}
+	if err = (&temporaliov1alpha1.WorkerDeployment{}).SetupWebhookWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create webhook", "webhook", "WorkerDeployment")
+		os.Exit(1)
+	}
 	//+kubebuilder:scaffold:builder
 
 	if err := mgr.AddHealthzCheck("healthz", healthz.Ping); err != nil {
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -138,7 +138,7 @@ func main() {
 		setupLog.Error(err, "unable to create webhook", "webhook", "WorkerResourceTemplate")
 		os.Exit(1)
 	}
-	if err = (&temporaliov1alpha1.WorkerDeployment{}).SetupWebhookWithManager(mgr); err != nil {
+	if err := (&temporaliov1alpha1.WorkerDeployment{}).SetupWebhookWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create webhook", "webhook", "WorkerDeployment")
 		os.Exit(1)
 	}
--- a/api/v1alpha1/webhook_suite_test.go
+++ b/api/v1alpha1/webhook_suite_test.go
@@ -63,7 +63,11 @@ var _ = BeforeSuite(func() {
 	ctx, cancel = context.WithCancel(context.TODO())
 
 	By("bootstrapping test environment")
-	wrtWebhook := loadWRTWebhookFromHelmChart(filepath.Join("..", "..", "helm", "temporal-worker-controller"))
+	chartPath := filepath.Join("..", "..", "helm", "temporal-worker-controller")
+	wrtWebhook := loadWebhookFromHelmChart(chartPath, "vworkerresourcetemplate.kb.io")
+	// The WorkerDeployment webhook config is gated behind the chart's webhook.enabled flag,
+	// so render it the same way a real install with the flag on would.
+	wdWebhook := loadWebhookFromHelmChart(chartPath, "vworkerdeployment.kb.io", "--set", "webhook.enabled=true")
 	testEnv = &envtest.Environment{
 		CRDDirectoryPaths: []string{
 			// CRDs live in the crds chart's templates directory
@@ -71,7 +75,7 @@ var _ = BeforeSuite(func() {
 		},
 		ErrorIfCRDPathMissing: true,
 		WebhookInstallOptions: envtest.WebhookInstallOptions{
-			ValidatingWebhooks: []*admissionregistrationv1.ValidatingWebhookConfiguration{wrtWebhook},
+			ValidatingWebhooks: []*admissionregistrationv1.ValidatingWebhookConfiguration{wrtWebhook, wdWebhook},
 		},
 	}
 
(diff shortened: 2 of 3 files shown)


jaypipes:
@MalithaPerera90 please do sign the contributor license agreement (CLA) so we can proceed with this PR review. thank you! :)

jaypipes (approved):
👍 excellent PR and good integration tests, thank you @MalithaPerera90.

Note to reviewers: #364 is still an issue even with this PR.

jaypipes (approved):

cmd/main.go
```diff
@@ -138,6 +138,10 @@ func main() {
 		setupLog.Error(err, "unable to create webhook", "webhook", "WorkerResourceTemplate")
 		os.Exit(1)
 	}
+	if err = (&temporaliov1alpha1.WorkerDeployment{}).SetupWebhookWithManager(mgr); err != nil {
```
jlegrone: tiny nit (the above `err` assignments should be updated too, but not needed in this PR)

```suggestion
	if err := (&temporaliov1alpha1.WorkerDeployment{}).SetupWebhookWithManager(mgr); err != nil {
```

jlegrone (approved):
Excellent, thank you for the fix and the updated tests!

jlegrone (approved):

PR #466, opened 2026-07-27
helm: add rbac.createEndUserRoles toggle to skip optional ClusterRoles

aldredb:
## What

Adds `rbac.createEndUserRoles` (default: `true`) to control whether the chart creates the optional end-user editor/viewer ClusterRoles.

## Why

The chart creates 10 ClusterRoles granting end users edit/view access to the `temporal.io` custom resources — `connections`, `workerdeployments`, `workerresourcetemplates`, and their deprecated equivalents. These ship with no bindings; they exist so cluster admins can bind them as needed.

In environments with strict ClusterRole policies, unbound ClusterRoles draw review pushback even though they grant no access without an explicit binding. Right now the only way to avoid them is `rbac.create=false`, which also drops the roles the controller itself needs.

This is a follow-up to #450 (namespaced manager Role). Together the two let an operator run the controller with a minimal cluster-scoped footprint:

```
--set 'rbac.restrictWatchNamespaces={ns-a,ns-b}' --set rbac.createEndUserRoles=false
```

leaves only three ClusterRoles, all of them load-bearing: `metrics-reader`, `proxy-role`, `manager-cluster-role`.

## Changes

- `values.yaml` — new `rbac.createEndUserRoles` key, documented, defaults to `true`
- `templates/rbac.yaml` — guard around the 10 end-user ClusterRoles
- `values.schema.json` — schema entry for the new key
- `.github/workflows/helm-validate.yml` — template the chart with `createEndUserRoles=false`

Default behavior is unchanged.

## Verification

| Config | ClusterRoles rendered |
|---|---|
| default | 13 (3 functional + 10 end-user) |
| `createEndUserRoles=false` | 3 (`metrics-reader`, `proxy-role`, `manager-role`) |
| `createEndUserRoles=false` + `restrictWatchNamespaces` | 3 (manager perms as namespaced Roles) |
| `rbac.create=false` | 2 (`metrics-reader`, `proxy-role`) — unchanged |

- `helm lint --strict` passes with the toggle on and off
- Rendered output parses as valid YAML in both modes (no stray document separators from the conditional)
- `make manifests` leaves `templates/rbac.yaml` unchanged, so `helm-check-rbac` stays green

### DIFF
--- a/helm/temporal-worker-controller/templates/rbac.yaml
+++ b/helm/temporal-worker-controller/templates/rbac.yaml
@@ -371,6 +371,7 @@ subjects:
     name: {{ .Values.serviceAccount.name | default (printf "%s-service-account" .Release.Name) }}
     namespace: {{ .Release.Namespace }}
 {{- end }}
+{{- if ne (.Values.rbac.createEndUserRoles | toString) "false" }}
 ---
 # permissions for end users to edit connections (new name).
 apiVersion: rbac.authorization.k8s.io/v1
@@ -632,5 +633,6 @@ rules:
       - workerresourcetemplates/status
     verbs:
       - get
+{{- end }}
 ---
 {{- end }}
--- a/helm/temporal-worker-controller/templates/rbac.yaml
+++ b/helm/temporal-worker-controller/templates/rbac.yaml
@@ -371,7 +371,7 @@ subjects:
     name: {{ .Values.serviceAccount.name | default (printf "%s-service-account" .Release.Name) }}
     namespace: {{ .Release.Namespace }}
 {{- end }}
-{{- if ne (.Values.rbac.createEndUserRoles | toString) "false" }}
+{{- if .Values.rbac.createEndUserRoles }}
 ---
 # permissions for end users to edit connections (new name).
 apiVersion: rbac.authorization.k8s.io/v1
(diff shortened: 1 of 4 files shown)


helm/temporal-worker-controller/templates/rbac.yaml
```diff
@@ -371,6 +371,7 @@ subjects:
     name: {{ .Values.serviceAccount.name | default (printf "%s-service-account" .Release.Name) }}
     namespace: {{ .Release.Namespace }}
 {{- end }}
+{{- if ne (.Values.rbac.createEndUserRoles | toString) "false" }}
```
jlegrone: Why the type cast? `createEndUserRoles` is already a boolean. We can follow this example: https://helm.sh/docs/chart_best_practices/rbac/#using-rbac-resources

aldredb: Fixed

helm/temporal-worker-controller/values.yaml
```diff
@@ -47,6 +47,12 @@ rbac:
   # via a RoleBinding, plus a minimal cluster-scoped role for the two grants that
   # cannot be namespaced (namespaces get, subjectaccessreviews create).
   restrictWatchNamespaces: []
+  # createEndUserRoles controls whether the chart creates the optional ClusterRoles
+  # that grant end users permission to edit/view temporal.io custom resources
+  # (connections, workerdeployments, workerresourcetemplates). These roles have no
+  # bindings by default — they exist so cluster admins can bind them as needed.
+  # Set to false to skip creating them entirely
```
jlegrone:
```suggestion
  # Set to false to skip creating them entirely.
```

jlegrone (approved):

jaypipes (approved):
Nice :) thanks @aldredb

PR #467, opened 2026-07-27
fix(clientpool): attach namespace header to every request

### LINKED ISSUE #440: [Feature Request] attach temporal-namespace header to GetSystemInfo requests on client

Some users use a proxy that supports connecting to multiple different upstream temporal servers. A common convention for this has been to use the `temporal-namespace` header (which SDKs already attach for all namespaced requests), so that the proxy can route to the appropriate server.

However, `GetSystemInfo` is not a namespaced request, and the SDK does not attach this header by default.

The [KEDA scaler](https://github.com/kedacore/keda/blob/main/pkg/scalers/temporal_scaler.go#L384) does this via gRPC interceptor.

Set the `temporal-namespace` header on every outbound client request.
A more complete solution might be to allow users to set customizable gRPC headers on the connection (set through the `Connection` CRD).

Guflly:
## Summary

Attach the `temporal-namespace` header through the SDK header provider so it is included on every request, including `GetSystemInfo`. All authentication modes now share the same base client options.

Fixes #440.

## What was verified

- `go test ./internal/controller/clientpool -count=1`
- `go vet ./internal/controller/clientpool`

### DIFF
--- a/internal/controller/clientpool/clientpool_test.go
+++ b/internal/controller/clientpool/clientpool_test.go
@@ -104,6 +104,19 @@ func makeOpts(hostPort string) NewClientOptions {
 	}
 }
 
+func TestNewClientOptionsSetsTemporalNamespaceHeader(t *testing.T) {
+	opts := makeOpts("localhost:7233")
+	opts.TemporalNamespace = "routing-namespace"
+	clientOpts := newTestPool().newClientOptions(opts)
+
+	headers, err := clientOpts.HeadersProvider.GetHeaders(context.Background())
+
+	require.NoError(t, err)
+	assert.Equal(t, map[string]string{
+		"temporal-namespace": "routing-namespace",
+	}, headers)
+}
+
 func makeTLSSecret(certPEM, keyPEM, caPEM []byte) corev1.Secret {
 	data := map[string][]byte{
 		"tls.crt": certPEM,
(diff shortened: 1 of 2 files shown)


jaypipes:
Hi @Guflly thank you for your PR! Please do sign the contributor license agreement so we can review your submission ASAP. :)

internal/controller/clientpool/clientpool_test.go
```diff
@@ -104,6 +104,19 @@ func makeOpts(hostPort string) NewClientOptions {
 	}
 }
 
+func TestNewClientOptionsSetsTemporalNamespaceHeader(t *testing.T) {
+	opts := makeOpts("localhost:7233")
+	opts.TemporalNamespace = "routing-namespace"
+	clientOpts := newTestPool().newClientOptions(opts)
+
+	headers, err := clientOpts.HeadersProvider.GetHeaders(context.Background())
```
jlegrone: nit

```suggestion
	headers, err := clientOpts.HeadersProvider.GetHeaders(t.Context())
```

jlegrone (approved):

Guflly:
Signed. Thanks.

internal/controller/clientpool/clientpool_test.go
Guflly: Switched this to t.Context() and rebased onto current main.

jaypipes (approved):
Thanks @Guflly :)

PR #469, opened 2026-07-27
Skip unavailable deprecated CRD watches

### LINKED ISSUE #413: [Feature Request] opt out of watching deprecated crds

The `TemporalWorkerDeployment` and `TemporalConnection` CRDs were deprecated, but the worker controller currently still watches these resources (for backwards compatibility reasons). Newer users of the worker controller do not want to install these deprecated CRDs, and receive a bunch of errors like this when these deprecated CRDs aren't installed. I would like to have a flag exposed that opts out of these watch streams.

```
if kind is a CRD, it should be installed before calling Start
"no matches for kind \"TemporalConnection\" in version \"temporal.io/v1alpha1\""
"no matches for kind \"TemporalWorkerDeployment\" in version \"temporal.io/v1alpha1\""
```
 
Due to code like this:

https://github.com/temporalio/temporal-worker-controller/blob/main/cmd/main.go#L111
https://github.com/temporalio/temporal-worker-controller/blob/main/internal/controller/worker_controller.go#L886

Expose a configuration flag like `WATCH_DEPRECATED_CRDS` that when set to false (true could be the default value if not set), opts out of these watches. Naming can be flexible of course

Guflly:
## What changed

Checks at startup whether the deprecated `TemporalWorkerDeployment` and `TemporalConnection` resources can be listed and watched in each configured namespace.

If a resource is missing or access is forbidden, its deprecated reconciler and watch paths are skipped. The two resources are checked independently.

Fixes #413

## Tested

- `go test ./internal/controller -run TestDetectDeprecatedCRDWatches -count=1`
- `go test ./internal/controller -count=1`
- `go vet ./...`
- `go build ./cmd/...`
- `golangci-lint run --build-tags test_dep --new-from-rev=HEAD~ --config=.github/.golangci.yml`


cmd/main.go
```diff
@@ -54,6 +60,8 @@ func main() {
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	flag.BoolVar(&watchDeprecatedCRDs, "watch-deprecated-crds", watchDeprecatedCRDs,
+		"Watch the deprecated TemporalWorkerDeployment and TemporalConnection resources.")
```
jlegrone: Rather than introducing a new flag (which we will also need to deprecate & remove in subsequent releases), I think it would be preferable for the controller to automatically skip watching the deprecated CRDs if they aren't installed in the cluster or if it doesn't have RBAC bindings to allow watching them.

Can we do that check at startup instead of parsing a flag?

Guflly: Updated this to auto-detect both deprecated resources at startup. Each one is checked independently in every configured watch namespace; missing or forbidden resources are skipped, and probe watches are stopped immediately. I removed the flag and Helm changes and added tests for missing CRDs and denied list/watch permissions. The affected tests, vet, build, and lint pass.

internal/controller/deprecated_crd_watch.go
```diff
@@ -0,0 +1,86 @@
...
+		return DeprecatedCRDWatches{}, err
+	}
+	return detectDeprecatedCRDWatches(ctx, client, namespaces)
+}
+
+func detectDeprecatedCRDWatches(
+	ctx context.Context,
+	client dynamic.Interface,
+	namespaces []string,
+) (DeprecatedCRDWatches, error) {
```
jaypipes, line 32: You can probably just combine this into `DetectDeprecatedCRDWatches`. Also, a quick docstring stating what the returned value contains would be good.

jaypipes (approved):
Minor suggestion inline but nothing major. @Guflly thanks for this PR! Please don't forget to update the PR summary with an accurate description of the code after you changed it.

jaypipes (approved):
👍 thanks @Guflly :)

internal/controller/deprecated_crd_watch_test.go
```diff
@@ -0,0 +1,131 @@
...
+func TestDetectDeprecatedCRDWatchesUnexpectedError(t *testing.T) {
+	client := newDeprecatedCRDWatchClient()
+	client.PrependReactor("list", deprecatedTWDResource.Resource, func(k8stesting.Action) (bool, runtime.Object, error) {
+		return true, nil, errors.New("connection reset")
+	})
+
+	_, err := DetectDeprecatedCRDWatches(context.Background(), client, nil)
+	if err == nil {
+		t.Fatal("expected an error")
+	}
```
jlegrone: Nit: it would be safer for the test to assert that the error returned was the `connection reset` error from the test harness, not simply non-nil.

internal/controller/deprecated_crd_watch.go
```diff
@@ -0,0 +1,74 @@
...
+	twd, err := canListAndWatch(ctx, client, deprecatedTWDResource, namespaces)
+	if err != nil {
+		return DeprecatedCRDWatches{}, fmt.Errorf("check TemporalWorkerDeployment watch: %w", err)
+	}
+	tc, err := canListAndWatch(ctx, client, deprecatedTCResource, namespaces)
+	if err != nil {
+		return DeprecatedCRDWatches{}, fmt.Errorf("check TemporalConnection watch: %w", err)
```
jlegrone: Nit: this error text assumes it was the watch that failed, but it could have been the list call instead. I'd wrap the error from inside `canListAndWatch` instead so that it's clear which API call the error was from.

internal/controller/deprecated_crd_watch.go
```diff
@@ -0,0 +1,74 @@
...
+	resource schema.GroupVersionResource,
+	namespaces []string,
+) (bool, error) {
+	if len(namespaces) == 0 {
+		namespaces = []string{metav1.NamespaceAll}
+	}
+	for _, namespace := range namespaces {
+		resourceClient := client.Resource(resource).Namespace(namespace)
+		if _, err := resourceClient.List(ctx, metav1.ListOptions{Limit: 1}); err != nil {
+			if apierrors.IsNotFound(err) || apierrors.IsForbidden(err) {
```
jlegrone, line 59: Is `NotFound` definitely the error type returned if the CRD is not installed? Just double checking since I don't think we exercise this in an end to end test (I don't think an integration test is unnecessary, but I want to be sure a real k8s cluster behaves as expected).

jlegrone (approved):

internal/controller/deprecated_crd_watch_test.go
Guflly: Updated the test to check that the returned error wraps the exact connection reset error.

internal/controller/deprecated_crd_watch.go
Guflly: Moved the operation context into canListAndWatch, so the error now says whether list or watch failed.

internal/controller/deprecated_crd_watch.go
Guflly: Yes. The dynamic client sends the GVR directly, and an unserved resource endpoint returns 404. client-go maps that to StatusReasonNotFound, so apierrors.IsNotFound handles a missing CRD. A server availability failure remains an error.

eniko-dif (approved):

PR #497, opened 2026-08-03
feat: allow RollingUpdate to be specified for k8s Deployment

### LINKED ISSUE #496: Expose Deployment rollingUpdate strategy (maxUnavailable/maxSurge) on TemporalWorkerDeployment
## Summary

The controller creates and manages the child `Deployment` for each worker version, but there's no way to configure that Deployment's rolling-update strategy. Since the controller never sets `spec.strategy`, owned Deployments fall back to the Kubernetes default (`RollingUpdate`, `maxUnavailable`/`maxSurge` 25%/25%). On large fleets this makes in-place rolling restarts of a Current version disruptive: up to ~25% of pods can go unavailable at once, which spikes `schedule_to_start`.

Users would like a field on the `WorkerDeployment` spec to set `maxUnavailable`/`maxSurge`, which the controller then applies to the Deployments it owns and reconciles.

## Current behavior

`TemporalWorkerDeploymentSpec` exposes `replicas`, `template`, `minReadySeconds`, `progressDeadlineSeconds`, `rollout`, `sunset`, and `workerOptions` — there is no Deployment-level `strategy` field. `rollout.strategy` (Manual/AllAtOnce/Progressive) controls Temporal traffic routing across versions, not how pods roll within a single version's Deployment. The builder in `internal/k8s/deployments.go` doesn't set `spec.strategy` on the owned Deployment, so it inherits the cluster default of 25%/25%.

## Why this matters

Users may periodically do in-place rolling restarts of Current workers with the same build ID (e.g. non-workflow config/bug-fix changes — the `UnsafeCustomBuildID` path, plus ordinary pod-template drift on the current version). In those cases the restart happens on the *same* Deployment, so its own `strategy.rollingUpdate` is the only thing governing availability. With 25% of a large fleet cycling at once, they lose enough pollers to build a backlog. They'd like to declare something conservative like `maxUnavailable: 5%` and have it stick.

> Filing on behalf of a Temporal Cloud customer running large worker fleets in production.

otan:
## What

Adds an optional `spec.deploymentStrategy` field on `WorkerDeployment` that mirrors `apps/v1` `Deployment.spec.strategy`. The controller sets it on owned versioned Deployments on create/update and reconciles drift so settings like `maxUnavailable` / `maxSurge` stick.

This is distinct from `spec.rollout.strategy`, which controls Temporal traffic routing across versions.

RollingUpdate settings mainly matter for in-place pod rolls on an existing version Deployment (for example `kubectl rollout restart`, or a pod-template change while `unsafeCustomBuildID` stays the same). The common new-build-ID path creates a fresh Deployment scaled up from zero, so those settings have little practical effect there even though the field is still written onto the object.

## Why

Owned Deployments currently inherit the Kubernetes default rolling update (`25%` / `25%`). On large fleets, in-place restarts of a Current version (same build ID) can take too many pollers offline at once and spike `schedule_to_start`. Users need a conservative strategy such as `maxUnavailable: 5%`.

## Testing

- `make generate manifests`
- `go test ./internal/k8s/ ./internal/planner/ -run 'Strategy|ApplyDeploymentStrategyDefaults|ReplicasNilPreserved'`

Resolves #496

### DIFF
--- a/api/v1alpha1/workerdeployment_types.go
+++ b/api/v1alpha1/workerdeployment_types.go
@@ -87,11 +87,7 @@ type WorkerDeploymentSpec struct {
 	// maxUnavailable/maxSurge 25%).
 	//
 	// This is distinct from spec.rollout.strategy, which controls Temporal
-	// traffic routing across worker versions. It also does not apply when the
-	// controller creates a new build ID / Deployment for a new worker version.
-	// It matters for in-place pod replacement on an existing version Deployment
-	// (for example kubectl rollout restart, or a pod-template change while
-	// spec.workerOptions.unsafeCustomBuildID stays the same).
+	// traffic routing across worker versions.
 	// +optional
 	DeploymentStrategy *appsv1.DeploymentStrategy `json:"deploymentStrategy,omitempty"`
 
--- a/api/v1alpha1/workerdeployment_types.go
+++ b/api/v1alpha1/workerdeployment_types.go
@@ -5,6 +5,7 @@
 package v1alpha1
 
 import (
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -79,6 +80,20 @@ type WorkerDeploymentSpec struct {
 	// not be estimated during the time a deployment is paused. Defaults to 600s.
 	// +kubebuilder:default=600
 	ProgressDeadlineSeconds *int32 `json:"progressDeadlineSeconds,omitempty" protobuf:"varint,9,opt,name=progressDeadlineSeconds"`
+
+	// DeploymentStrategy describes how to replace Pods within each versioned
+	// Deployment this controller owns. Mirrors apps/v1 Deployment.spec.strategy.
+	// When omitted, Kubernetes defaults apply (RollingUpdate with
+	// maxUnavailable/maxSurge 25%).
+	//
+	// This is distinct from spec.rollout.strategy, which controls Temporal
+	// traffic routing across worker versions. It also does not apply when the
+	// controller creates a new build ID / Deployment for a new worker version.
+	// It matters for in-place pod replacement on an existing version Deployment
+	// (for example kubectl rollout restart, or a pod-template change while
+	// spec.workerOptions.unsafeCustomBuildID stays the same).
+	// +optional
+	DeploymentStrategy *appsv1.DeploymentStrategy `json:"deploymentStrategy,omitempty"`
 
 	// How to rollout new workflow executions to the target version.
 	RolloutStrategy RolloutStrategy `json:"rollout"`
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -348,6 +348,32 @@ Gate workflow input details:
 
 ## Advanced Configuration
 
+### Deployment Rolling Update Strategy
+
+Controls how Pods are replaced **within a single versioned Kubernetes `Deployment`**. It does **not** affect Temporal traffic routing across worker versions (`spec.rollout.strategy`), and it does **not** apply when the controller creates a new version.
+
+By default the controller derives a new build ID from the pod template. A new build ID creates a **new** Deployment object, so pod replacement for that rollout is just scaling up the new Deployment (and later scaling down/deleting the old one). In that common path, `deploymentStrategy` has no effect.
+
+`deploymentStrategy` matters only when pods roll on an **existing** version Deployment, for example:
+
+- You run `kubectl rollout restart` (or equivalent) against a specific versioned Deployment.
+- You set `spec.workerOptions.unsafeCustomBuildID` and change the pod template without changing that build ID — the controller updates that version's Deployment in place, which triggers a Kubernetes rolling update of its pods.
+- Connection-related fields change and the controller patches the existing Deployment's pod template (env vars / volumes), which also rolls pods in place.
+
+When omitted, Kubernetes defaults apply (`RollingUpdate` with `maxUnavailable`/`maxSurge` of `25%`). On large fleets, set a more conservative strategy if you rely on in-place restarts of Current workers:
+
+```yaml
+spec:
+  replicas: 100
+  deploymentStrategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 5%
+      maxSurge: 0
+  rollout:
+    strategy: Progressive
+```
+
 ### Environment-Specific Configurations
 
 **Production Configuration:**
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -350,15 +350,7 @@ Gate workflow input details:
 
 ### Deployment Rolling Update Strategy
 
-Controls how Pods are replaced **within a single versioned Kubernetes `Deployment`**. It does **not** affect Temporal traffic routing across worker versions (`spec.rollout.strategy`), and it does **not** apply when the controller creates a new version.
-
-By default the controller derives a new build ID from the pod template. A new build ID creates a **new** Deployment object, so pod replacement for that rollout is just scaling up the new Deployment (and later scaling down/deleting the old one). In that common path, `deploymentStrategy` has no effect.
-
-`deploymentStrategy` matters only when pods roll on an **existing** version Deployment, for example:
-
-- You run `kubectl rollout restart` (or equivalent) against a specific versioned Deployment.
-- You set `spec.workerOptions.unsafeCustomBuildID` and change the pod template without changing that build ID — the controller updates that version's Deployment in place, which triggers a Kubernetes rolling update of its pods.
-- Connection-related fields change and the controller patches the existing Deployment's pod template (env vars / volumes), which also rolls pods in place.
+Controls how Pods are replaced **within a single versioned Kubernetes `Deployment`** via `Deployment.spec.strategy`. It does **not** affect Temporal traffic routing across worker versions (see [Rollout Strategies](#rollout-strategies)).
 
 When omitted, Kubernetes defaults apply (`RollingUpdate` with `maxUnavailable`/`maxSurge` of `25%`). On large fleets, set a more conservative strategy if you rely on in-place restarts of Current workers:
 
--- a/internal/k8s/deployments.go
+++ b/internal/k8s/deployments.go
@@ -302,8 +302,29 @@ func NewDeploymentWithOwnerRef(
 				Spec: *podSpec,
 			},
 			MinReadySeconds: spec.MinReadySeconds,
+			Strategy:        DesiredDeploymentStrategy(spec),
 		},
 	}
+}
+
+// DesiredDeploymentStrategy returns the Deployment strategy to apply from the
+// WorkerDeployment spec. When spec.DeploymentStrategy is nil, returns the zero
+// value so Kubernetes defaults apply. When set, unset subfields are filled with
+// the same defaults the Deployment API server uses so create/update/reconcile
+// stay stable. Defensive against webhook-disabled installs where CR defaulting
+// may not have run.
+func DesiredDeploymentStrategy(spec *temporaliov1alpha1.WorkerDeploymentSpec) appsv1.DeploymentStrategy {
+	if spec == nil || spec.DeploymentStrategy == nil {
+		return appsv1.DeploymentStrategy{}
+	}
+	return *temporaliov1alpha1.DefaultDeploymentStrategy(spec.DeploymentStrategy)
+}
+
+// ApplyDeploymentStrategyDefaults fills Deployment strategy fields the same way
+// the apps/v1 Deployment defaulter does, so comparisons against live objects do
+// not flap on omitted Type / maxUnavailable / maxSurge. The input is not mutated.
+func ApplyDeploymentStrategyDefaults(s appsv1.DeploymentStrategy) appsv1.DeploymentStrategy {
+	return *temporaliov1alpha1.DefaultDeploymentStrategy(&s)
 }
 
 // TODO (Shivam): Change hash when secret name is updated as well.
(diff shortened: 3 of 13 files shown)


jaypipes:
@otan thanks so much for this contribution! I'll review a little later today or tomorrow morning. Been looking into some bug fixes earlier this week.

api/v1alpha1/deprecated_temporalworkerdeployment_types.go
```diff
@@ -79,6 +80,14 @@ type TemporalWorkerDeploymentSpec struct {
 	// +kubebuilder:default=600
 	ProgressDeadlineSeconds *int32 `json:"progressDeadlineSeconds,omitempty" protobuf:"varint,9,opt,name=progressDeadlineSeconds"`
 
+	// Strategy describes how to replace Pods for each versioned Deployment this
+	// controller owns. Mirrors apps/v1 Deployment.spec.strategy. When omitted,
+	// Kubernetes defaults apply (RollingUpdate with maxUnavailable/maxSurge 25%).
+	// This is distinct from spec.rollout.strategy, which controls Temporal traffic
+	// routing across worker versions.
+	// +optional
+	Strategy *appsv1.DeploymentStrategy `json:"strategy,omitempty"`
```
jaypipes: Please remove this change to the deprecated TemporalWorkerDeployment CRD, since that definition is frozen and the controller no longer supports changes to this CRD.

api/v1alpha1/workerdeployment_types.go
```diff
@@ -80,6 +81,14 @@ type WorkerDeploymentSpec struct {
 	// +kubebuilder:default=600
 	ProgressDeadlineSeconds *int32 `json:"progressDeadlineSeconds,omitempty" protobuf:"varint,9,opt,name=progressDeadlineSeconds"`
 
+	// Strategy describes how to replace Pods for each versioned Deployment this
+	// controller owns. Mirrors apps/v1 Deployment.spec.strategy. When omitted,
+	// Kubernetes defaults apply (RollingUpdate with maxUnavailable/maxSurge 25%).
+	// This is distinct from spec.rollout.strategy, which controls Temporal traffic
+	// routing across worker versions.
+	// +optional
+	Strategy *appsv1.DeploymentStrategy `json:"strategy,omitempty"`
```
jaypipes: Let's call this `DeploymentStrategy` to make it a little clearer that this refers to the underlying Kubernetes Deployment's deployment strategy :)

internal/k8s/deployments.go
```diff
@@ -302,10 +303,45 @@ func NewDeploymentWithOwnerRef(
...
+			maxUnavailable := intstr.FromString("25%")
+			s.RollingUpdate.MaxUnavailable = &maxUnavailable
+		}
+		if s.RollingUpdate.MaxSurge == nil {
+			maxSurge := intstr.FromString("25%")
+			s.RollingUpdate.MaxSurge = &maxSurge
+		}
+	}
+	return s
+}
```
jaypipes, line 328: Note to future self... we should standardize the way we do defaulting for custom resources, probably using mutating admission webhooks and the `Defaulter` interface

jaypipes (changes requested):
@otan thanks for this! :) Code looks good to me except for a couple requests (see inline). Address those small issues and I'll aim to get this into the v1.9.0 release (set to be cut on August 18th)

otan:
all good!

jaypipes (approved):
Awesomesauce, thank you so much @otan :)

internal/planner/planner.go
```diff
@@ -569,6 +570,36 @@ func updateDeploymentWithPodTemplateSpec(
...
+	if spec.DeploymentStrategy == nil {
+		return nil
+	}
+	existingDeployment, exists := k8sState.Deployments[buildID]
+	if !exists {
+		return nil
+	}
+
+	desired := k8s.DesiredDeploymentStrategy(spec)
+	actual := k8s.ApplyDeploymentStrategyDefaults(existingDeployment.Spec.Strategy)
```
eniko-dif, line 596: Claude flagged this and I think it's right: 

ApplyDeploymentStrategyDefaults takes/returns appsv1.DeploymentStrategy by value, but RollingUpdate is a *RollingUpdateDeployment. Copying the struct doesn't copy what that pointer points to, so when RollingUpdate is non-nil with only one of MaxUnavailable/MaxSurge set, the defaulting inside the function writes through the shared pointer and mutates existingDeployment.Spec.Strategy.RollingUpdate directly — even though this call is meant to be a read-only normalization for comparison.

Repro: build a DeploymentStrategy with RollingUpdate.MaxUnavailable set and MaxSurge: nil, call ApplyDeploymentStrategyDefaults on it, then check the original — MaxSurge goes from nil to 25% on the original struct.

The DesiredDeploymentStrategy call site avoids this by deep-copying first (spec.DeploymentStrategy.DeepCopy()), but this one doesn't. Suggest either deep-copying existingDeployment.Spec.Strategy before defaulting here, or having ApplyDeploymentStrategyDefaults deep-copy RollingUpdate internally before mutating it.

Impact is muted for real cluster objects (apiserver usually pre-defaults on decode) but will bite with the fake client in tests or any partially-populated object, and it breaks the implicit "pure function" contract of the helper.

internal/k8s/deployments.go
```diff
@@ -302,10 +303,46 @@ func NewDeploymentWithOwnerRef(
...
+	if spec == nil || spec.DeploymentStrategy == nil {
+		return appsv1.DeploymentStrategy{}
+	}
+	return ApplyDeploymentStrategyDefaults(*spec.DeploymentStrategy.DeepCopy())
+}
+
+// ApplyDeploymentStrategyDefaults fills Deployment strategy fields the same way
+// the apps/v1 Deployment defaulter does, so comparisons against live objects do
+// not flap on omitted Type / maxUnavailable / maxSurge.
+func ApplyDeploymentStrategyDefaults(s appsv1.DeploymentStrategy) appsv1.DeploymentStrategy {
```
eniko-dif, line 326: `api/v1alpha1/workerdeployment_webhook.go` already has an established pattern for every spec related defaults and validation. Could you please move this there too? So more precisely:
- The default `DeploymentStrategy` could be added in `WorkerDeploymentSpec.Default()`. (with a default value in `internal/defaults/defaults.go`)
- You could add a validator alongside `validateRolloutStrategy` rejecting invalid config.

otan:
all done

jaypipes (approved):
Thank you @otan!

api/v1alpha1/workerdeployment_types.go
```diff
@@ -80,6 +81,14 @@ type WorkerDeploymentSpec struct {
 	// +kubebuilder:default=600
 	ProgressDeadlineSeconds *int32 `json:"progressDeadlineSeconds,omitempty" protobuf:"varint,9,opt,name=progressDeadlineSeconds"`
 
+	// DeploymentStrategy describes how to replace Pods for each versioned
+	// Deployment this controller owns. Mirrors apps/v1 Deployment.spec.strategy.
+	// When omitted, Kubernetes defaults apply (RollingUpdate with
+	// maxUnavailable/maxSurge 25%). This is distinct from spec.rollout.strategy,
+	// which controls Temporal traffic routing across worker versions.
+	// +optional
+	DeploymentStrategy *appsv1.DeploymentStrategy `json:"deploymentStrategy,omitempty"`
```
jlegrone: I have two concerns with this:
1. Having both `deploymentStrategy` and `rolloutStrategy` may be confusing.
2. So far we haven't committed to making the underlying deployment resource type part of the public API. Ie. we could switch to statefulsets or replicasets instead in the future. Ideally we should keep the worker spec agnostic to the underlying k8s resource types.

I also think it needs to be clarified in the documentation when `deploymentStrategy` would apply, since the controller computes a new build ID (and thus creates a fresh k8s deployment object) for each new worker version. For example does this only matter if you manually perform a rolling update via `kubectl rollout restart` for a specific version?

Do we actually need to support the `Recreate` deployment strategy? If not, what would it look like if we only exposed `maxUnavailable` as an option? I think my concerns would be resolved if the new config was named like `spec.maxUnavailablePerVersion` which doesn't tie us to deployments as the underlying resource type.

otan: > Having both deploymentStrategy and rolloutStrategy may be confusing.

>  For example does this only matter if you manually perform a rolling update via kubectl rollout restart for a specific version?

yep. we just have no control over `kubectl rollout restart` at the moment (and we do a manual HPA patch to get around it). i can amend that - i originally wanted to keep it to just `strategy` to meet the original k8s Worker definition but was told to change it. i'm happy to rename it to whatever makes you more comfortable.

> So far we haven't committed to making the underlying deployment resource type part of the public API. Ie. we could switch to statefulsets or replicasets instead in the future. Ideally we should keep the worker spec agnostic to the underlying k8s resource types.

AFAICT, this `deployment.strategy` is an option available to both statefulsets or replicasets

jaypipes:
@otan we merged a couple fixes in internal/planner/planner.go that have caused some conflicts for this patch. would you mind pushing a quick conflict-resolving patch? If you want me to do it, let me know, I just didn't want to step on your toes. :)

otan:
all done i think

api/v1alpha1/workerdeployment_types.go
```diff
@@ -80,6 +81,20 @@ type WorkerDeploymentSpec struct {
...
 	ProgressDeadlineSeconds *int32 `json:"progressDeadlineSeconds,omitempty" protobuf:"varint,9,opt,name=progressDeadlineSeconds"`
 
+	// DeploymentStrategy describes how to replace Pods within each versioned
+	// Deployment this controller owns. Mirrors apps/v1 Deployment.spec.strategy.
+	// When omitted, Kubernetes defaults apply (RollingUpdate with
+	// maxUnavailable/maxSurge 25%).
+	//
+	// This is distinct from spec.rollout.strategy, which controls Temporal
+	// traffic routing across worker versions. It also does not apply when the
+	// controller creates a new build ID / Deployment for a new worker version.
```
jaypipes: The PR summary states:

> The controller applies it to owned versioned Deployments on create/update

But here, the code comment says:

> It also does not apply when the controller creates a new build ID / Deployment for a new worker version.

@otan that seems to be a contradiction?

docs/configuration.md
```diff
@@ -348,6 +348,32 @@ Gate workflow input details:
 
 ## Advanced Configuration
 
+### Deployment Rolling Update Strategy
+
+Controls how Pods are replaced **within a single versioned Kubernetes `Deployment`**. It does **not** affect Temporal traffic routing across worker versions (`spec.rollout.strategy`), and it does **not** apply when the controller creates a new version.
```
jaypipes: Same comment about the PR summary seemingly contradicting this:

> and it does **not** apply when the controller creates a new version.

jaypipes (commented):
@otan I *think* the PR summary might need to be updated to reflect the reality of the patch. See inline...

docs/configuration.md
otan: hehe, sorry a bit of a kubernetes noob and this slid through me (deployment version vs temporal version so I thought they made sense) - updating accordingly.

otan: hmm, these comments are too verbose / detail oriented (AI justifying itself) - to cut the crap i'll just remove them

otan:
i'm about to nod off (sydney time), feel free to amend accordingly!

api/v1alpha1/workerdeployment_types.go
carlydf: > AFAICT, this deployment.strategy is an option available to both statefulsets or replicasets

Could you share what gives you that impression? I did some reading on both, and it seems that ReplicaSets (https://kubernetes.io/docs/concepts/workloads/controllers/replicaset/) don't have such a strategy at all; if you update the pod template of a ReplicaSet, the ReplicaSet won't change or replace any of its Pods that are already running with a different template. If, later on, the ReplicaSet wants more replicas, it will use the new pod template to create new pods (still not touching the existing pods).

StatefulSets (https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#update-strategies) have a rolling update strategy option which goes one-by-one by default. In k8s v1.35, there is a `maxUnavailable` field, but it is in beta.

carlydf: >  we just have no control over kubectl rollout restart at the moment (and we do a manual HPA patch to get around it).

Could you explain more what you mean by these two things? Curious about your use case (so we can try to support it without re-packaging `Deployment` fields if we can)

otan: I purely want to be able to do a `kubectl rollout restart deployment <name>` without having it restart up to 25% pods at a time. I have to manually patch my deployment to do this at the moment.

we ... currently have to do restarts regularly to clean up some in-app-worker-state that can get borked because, e.g. node memory leaks. forcing a new deployment to do this means we have extra pinned worker versions for no reason.

carlydf: thanks for elaborating! I was hung up on trying to understand how a "manual HPA patch" would help you, but patching the Deployments makes sense. And I agree folks should be able to do a rolling restart with conservative settings.

carlydf: Maybe a crazy idea: what if instead of exposing a new knob here, we just make the rolling update settings for all worker-controller-managed deployments conservative by default (ie. maxUnavailable=1)?

That is what the default `StatefulSet` update strategy does, fwiw

jaypipes: > > AFAICT, this deployment.strategy is an option available to both statefulsets or replicasets
> 
> Could you share what gives you that impression? I did some reading on both, and it seems that ReplicaSets (https://kubernetes.io/docs/concepts/workloads/controllers/replicaset/) don't have such a strategy at all; if you update the pod template of a ReplicaSet, the ReplicaSet won't change or replace any of its Pods that are already running with a different template. If, later on, the ReplicaSet wants more replicas, it will use the new pod template to create new pods (still not touching the existing pods).

Yeah, because ReplicaSets are kind of a hidden behind-the-scenes resource grouping mechanism. The Deployment is the thing that has multiple ReplicaSets that correspond to the unique hash of the Deployment's pod template spec so when referring to `deployment.strategy` it's really just referring to the `DeploymentSpec.strategy`...

jaypipes:
@otan Super sorry for the delay on this and I don't want you working on your weekend, so I'm gonna take over this PR, make the changes we discussed at the community meeting and push those changes up to your branch. Really want to get this in for `v1.10.0` which is being cut on Tuesday.

Issue #377, opened 2026-06-16
Controller never deletes Temporal-side Worker Deployment version records during normal rollout (only on CRD deletion)


noamyehudai:
### Summary

During a normal rollout, when a Worker Deployment Version drains, the controller deletes that version's Kubernetes Deployment (and the HPA rendered from the WorkerResourceTemplate), but it never deletes the Temporal-side Worker Deployment Version record. The only call to `DeleteWorkerDeploymentVersion` is in the CRD-deletion finalizer (added in #240), which runs only when the whole WorkerDeployment custom resource is deleted, not when an individual version drains during a rollout.

As a result, Temporal-side version records accumulate by one per deploy, and the controller relies entirely on server-side GC to keep the count under `matching.maxVersionsInDeployment`.

### Why this is a problem

In practice the server's at-cap reclamation does not keep up. We see deployments accumulate 100+ drained versions that are never reclaimed (companion issue: temporalio/temporal#10737). Once a deployment reaches the cap, a new build cannot register as a poller and the rollout silently wedges: a merged change ships to a fleet still running the old version.

### Note on existing issues

\#270 ("version retention policy: keep last N after drain") was closed pointing at the Kubernetes `sunset` config (`scaledownDelay` / `deleteDelay`). But `sunset` only deletes the Kubernetes Deployment, not the Temporal-side version record, so it does not bound server-side version growth. The two cleanups are independent.

### Request

An opt-in, controller-side prune of drained versions (for example a `minVersionsToKeep` retention policy that deletes drained, non-current versions beyond the newest N), or at minimum documentation that operators must externally prune drained versions to avoid hitting `maxVersionsInDeployment`. Happy to contribute a PR if there is interest.

### Version

v1.7.0.

aarontsharp:
Also hitting this on v1.7.0. As a stopgap we've added an external CronJob that prunes drained, non-current versions via `delete-version`.

Worth flagging that doing this externally isn't trivial: `delete-version` is gated by the Worker Deployment's `ManagerIdentity`, which the controller owns, so the pruner has to reassign the manager identity to itself, delete, then restore it

gcsabbagh:
> Also hitting this on v1.7.0. As a stopgap we've added an external CronJob that prunes drained, non-current versions via `delete-version`.
> 
[...]

Issue #438, opened 2026-07-12
[Bug] New version can get stuck INACTIVE and never promote when a scale-to-zero autoscaler holds its Deployment at 0 replicas


noamyehudai:
## Summary

With `AllAtOnce` rollout and a **backlog-based scale-to-zero autoscaler** on the versioned worker Deployments (e.g. KEDA's Temporal scaler keyed per `workerDeploymentBuildId`, `minReplicaCount: 0`), a newly rolled-out version can **deadlock and never become Current**: the autoscaler holds its Deployment at 0 replicas, so it has no pollers; the controller then refuses to promote it; the next deploy supersedes it; and it is left permanently as an `INACTIVE` / never-Current version. It never self-recovers.

## Environment

- `temporal-worker-controller` v1.8.0 (Helm chart 0.27.0), controller-runtime v0.24.0
- Temporal Server v1.30.x, CLI v1.6.x
- Rollout strategy: `AllAtOnce`
- Per-version pod autoscaling via a KEDA `ScaledObject` (`temporal` scaler), `minReplicaCount: 0`, trigger keyed to the version's own `workerDeploymentBuildId` backlog

## The deadlock

1. A new build lands → the controller creates the versioned Deployment.
2. The autoscaler scales that Deployment on the backlog **for that build ID**. A version that is not yet Current has no workflows pinned to its build ID → its build-id backlog is `0` → the autoscaler scales it to **0 replicas → 0 pollers**.
3. The controller (`AllAtOnce`) attempts to set it Current and fails:
   ```
   unable to set current deployment version: proposed current version
   'default/example-worker:1.4.0-a1b2' is missing active task queues from the
   current version; these would become unversioned if it is set as the current version
   ```
   A version with 0 pollers polls **no** task queues, so it is "missing the active task queues from the current version." (This guard is correct — promoting a poller-less version would orphan those task queues as unversioned.)
4. **Deadlock**: no pollers → can't be promoted; not Current → no backlog → the autoscaler grants no pollers.
5. The next deploy retargets the controller to a newer build. The stranded version stays **never-Current → `INACTIVE` forever** — nothing ever routes to it, so it never gets pollers and never recovers.

## Trigger condition

[...]

PR #498, opened 2026-08-04
fix(controller): prune Temporal server-side version record on normal drain

jaypipes:
Port of https://github.com/atlanhq/temporal-worker-controller/pull/25.

When a Worker Deployment Version drains during a normal rollout, the controller
deletes its Kubernetes Deployment but never calls DeleteVersion on the Temporal
server. The only existing DeleteVersion call site is the CRD-deletion finalizer,
which runs only when the whole TemporalWorkerDeployment is deleted. Version
records therefore accumulate one per deploy until the per-deployment cap
(matching.maxVersionsInDeployment) is hit. Temporal server is supposed to perform
an eviction/garbage collection of the oldest drained WDV when the per-deployment
limit is reached. However, sometimes Temporal fails to evict these old versions
(see temporalio/temporal#10737), causing the per-deployment version limit to be hit,
and subsequent polls from new worker versions to fail, since the new version
cannot be created.

This PR introduces a new `deleteDrainedVersions` step that runs in the Temporal
phase of `executePlan`, calling `DeleteVersion` for each sunset version with
Identity=getControllerIdentity() so  it satisfies the ManagerIdentity guard,
same as the finalizer. This essentially allows for a manual garbage collection to
side-step temporalio/temporal#10737 while a fix for that is being worked on.

Addresses #377 

Co-authored-by: @tczhao
Co-authored-by: @AshutoshM10
Signed-off-by: Jay Pipes <jay.pipes@temporal.io>


internal/controller/execplan.go
```diff
@@ -579,3 +582,88 @@ func (r *WorkerDeploymentReconciler) executePlan(ctx context.Context, l logr.Log
...
+// the Deployment is gone there is no way to retry on a later reconcile. Left unpruned,
+// these accumulate one per rollout and eventually hit the server's per-deployment
+// version cap, after which every new build ID fails to register (#377).
+//
+// Build IDs are read off the in-memory Deployment objects, which survive their cluster
+// deletion, so this runs in the Temporal phase without reaching back into k8sState.
+// Best-effort: the Kubernetes Deployment is already gone (the primary action), so a
+// failure here only means the Temporal-side record lingers until an operator prunes it
+// or the CRD is deleted. NotRegistered Deployments are also carried in DeleteDeployments;
+// they have no server-side version and return NotFound, which is skipped.
```
GonzaloLuminary, line 604: It would be nice if we had stronger guarantees about this. I don't think it's best practices for operators to manually fix clusters due to these leaks

GonzaloLuminary: I wonder if we can have a way to poll temporal state and remove drained versions?

internal/controller/execplan.go
```diff
@@ -579,3 +582,88 @@ func (r *WorkerDeploymentReconciler) executePlan(ctx context.Context, l logr.Log
...
+	if err != nil {
+		return err
+	}
+
+	deploymentHandler := temporalClient.WorkerDeploymentClient().GetHandle(p.WorkerDeploymentName)
+
+	// Prune Temporal server-side version records for the versions whose Deployments
+	// were just deleted in executeK8sOperations. Must happen in the same reconcile as
+	// the Deployment delete; see deleteDrainedVersions.
+	r.deleteDrainedVersions(ctx, l, deploymentHandler, p)
```
noamyehudai, line 654: The Deployment is already gone by this point, so a failed `DeleteVersion` strands the server-side record permanently, as the doc comment acknowledges. That turns a 100% leak into a silent ~0.3% one.

Data from our external reaper (same API and identity, 3 envs, 30 days): 6,112 attempts, 6,094 succeeded, 18 failed, and **all 18 were transport-level** (12x `RST_STREAM ... CANCEL`, 6x `context deadline exceeded`), zero precondition rejections. So the residual failures are exactly the transient class a retry would absorb, and here they are the one class that can never be retried.

Would inverting the order work? Call `DeleteVersion` before deleting the Deployment:
- delete fails -> skip the Deployment delete, return the error; status entry still exists, next reconcile retries. Converges.
- delete succeeds, Deployment delete fails -> retried next reconcile, `DeleteVersion` returns NotFound, already skipped. Converges.

Preconditions should hold there anyway, since `EligibleForDeletion` asserts drained with no active pods. If the current order is preferred, a Warning event plus a metric would at least make the residue observable rather than log-only.

internal/planner/planner.go
```diff
@@ -644,9 +644,17 @@ func getDeleteDeployments(
...
+			//    executePlan prune the Temporal-side version record in the same
+			//    reconcile as the Deployment delete: EligibleForDeletion is only
+			//    computable while the Deployment (and thus this DeprecatedVersions
+			//    entry) still exists, so this is the only point that can reliably
+			//    prune it. See execplan.deleteDrainedVersions.
 			if version.DrainedSince != nil &&
 				(time.Since(version.DrainedSince.Time) > spec.SunsetStrategy.DeleteDelay.Duration+spec.SunsetStrategy.ScaledownDelay.Duration) &&
-				d.Spec.Replicas != nil && *d.Spec.Replicas == 0 {
+				d.Spec.Replicas != nil && *d.Spec.Replicas == 0 &&
+				version.EligibleForDeletion {
```
noamyehudai, line 657: `eligibleForDeletion` is `Drained && !(Deployment.Status.Replicas > 0)`, so deletion now needs `spec.Replicas == 0` and `status.Replicas == 0` in the same reconcile.

Where that may never coincide: if the rendered per-version autoscaler has `minReplicaCount >= 1` (currently the only way to keep the current version warm, #439, and to dodge the promotion deadlock, #438), it re-raises a drained version's replicas every poll while the controller re-zeroes it. We measured ~120 autoscaler raises against 122 controller re-zeroes in one 30-minute window on two drained versions. Under that oscillation the Deployment, and now the server record too, may never be deleted.

Today only `spec.Replicas == 0` is required and deletes do land on schedule for us. Adding the status condition narrows the window, and narrows it hardest for the users most affected by #377. Worth documenting at minimum; ideally the gate tolerates pods still terminating.

internal/controller/execplan.go
```diff
@@ -579,3 +582,88 @@ func (r *WorkerDeploymentReconciler) executePlan(ctx context.Context, l logr.Log
...
+// these accumulate one per rollout and eventually hit the server's per-deployment
+// version cap, after which every new build ID fails to register (#377).
+//
+// Build IDs are read off the in-memory Deployment objects, which survive their cluster
+// deletion, so this runs in the Temporal phase without reaching back into k8sState.
+// Best-effort: the Kubernetes Deployment is already gone (the primary action), so a
+// failure here only means the Temporal-side record lingers until an operator prunes it
+// or the CRD is deleted. NotRegistered Deployments are also carried in DeleteDeployments;
+// they have no server-side version and return NotFound, which is skipped.
+func (r *WorkerDeploymentReconciler) deleteDrainedVersions(
```
noamyehudai, line 605: Two other paths leak server-side records and won't reach this function, so this reduces #377 rather than closing it:

- Never-current (INACTIVE) versions, superseded before promotion: they never take traffic, never drain, never reach `Drained`. We hold 106 across three envs, all counting toward the cap. Adjacent to #438.
- Records whose CR vanished while the version was still CURRENT: we found five, each pinning a non-drainable current version as a permanent floor.

Not asking for either here, just worth saying explicitly that external cleanup is still needed after this lands.

noamyehudai (commented):
Right fix, right hook point, and using `getControllerIdentity()` matches what actually works: we run an external reaper doing this same delete via the same API and identity, ~6,100 successful deletes over 30 days with zero precondition rejections. Three inline comments, one worth acting on before merge.

internal/controller/execplan.go
jaypipes: @GonzaloLuminary that's a good idea but I think we could add that enhancement in a followup PR. I think this PR that addresses the cleanup of Temporal-side resources by calling DeleteVersion on the drained versions is still worth considering as an iterative improvement to the leaking resource bugs we have here.

internal/controller/execplan.go
jaypipes: @noamyehudai Can you explain this statement further please?

> Records whose CR vanished while the version was still CURRENT: we found five, each pinning a non-drainable current version as a permanent floor.

"pinning ... as a permanent floor" doesn't make sense to me and sounds like something Claude would generate.

internal/controller/execplan.go
jaypipes: > I wonder if we can have a way to poll temporal state and remove drained versions?

@GonzaloLuminary that's precisely what this code is doing. The reconciliation loop fires and asks for Temporal state and removes any drained versions that are returned in Temporal's DescribeWorkerDeploymentVersion's VersionInfoSummary structs.

GonzaloLuminary: Sure, but (at a first glance) if the deployment is gone or if we were not able to set the finalizer (which may happen since it's best effort it seems) then versions will stay in drained state in temporal. For now, we have written a CronJob that deletes drained versions after certain time has passed since the latest update. It'd be better in the controller had a background loop that does said cleanups so that we can sunset the CronJob. In any case, this is just a suggestion for future developments, it should not block merging this PR!

internal/controller/execplan.go
jaypipes: Inverting the order of operations is going to paper over timing issues. Reconciliation should construct its view of observed state and compare that to the latest desired state and actuate any changes to make the observed state match the desired state. If the observed state of the Temporal Worker Deployment returns a bunch of drained/deprecated build IDs and there is no Kubernetes Deployment associated with that deprecated build ID, then the reconciler does not need to actuate any changes since it relies on the Temporal Server to garbage collect the server-side Worker Deployment Versions associated with the deprecated build IDs.

Of course, in temporalio/temporal#10737 there is an indication that that Temporal Server garbage collection may not be working properly, which is why @GonzaloLuminary said:

> I don't think it's best practices for operators to manually fix clusters due to these leaks

which I agree with but adding some sort of "pre-emptive garbage collection" within temporal-worker-controller to deal with the potential bug in temporalio/temporal#10737 is out of scope for *this* particular PR.

internal/controller/execplan.go
```diff
@@ -579,3 +582,88 @@ func (r *WorkerDeploymentReconciler) executePlan(ctx context.Context, l logr.Log
...
+// first). The planner only adds a drained version to DeleteDeployments once it is
+// EligibleForDeletion (see planner.getDeleteDeployments): drained past the sunset
+// delays with no active worker pods. Deleting the Kubernetes Deployment alone would
+// leave the server-side version registered forever — the only other cleanup path is
+// the CRD-deletion finalizer, which never runs during a normal rollout. This is also
+// the only point that can reliably prune it: a version's status entry only exists in
+// status.DeprecatedVersions while its Deployment does (see state_mapper.go), so once
+// the Deployment is gone there is no way to retry on a later reconcile. Left unpruned,
+// these accumulate one per rollout and eventually hit the server's per-deployment
+// version cap, after which every new build ID fails to register (#377).
```
Shivs11: I think there is a small nit here: In the most perfect world, this should only be true if the server is not able to GC it's older, drained versions (and also ofc if there are no drained versions at all in this worker deployment)

Thus, even in the case where for some reason deleting a drained version server-side fails and the next reconcile does not remember it since it's associated k8s deployment is nuked, the best case scenario would nuke this record from the server when someone were to attempt adding another build *after* they have hit the version limit.

internal/planner/planner.go
Shivs11: @jaypipes , I think @noamyehudai does raise something valid here, although I would argue that in this world not removing the temporal side record is not wrong since, well, you do have a poller alive thanks to your per-version HPA.

I think we should just document this.

internal/planner/planner.go
```diff
@@ -726,9 +726,17 @@ func getDeleteDeployments(
...
+			//    worker pods, i.e. Status.Replicas == 0). Requiring this lets
+			//    executePlan prune the Temporal-side version record in the same
+			//    reconcile as the Deployment delete: EligibleForDeletion is only
+			//    computable while the Deployment (and thus this DeprecatedVersions
+			//    entry) still exists, so this is the only point that can reliably
+			//    prune it. See execplan.deleteDrainedVersions.
 			if version.DrainedSince != nil &&
 				(time.Since(version.DrainedSince.Time) > spec.SunsetStrategy.DeleteDelay.Duration+spec.SunsetStrategy.ScaledownDelay.Duration) &&
-				d.Spec.Replicas != nil && *d.Spec.Replicas == 0 {
+				d.Spec.Replicas != nil && *d.Spec.Replicas == 0 &&
```
Shivs11, line 728: nit: I think this entire condition is becoming hard to read and I would be curious to hear if you also felt that. I just realized that for us to assign something to deleteDeployments, we need to have two things from a k8s deployment object perspective. They are:
- it's real replica count should be 0
- it's desired replica count should be 0

looking at the changes, you have added the condition for the desired count, which makes sense to me. But then I realized that the real replica count, which is taken from the live status of the deployment, is calculated and stored as part of EligibleForDeletion, which took me a minute (more actually) to realize

thoughts?

internal/controller/execplan.go
```diff
@@ -579,3 +582,88 @@ func (r *WorkerDeploymentReconciler) executePlan(ctx context.Context, l logr.Log
...
+			},
+		)
+		if err != nil {
+			var notFound *serviceerror.NotFound
+			if errors.As(err, &notFound) {
+				continue
+			}
+			l.Info("could not delete worker deployment version, may require manual cleanup", "buildID", buildID, "error", err)
+		}
+		l.Info("deleted drained worker deployment version", "buildID", buildID)
```
Shivs11: thoughts on adding this event, rather than a log which may require folks to search through them and have a hard time

Shivs11: why are we running this log line if we are not able to actually go ahead and delete a version server side? for example, if we have an error on the server side while nuking a version, we would log the error on line 630 but also log this one, no?

Shivs11 (approved):
just a few nits here and there, but "awesomesauce @jaypipes!" 😉

internal/controller/execplan.go
jaypipes: good catch. I'll address.

internal/controller/execplan.go
jaypipes: @Shivs11 added a note to clarify. Please let me know if that is ok.

internal/controller/execplan.go
jaypipes: fixed.

jaypipes: followup PR, sure, great idea.

internal/controller/execplan.go
carlydf: Hi @noamyehudai, really good point about the inactive versions. I don't think the server even tries to evict those from the list because they are not drained.

I have another follow-up question about the 

> Records whose CR vanished while the version was still CURRENT"

Do you know what version of the controller you were using when those CRs appeared to vanish? The finalizer we added in v1.6.0 (PR #240) is the first time that the worker controller took responsibility for deleting server-side resources on CR deletion. So if you were using an earlier controller version and the CR was deleted, a leftover server-side record would make sense.

internal/controller/execplan.go
carlydf: @noamyehudai, if you are getting a lot of context cancelled / timeouts on your Worker Deployment and Worker Version related API calls (or just elevated latency), you may benefit from scaling up the per-namespace worker for namespaces that make heavy use of worker versioning. Like batch and schedules, these APIs depend on the worker service. See `worker.perNamespaceWorkerCount` and `worker.perNamespaceWorkerOptions` in the dynamic config to scale it.

Depending on metrics are set up for your temporal cluster, you can check to see if there were periods of high backlog on the `temporal-sys-per-ns-tq` task queue, or check the schedule-to-start latency metric emitted by the worker service. 

If worker scale is not the issue, I would be curious to know that! In that case, upgrading your server to v1.31.0, when worker versioning became GA, could reduce the instances of `RST_STREAM ... CANCEL`. We made some reliability improvements to the system workflows between v1.30.2 and v1.31.0 -- I'm not sure if that specific error was a symptom of the issues we fixed, but it's possible.

internal/temporal/worker_deployment.go
```diff
@@ -127,69 +125,94 @@ func GetWorkerDeploymentState(
...
+// versionInfoFromVersionSummary returns a VersionInfo constructed from the
+// supplied Temporal VersionSummary message.
+//
+//nolint:revive // cyclomatic complexity acceptable given breadth of plan execution
+func versionInfoFromVersionSummary(
+	ctx context.Context,
+	client temporalClient.Client,
+	targetBuildID string,
+	strategy temporaliov1alpha1.DefaultVersionUpdateStrategy,
+	depHandle temporalClient.WorkerDeploymentHandle,
```
carlydf, line 150: nit: since the depHandle is not used at all in the above function, imo slightly cleaner to pass the deploymentName string here and then create the depHandle, since we're passing the client anyways. NBD though, feel free to ignore this comment

internal/temporal/worker_deployment.go
```diff
@@ -127,69 +125,94 @@ func GetWorkerDeploymentState(
...
+			// to describe with backoff.
+			//
+			// Note: We can only check whether the task queues that we know of have unversioned pollers.
+			//       If, later on, a poll request arrives tying a new task queue to the target summary, we
+			//       don't know whether that task queue has unversioned pollers.
+			if err = withBackoff(10*time.Second, 1*time.Second, describeVersion); err == nil { //revive:disable-line:max-control-nesting
+				out.AllTaskQueuesHaveUnversionedPoller = allTaskQueuesHaveUnversionedPoller(ctx, client, desc.Info.TaskQueuesInfos)
+			}
+			// NOTE(jaypipes): We swallow any non-nil error here. Should we
+			// at least log the error?
```
carlydf, line 207: I am in support of logging non-nil error here (does not need to block this PR though). Not the initial errors (since those might be expected), but if after 10s we still error, we should log that.

Error message could be something like "could not confirm that all task queues in version %v have unversioned pollers: <err>"

or

"could not confirm that all task queues in version %v have unversioned pollers, will skip progressive rollout steps and set target to current immediately, to avoid routing tasks to unversioned pollers that don't exist: <err>"

internal/planner/planner.go
carlydf: would it address your confusion to say this?

```
// 2. The deployment is scaled to 0 replicas (i.e. Spec.Replicas == 0)
```

alternatively, we could just no longer require condition #2. At the same time, checking #2 before checking #3 is slightly more efficient, because if #2 is false, #3 will certainly be false

internal/controller/execplan.go
noamyehudai: Hmmm interesting observation, I ran a thorough scan on our audit logs.

All the records one had a CR under the old kind (TemporalWorkerDeployment), created
under a ~v1.1/v1.2-era controller.
It was removed on the same day we upgraded to v1.7.0, **but** not by anything controller-shaped: we deleted the deprecated old-kind CRDs as part of that upgrade, and the CRD deletion cascade removed the CR. Since the finalizer-bearing controller only watches the renamed kind, no finalizer was ever attached to that CR, so this isn't a finalizer failure, though it does mean the kind rename itself was an orphan-creation path: any old-kind CR alive at migration time cascade-dies with no server-side teardown.
Probably rare enough not to matter now, but worth remembering if a kind or group rename ever happens again.
Thanks @carlydf

internal/controller/execplan.go
noamyehudai: I'll implement that, do you have any best practices for these settings (and others)? 
maybe worth creating a doc for it

internal/planner/planner.go
carlydf: > We measured ~120 autoscaler raises against 122 controller re-zeroes in one 30-minute window on two drained versions. 

Did you experience this with KEDA ScaledObject or with HPA?

HPA is not expected to scale back up versions that have been explicitly scaled to zero by the controller (which the controller does after version is drained + sunset policy scaledown delay). See https://kubernetes.io/docs/concepts/workloads/autoscaling/horizontal-pod-autoscale/#implicit-maintenance-mode-deactivation for how that works with HPA.

Based on some brief agent conversation, it seems like KEDA does not handle this the same way. So maybe you experienced this with KEDA? I think to solve this issue, the worker controller needs to pause the KEDA ScaledObject right before scaling down a Deployment due to sunset policy, and remove the pause annotation if there is a roll back and that drained version becomes the target version again. See https://keda.sh/docs/2.20/concepts/scaling-deployments/#pausing-autoscaling

carlydf: This issue of KEDA fighting TWC sunset scaledown should be documented and solved as a separate issue!

internal/controller/execplan.go
carlydf: You should only scale it out if you are seeing backlog in the per-ns-worker task queue (indicating that it cannot keep up with the workflow task load being generated by versioning API calls).

We do have guidelines for this that we follow for Temporal Cloud. If you are sticking to the default `frontend.globalNamespaceWorkerDeploymentReadRPS`, and are not heavily using Schedules in the same namespace, the default worker settings should be sufficient.

But it's possible that if persistence is perhaps slower in a self-hosted cluster, workflow task completions / history operations could be slower, and therefore those best practice settings don't apply?

internal/controller/execplan.go
```diff
@@ -579,3 +582,92 @@ func (r *WorkerDeploymentReconciler) executePlan(ctx context.Context, l logr.Log
 
 	return errors.Join(append(applyErrs, statusErrs...)...)
 }
+
+// deleteDrainedVersions prunes the Temporal server-side Worker Deployment Version
```
carlydf, line 586: nit: it feels like this whole comment could be less verbose. It reads like claude output. I don't think every future consumer of this code needs to have all this detail .

internal/controller/execplan.go
carlydf: You should only scale it out (via increasing pollers-per-worker and/or number of workers) if you are seeing backlog in the per-ns-worker task queue (indicating that it cannot keep up with the workflow task load being generated by versioning API calls).

We do have guidelines for this that we follow for Temporal Cloud. If you are sticking to the default frontend.globalNamespaceWorkerDeploymentReadRPS, and are not heavily using Schedules in the same namespace, the default worker settings should be sufficient.

But it's possible that if persistence is perhaps slower in a self-hosted cluster, workflow task completions / history operations could be slower, and therefore those best practice settings don't apply?

carlydf:
Can the PR description also be updated to more clearly call out the specific issue? It feels distracting for this PR description to call out `KEDAScaledObjectErrors` and get into the details of the controller deletion mechanism when the real issue is this Temporal eviction bug: "sometimes Temporal fails to evict old versions as it is supposed to (https://github.com/temporalio/temporal/issues/10737), causing the per-deployment version limit to be hit, and subsequent polls from new worker versions to fail, since the new versions cannot be created."

jaypipes:
> Can the PR description also be updated to more clearly call out the specific issue? It feels distracting for this PR description to call out `KEDAScaledObjectErrors` and get into the details of the controller deletion mechanism when the real issue is this Temporal eviction bug: "sometimes Temporal fails to evict old versions as it is supposed to ([temporalio/temporal#10737](https://github.com/temporalio/temporal/issues/10737)), causing the per-deployment version limit to be hit, and subsequent polls from new worker versions to fail, since the new versions cannot be created."

@carlydf updated the PR. lemme know if that's good with you. If it is, I will address all the other comments in a followup PR if that's OK with you. I'd like to get this one into tomorrow's v1.9.0 release.

carlydf (approved):
I like the description, thank you!

jaypipes:
@carlydf @Shivs11 I will address all the review comments in a followup PR. sorry for resolving those comment threads -- the settings on this repo force me to resolve all comment threads otherwise I can't merge the PR and I wanted to get this in for tomorrow's release...

PR #508, opened 2026-08-07
fix(controller): delete k8s worker Deployments during WD cleanup

### LINKED ISSUE #463: WorkerDeployment deletion finalizer deadlocks with active pollers
## What happened

Deleting an active `WorkerDeployment` leaves the resource in `Terminating` indefinitely.

The deletion finalizer changes the Temporal routing configuration and then calls `DeleteVersion` with `SkipDrainage: true`. Temporal still rejects that request while the generated Kubernetes Deployment has active pollers:

```
version <deployment>:<build-id> cannot be deleted since it has active pollers
```

The generated Deployment keeps its replicas because its owner `WorkerDeployment` is held by the finalizer, so Kubernetes garbage collection cannot terminate the pods. Every reconciliation retries the same server-side deletion against the same active pollers.

## Versions

- temporal-worker-controller: v1.8.0
- Helm chart: 0.27.0
- Temporal Server: 1.31.2
- Kubernetes: kind

I also checked current `main` at `7e13417f0f7009f637b0dd09e7c8ae3652729ac5`; `handleDeletion` still has the same sequence.

## Reproduction

1. Create a `WorkerDeployment` with one replica.
2. Wait for `Ready=True` and confirm that its Worker Deployment Version has active pollers.
3. Delete the `WorkerDeployment` directly.
4. Observe that the finalizer sets the current Temporal version to unversioned.
5. Observe repeated `DeleteVersion` failures because the generated Deployment and its pollers remain active.
6. Scale the generated Deployment to zero and wait until Temporal no longer reports active pollers. The next reconciliation then completes cleanup.

## Expected behavior

The supported deletion flow should stop the controller-managed pollers before attempting `DeleteVersion`, wait observably until the pollers are gone, and only then remove the Temporal Worker Deployment Version, generated Kubernetes resources, and finalizer.

A possible controller-owned sequence would be:

1. Scale all generated Deployments for the deleting `WorkerDeployment` to zero.
2. Requeue until the Kubernetes pods are gone and Temporal reports no active pollers.
3. Clear routing, delete versions and the server-side deployment.
4. Remove the finalizer so Kubernetes garbage collection can finish.

## Relevant code

`internal/controller/worker_controller.go`, `handleDeletion`.

The current comment says pollers will disappear once pods terminate, but the pods are owner-referenced by the resource whose deletion the same finalizer is blocking.

tomba7:
## Why the deadlock happened

Two things could remove a version's worker pods:

1. executePlan, via the sunset path, but only for deprecated versions, which
   are already scaled to zero by then and so aren't polling anyway. 
	 
2. ownerReference GC which reaps every child Deployment, active one
   included, once the WD object is gone. This is what normally cleans up an active
   version.
   
Note: deprecated versions are already scaled to zero by the sunset path,
so the ones that matter here are the versions whose k8s deployments still 
have running pods.

Manually deleting the WD (repro step #3) blocks both 1 and 2. Reconcile sees the
deletionTimestamp, enters handleDeletion and returns, never reaching
executePlan. And GC won't touch the children while the finalizer keeps the WD alive.

So the pods keep polling, and the server rejects DeleteVersion since the version
has active pollers, even with SkipDrainage set. Redirecting current to
unversioned doesn't help either since it just reroutes new tasks but doesn't stop
the currently running pods.

So handleDeletion keeps erroring, finalizer never gets removed, pods keep polling,
WDVs on the Temporal server stay non deletable forever. This is the deadlock.

## How the fix works

We break the cycle at the only point that isn't blocked, by deleting child Deployments
directly, i.e. by handling it inside handleDeletion itself, before asking the server
to delete the WDV.

The cleanup sequence gains a new step 3:

1. Clear the ramping version
2. Set the current version to unversioned
3. Delete the k8s deployments so their pods stop polling   <----- here
4. Delete all registered versions (SkipDrainage: true)
5. Delete the worker deployment record on the Temporal server

The new step lists the WD's children with GetDeploymentState, the same lookup normal
reconciliation uses, so it only ever sees this WD's own worker Deployments.

It deletes each one, ignoring NotFound so that retries are idempotent. Any other error
requeues after 10s with the finalizer intact. Iteration is over the DeploymentsByTime
slice rather than the Deployments map so that the log lines stay stable across retries

Deleting the Deployments terminates the pods and stops the polling. The server caches
pollers for a few minutes after they vanish, so steps 4-5 usually still fail on the
first pass, which is expected. And each failure just requeues.

Once the cached pollers age out, DeleteVersion succeeds, the finalizer comes off, and K8s
removes the WD. Everything here is inside the deletion path, so normal rollout and sunset
behavior is untouched.

## Testing

Added coverage in `deletion_integration_test.go` which asserts that both versioned
k8s deployments are deleted once the finalizer completes. Note that this test overrides
matching.PollerHistoryTTL to 1s, so the test verifies the teardown rather than reproducing
the multi minute poller expiry.

Verified manually on a local cluster running Temporal server v1.31.2 with the controller
built from this branch

- Reproduced the deadlock by deleting a WD with an active version and workers running. The
  WD stayed in Terminating, the worker pods stayed Running, and the controller logged
  `cannot be deleted since it has active pollers` on every 10s requeue.
- Reran the same steps with the fix and the child Deployments were deleted immediately,
  the pods terminated, steps 4-5 requeued for ~5 minutes until the cached pollers aged
  out, then the finalizer was removed and the WD was deleted.
- Verified server-side afterwards and no versions remain and the worker deployment record
  is gone.
- Verified nothing is orphaned, i.e. no k8s deployments or worker pods left in the namespace.
- `go test -v -tags test_dep ./internal/tests/internal -run 'TestIntegration/deletion-sets-current-to-unversioned' -timeout 30m`
- `go test -v -tags test_dep ./internal/tests/internal -run TestIntegration -timeout 30m`
- `go vet ./internal/controller ./internal/tests/internal`

Fixes #463.


jaypipes:
@tomba7 apologies for the delayed review.

From your PR summary, you wrote this:

> Note: The active version is the one still running pods, so it's the one we need gone, and its Deployment has to go before its WDV can be deleted, not after.

Be aware that a Worker Deployment may have many "active versions" at any given time. If by "The active version" you are actually referring to the *current version*, that is a different thing. The current version is the specific build ID in the Worker Deployment that new Workflow Executions will be routed to when those Workflow Executions specify the Worker Deployment name *and* use  `AutoUpgrade` versioning behaviour. If Workflow Executions specify a particular Build ID and use `Pinned` versioning behaviour, the matching service in Temporal Server will route those Workflow Executions to a Worker Deployment Version with that matching Build ID -- and thus route the Workflow Execution to one of the Pods in the Kubernetes Deployment that is associated with that Build ID. This means that there can be many "active versions" in a Worker Deployment at any given time.

Just want to make sure you understood that nuance...

tomba7:
@jaypipes My apologies for the bad terminology. What I meant to say was "active version" -> "current WorkerDeployment in K8s", and not anything to do with versioning.

jaypipes:
> @jaypipes My apologies for the bad terminology. What I meant to say was "active version" -> "current WorkerDeployment in K8s", and not anything to do with versioning.

No reason whatsoever to apologize @tomba7! This terminology is maddeningly overlapping and confusing.

internal/controller/worker_controller.go
```diff
@@ -643,11 +650,38 @@ func (r *WorkerDeploymentReconciler) handleDeletion(
 		l.Info("No current version set, skipping unversioned redirect")
 	}
 
-	// Step 3: Delete versions that are eligible. Versions that are still draining
+	// Step 3: Tear down the worker Deployments backing these versions so their pods
```
```suggestion
	// Step 3: Tear down the k8s Deployments backing these versions so their pods
```

internal/controller/worker_controller.go
```diff
@@ -643,11 +650,38 @@ func (r *WorkerDeploymentReconciler) handleDeletion(
...
+	// the finalizer is never removed (deadlock). Deleting them explicitly breaks it,
+	// rather than waiting for ownerRef GC, which is itself blocked on the finalizer.
+	k8sState, err := k8s.GetDeploymentState(
+		ctx,
+		r.Client,
+		workerDeploy.Namespace,
+		workerDeploy.Name,
+		workerDeploymentName)
+	if err != nil {
+		return fmt.Errorf("unable to list child deployments during cleanup: %w", err)
```
```suggestion
		return fmt.Errorf("unable to list k8s deployments during deletion of worker deployment: %w", err)
```

internal/controller/worker_controller.go
```diff
@@ -643,11 +650,38 @@ func (r *WorkerDeploymentReconciler) handleDeletion(
...
+	}
+	// Deletion order is irrelevant, but iterating over the DeploymentsByTime slice instead
+	// of the Deployments map gives deterministic ordering, so the log lines below stay
+	// stable across the 10s retries.
+	for _, d := range k8sState.DeploymentsByTime {
+		l.Info("Deleting worker deployment during cleanup", "deployment", d.Name)
+		if err := r.Delete(ctx, d); err != nil && !apierrors.IsNotFound(err) {
+			return fmt.Errorf("unable to delete worker deployment %s during cleanup (will retry): %w", d.Name, err)
```
```suggestion
			return fmt.Errorf("unable to delete k8s deployment %s during deletion of worker deployment (will retry): %w", d.Name, err)
```

internal/controller/worker_controller.go
```diff
@@ -643,11 +650,38 @@ func (r *WorkerDeploymentReconciler) handleDeletion(
...
+		}
+	}
+
+	// Step 4: Delete versions that are eligible. Versions that are still draining
 	// are force-deleted with SkipDrainage since the WD is being removed entirely.
```
jaypipes: as discussed on Slack yesterday with @carlydf, `SkipDrainage` actually does not do "force-deletion". There is actually no "force-deletion" API option. `SkipDrainage` causes the call to `DeleteVersion` to succeed if the WDV is in **draining** status. `DeleteVersion` will fail (even with `SkipDrainage`) if the WDV has active pollers within the last 4 minutes...

internal/controller/worker_controller.go
```diff
@@ -643,11 +650,38 @@ func (r *WorkerDeploymentReconciler) handleDeletion(
...
-	// If any version fails to delete (e.g. active pollers), return an error so the
-	// reconciler requeues. Pollers disappear once pods terminate and the next
-	// reconciliation will succeed.
+	// If any version fails to delete (e.g. active pollers still draining after the
+	// Deployment delete above), return an error so the reconciler requeues. Pollers
```
```suggestion
	// k8s Deployment delete above), return an error so the reconciler requeues. Pollers
```

internal/controller/worker_controller.go
```diff
@@ -643,11 +650,38 @@ func (r *WorkerDeploymentReconciler) handleDeletion(
...
+	// disappear once pods terminate and a subsequent reconciliation will succeed.
```
```suggestion
	// disappear once pods terminate (and a minutes-long delay on the temporal server side)
	// after which a subsequent reconciliation will succeed.
```

internal/controller/worker_controller.go
```diff
@@ -660,7 +694,7 @@ func (r *WorkerDeploymentReconciler) handleDeletion(
 		}
 	}
 
-	// Step 4: Delete the deployment itself. This only succeeds if all versions are gone.
+	// Step 5: Delete the deployment itself. This only succeeds if all versions are gone.
```
```suggestion
	// Step 5: Delete the worker deployment itself. This only succeeds if all versions and
	// their associated k8s deployments are gone.
```

jaypipes (commented):
@tomba7 can we add a unit test and/or integration test that exercises the code paths introduced in this PR please?

internal/controller/worker_controller.go
tomba7: Nice catch. Fixed the language to reflect that fact.

jaypipes (approved):
👍 thank you @tomba7!

internal/tests/internal/deletion_integration_test.go
```diff
@@ -186,6 +187,22 @@ func testDeletionSetsCurrentToUnversioned(
 	})
 	t.Log("WD deleted successfully (finalizer completed)")
 
+	// Verify the WD's k8s deployments were deleted.
```
carlydf, line 190: just confirming: does this part of the test fail without the code changes in `worker_controller.go`?

by the time we get past `WD deleted successfully` on line 188, were the k8s Deployments also deleted? I'd love it if this test could capture the indefinite deadlock that you saw in prod, so it's clear what needs to be fixed. Maybe we need to start some workflows and have the versions be not-drained to trigger the issue?

internal/controller/worker_controller.go
```diff
@@ -643,11 +650,38 @@ func (r *WorkerDeploymentReconciler) handleDeletion(
...
-	// reconciler requeues. Pollers disappear once pods terminate and the next
-	// reconciliation will succeed.
+	// Step 3: Tear down the k8s deployments backing these versions so their pods
+	// terminate and stop polling. This is required before a version can be deleted:
+	// the server rejects DeleteVersion (even with SkipDrainage=true) while a version
+	// still has active pollers. During normal reconciliation this teardown happens in
+	// executePlan, but that path is downstream of the deletion bail-out and never runs
+	// here. Without this step the pods keep polling, DeleteVersion keeps failing, and
+	// the finalizer is never removed (deadlock). Deleting them explicitly breaks it,
+	// rather than waiting for ownerRef GC, which is itself blocked on the finalizer.
```
carlydf: I'm going to be nit-picky about the verbosity of this comment, just because I feel like the more concise we can make it, the better this codebase will serve future consumers of it.

Right now, this comment says basically exactly what the comment on line 537-543 says, but in slightly more words. Could you revisit both of them and just make sure that they are saying exactly what is needed and no more?

internal/controller/worker_controller.go
```diff
@@ -643,11 +650,38 @@ func (r *WorkerDeploymentReconciler) handleDeletion(
...
+		r.Client,
+		workerDeploy.Namespace,
+		workerDeploy.Name,
+		workerDeploymentName)
+	if err != nil {
+		return fmt.Errorf("unable to list k8s deployments during deletion of worker deployment: %w", err)
+	}
+	// Deletion order is irrelevant, but iterating over the DeploymentsByTime slice instead
+	// of the Deployments map gives deterministic ordering, so the log lines below stay
+	// stable across the 10s retries.
```
carlydf: nit: same here about comment verbosity. why do future readers care about log stability across "the 10s retries" ? what does "the 10s retries" refer to?

internal/controller/worker_controller.go
```diff
@@ -643,11 +650,38 @@ func (r *WorkerDeploymentReconciler) handleDeletion(
...
+		if err := r.Delete(ctx, d); err != nil && !apierrors.IsNotFound(err) {
+			return fmt.Errorf("unable to delete k8s deployment %s during deletion of worker deployment (will retry): %w", d.Name, err)
+		}
+	}
+
+	// Step 4: Delete every registered version. SkipDrainage lets DeleteVersion succeed on
+	// versions that are still draining, which is acceptable here since the whole WD is
+	// going away. If any version fails to delete (e.g. it still has recent pollers after the
+	// k8s Deployment delete above), return an error so the reconciler requeues. Pollers age
+	// out on the server minutes after the pods terminate, so a later attempt succeeds.
```
carlydf: @tomba7 do you think it would be helpful for readers (esp those managing their own self-hosted clusters) to know that the dynamic config knob controlling this "age out" is `matching.PollerHistoryTTL`, which defaults to 5 minutes?

Just thinking it may be better to be specific in the places that mention this delay instead of just saying "minutes"

carlydf (commented):
Thanks for this PR @tomba7 ! I really appreciate the PR and have no technical concerns with the code change.

My one actual code-related comment is about the test. It would be great to adjust the test so that it fails without your fix (if it doesn't already).

In addition to my request for less verbose comments, could you also please make the PR description less verbose? Maybe refer to the issue that it is closing at the top, and let that issue carry most of the details about the bug, while this PR description can just be specific about what behavior was changed in the controller.

internal/tests/internal/deletion_integration_test.go
tomba7: > just confirming: does this part of the test fail without the code changes in worker_controller.go?

Yes, this part fails without the deletion changes in `worker_controller.go`. 

> by the time we get past WD deleted successfully on line 188, were the k8s Deployments also deleted?

Since the test has no controller manager, nothing else would remove the child k8s deployments once the WD is gone. And they are only gone here because cleanup fix now deletes them explicitly, which is what this assertion verifies.

> I'd love it if this test could capture the indefinite deadlock that you saw in prod, so it's clear what needs to be fixed.

The test also cannot reproduce the deadlock seen in prod. My understanding is that the 1 sec TTL in the harness (matching.PollerHistoryTTL) means active pollers stop being visible to the server within a second, so the DeleteVersion call succeeds. And since the test env has no kubelet, and the workers are in-process, deleting the k8s deployment doesn't stop the polling.

This test verifies the post-condition only. Reproducing the real hang would require a realistic TTL plus worker shutdown tied to deployment deletion. Happy to do that as part of a follow-up PR. 

> Maybe we need to start some workflows and have the versions be not-drained to trigger the issue?

Starting workflows would exercise the drainage path rather than the poller path, and we pass SkipDrainage: true, so I don't think it reaches this bug. I'm not an expert on the server side here, so please correct me if I've got this wrong.

carlydf: Ah yes, makes sense. I forgot about the gymnastics we are doing to emulate worker pods spinning up with only the env-test environment. I am good with just this test then.

carlydf (approved):

Issue #330, opened 2026-05-22
[Feature Request] Multi-pod worker deployments


dssysolyatin:
### Is your feature request related to a problem? Please describe.

Consider a workflow DSL that lets users define their own workflows inside the product. The workflow implementation that interprets this DSL uses two kinds of activities:

  1. Activities that perform heavy computation
  2. Activities that only make HTTP or DB calls

The different activity workers and the workflow worker should not live in the same pod, for a few reasons:
  - Each activity worker type and the workflow worker has its own security policy /
  ServiceAccount.
  - If an activity container fails, it should not affect the workflow worker. When
  containers share a pod, both contribute to whether the pod is considered healthy.
  - They have different scaling factors
  - ......

That said, a single release that covers all workers is usually desirable, except in specific cases.

### Describe the solution you'd like

Let users set the Temporal deployment name explicitly instead of always auto-generating it. With an explicit name, users could run multiple WorkerDeployment resources that share the same temporal deployment.

There might be a better way or practice how to handle such cases

yhwang-nydig:
We have a document-processing workload that runs into this limitation, and we’d be happy to contribute a fix. The shared `deploymentName` idea in the issue is a useful starting point; after tracing the current reconciliation paths, we think keeping one CR as the owner and adding worker pools would provide safer ownership and release semantics.

## The use case

A document-processing application commonly runs several worker roles as separate Kubernetes Deployments—for example, workflow/I/O workers, compute-intensive parsing workers, and extraction workers. Each role may need a different image, ServiceAccount, resource profile, node placement, failure domain, and scaling policy.

Those roles are still built and promoted as one release. They therefore need to share one Temporal Worker Deployment Version so that PINNED workflows dispatch activities to workers from the matching release.

Temporal already supports this server-side: [a Worker Deployment Version can contain multiple Workers polling multiple Task Queues while running the same build](https://docs.temporal.io/production-deployment/worker-deployments/worker-versioning). The missing piece is expressing that topology in the controller, where one CR currently produces one Kubernetes Deployment per version.

Running one `WorkerDeployment` CR per role does not preserve the release boundary. Activities on queues belonging to another Worker Deployment are independent activities and run on that queue’s Current or Ramping Version ([temporalio/temporal#6957](https://github.com/temporalio/temporal/pull/6957)). That allows roles to move to different versions during rollout and prevents an atomic release or rollback.

## Why we prefer pools over a shared `deploymentName`

[...]

PR #513, opened 2026-08-10
test: golden tests pinning worker identity derivation (#330)

yhwang-nydig:
This adds a focused regression test for the identity values derived from an existing `WorkerDeployment`:

- build ID
- Temporal Worker Deployment name
- Kubernetes Deployment name
- selector labels

It also covers the empty-image and custom-build-ID paths.

The multi-pod work discussed in #330 would touch this area. This gives us a concrete compatibility baseline so existing single-template deployments keep the same identities through that refactor.

Tested with `go test ./internal/k8s`.


internal/k8s/identity_test.go
```diff
@@ -0,0 +1,71 @@
...
+		},
+		k8s.ComputeSelectorLabels(worker.Name, buildID),
+	)
+}
+
+func TestComputeBuildIDCompatibility(t *testing.T) {
+	t.Run("empty image", func(t *testing.T) {
+		worker := legacyWorkerDeployment()
+		worker.Spec.Template.Spec.Containers[0].Image = ""
+		assert.Equal(t, "4444447645", k8s.ComputeBuildID(worker))
```
jaypipes, line 63: Say what now? :)

yhwang-nydig:
ah this was supposed to cover more regression testing, prematurely opened. going to close and reopen when ready

Issue #477, opened 2026-07-28
Connection and Credentials Handling Improvements


jaypipes:
This is a parent issue around a number of related sub-issues involving the Connection CR and Temporal credentials configuration.

PR #516, opened 2026-08-11
BUG[493] sub-task 1 - complete connection env/volume updates

niyomukiza-mechack:
## What was changed

Fixed `updateDeploymentWithConnection` to apply the complete set of connection env vars, volumes, and mounts when a Connection spec changes.
Before, when you changed the authentication from mTLS to API-key, this function only rewrote the temporal address and tls server name and left stale env vars and never added the new credentials 

updateDeploymentWithConnection now handles changing from mTLS to api-key and vice versa properly by applying the credentials and env vars and removing stale ones.
I added helper functions: setEnvVarFrom, ensureTLSVolume, removeTLSVolume, ensureTLSVolumeMount, removeTLSVolumeMount

## Why?

A customer migrated a namespace from mTLS to API-key auth by creating a new Connection and switching connectionRef on their WorkerDeployments. The controller re-pointed all versions at the new connection but only applied temporal address and never sets api key or remove the stale mTLS env vars and volumes. The draining pods crash looped.
Context: https://github.com/temporalio/temporal-worker-controller/issues/477?issue=temporalio%7Ctemporal-worker-controller%7C493

## Checklist

1. Closes 

2. How was this tested:

Unit tests
Locally (Kind cluster + Temporal Cloud)
I deployed this branch intro a kind cluster
Also deployed a test worker connected to a namespace(ns-nm-twc-01.temporal-dev) over mTLS and verified the expected state(env vars, volume, mount all present and no temporal api-key)
Switched the Connection CR from mTLS to API key and triggered reconciliation
And the controller correctly read the connection change and attempted the new auth mode

Full end to end mTLS ->API key live switch was not completable. I ran into infrastructure/environment limitations. 

I got this error`certificate signed by unknown authority` when the controller tried to connect to the regional staging environment.(`us-west-2.aws.api.tmprl-test.cloud:7233`)
And this error `tls: certificate required` when the controller tried connecting to the namespace. The namespace endpoint enforces mTLS at the transport layer
I tried setting both mutualTLSSecretRef and apiKeySecretRef on the Connection CR to solve the above but The Connection CRD's validating webhook rejected setting both mutualTLSSecretRef and apiKeySecretRef simultaneously.

3. Any docs updates needed?


internal/planner/planner.go
```diff
@@ -406,32 +406,53 @@ func checkAndUpdateDeploymentConnectionSpec(
 	return nil
 }
 
-// updateDeploymentWithConnection updates an existing deployment with new ConnectionSpec
+// updateDeploymentWithConnection updates an existing deployment in-place to match a new ConnectionSpec.
+// It rewrites the controller-managed connection env vars and the mTLS volume/mount, adding and removing them as needed,
+// So switching auth mode (mTLS <-> API key or to/from no-credentials) yields as fully-configured pod.
```
```suggestion
// So switching auth mode (mTLS <-> API key or to/from no-credentials) yields a fully-configured pod.
```

jaypipes (approved):
Good stuff, thank you @niyomukiza-mechack :)

PR #517, opened 2026-08-12
standardize all temporal Go package deps

jaypipes:
Downgrades all the Go package dependencies for Temporal (api, server, sdk) to the latest stable server release version (`v1.31.2`) and all of the api-go and sdk-go direct dependencies that server's go.mod includes:

```
	go.temporal.io/api v1.62.8
	go.temporal.io/auto-scaled-workers v0.0.0-20260407181057-edd947d743d2
	go.temporal.io/sdk v1.41.1
```

We don't use any server or API feature introduced in later Temporal Go packages and the dependabot version update PRs were getting borked over and over so I just disabled tracking of temporal go packages in the dependabot configuration.


Shivs11 (approved):

jlegrone (commented):
Late to the party, but thank you!

Issue #493, opened 2026-08-02
Changing `connectionRef` re-points every retained version, which can leave older versions unable to authenticate


nishijun:
## What happened

We migrated a namespace from mTLS to API-key authentication by adding a second `Connection` and
switching each `WorkerDeployment`'s `spec.workerOptions.connectionRef` to it, rather than editing
the existing `Connection` in place. Our reasoning was that adding a new `Connection` would leave
versions that were still draining on the old one, so no version would ever see a half-updated
configuration.

That reasoning was wrong. On apply, the controller re-pointed **every version it manages**,
including retained ones, and each of their Deployments rolled to a new ReplicaSet carrying the new
`Connection`'s `hostPort` as `TEMPORAL_ADDRESS`.

Retained versions' pod templates predated the migration, so they carried mTLS credentials and no
API-key configuration. They ended up dialling the regional endpoint with a client certificate,
which Temporal Cloud rejects:

```
temporal server not ready after 2m0s: failed reaching server:
rpc error: code = Unauthenticated desc = Jwt is missing
```

36 pods crash-looped on a 2-minute cycle. 21 sleeping workflows pinned to those versions stalled
until we moved them. This was staging and nothing was lost, but the same change in production
would be considerably worse — our production `sunset.scaledownDelay` is 2h rather than 15m, so the
crash loop would persist for two hours rather than fifteen minutes.

Version: controller v1.8.0, 20 `WorkerDeployment`s, all workflows `VersioningBehavior: Pinned`.

## Why we found it surprising

`connectionRef` lives on the `WorkerDeployment`, and [concepts.md][concepts] documents a
`Connection` as shareable across worker deployments. What is not stated anywhere we could find is
that changing it **mutates every existing version**, including versions the controller is
deliberately keeping alive to drain.

Everywhere else, the model is that a version is an immutable unit: change the pod template, get a
new version, old ones keep running as they were. `connectionRef` breaks that expectation — it is
the one field whose change reaches backwards into versions that already exist.

Read as an operator, "switch `connectionRef`" looks like an atomic, forward-only operation. It is
not one while versions are retained.

## Suggestions

**1. Document it.** A note in [concepts.md][concepts] under Worker Options, and in the sunset
section, would have been enough for us. Something to the effect of:

> Changing `connectionRef` applies to all versions the controller manages, including versions that
> are draining. Versions whose pod template is not compatible with the new `Connection` will fail
[...]

PR #525, opened 2026-08-12
Fix: pin connection on deprecated versions

niyomukiza-mechack:
## What was changed

In `internal/planner/planner.go` there's a `getUpdateDeployments` function that gets called during the planning process and looks at the observed state and returns a list of of existing Deployments that need their pod/connection updated.
Currently there's a loop that propagates connectionRef changes to draining and drained versions.
I deleted that loop so that only target and current versions now follow connectionRef changes. Deprecated versions(draining and drained) keep the connection baked into their Deployment at creation time.

During end to end testing I ran into a Kubernetes validation error(`volumeMounts[0].name: Not found: "/etc/temporal/tls")` from my previous pr(https://github.com/temporalio/temporal-worker-controller/pull/516#issuecomment-5258412619), `ensureTLSVolumeMount` was setting `Name` instead of `MountPath` so I addressed that here.
## Why?

Context: https://github.com/temporalio/temporal-worker-controller/issues/493
When switching `ConnectionRef` switching connectionRef on their WorkerDeployments. The controller re-pointed all versions(current,draining and drained) at the new connection. Draining versions had working credentials and were serving open pinned workflows so they should have been left untouched.

## Checklist

1. Closes 

2. How was this tested:

Unit tests
Local end 2 end (Kind cluster + Temporal Cloud)
I reproduced the bug and verified the fix with a 3 version rainbow deployment (v1 DRAINING, v2 DRAINING, v3 CURRENT) connected to ns-nm-twc-01.temporal-dev namespace over mTLS.

Before the fix: I changed mutualTLSSecretRef from temporal-cloud-mtls to temporal-cloud-mtls-2. All three versions updated:
Screenshot 2026-08-12 at 12 59 57 PM

After the fix: I made the same connection change. Only v3 (current version) updated; v1 and v2 (draining) stayed pinned
image

Auth switch of mTLS to API key was not testable in local staging due to environment constraint. I got a `certificate signed by unknown authority` and similar errors I mentioned in https://github.com/temporalio/temporal-worker-controller/issues/493

3. Any docs updates needed?

### DIFF
--- a/internal/planner/planner.go
+++ b/internal/planner/planner.go
@@ -519,7 +519,7 @@ func removeTLSVolume(volumes []corev1.Volume) []corev1.Volume {
 func ensureTLSVolumeMount(mounts []corev1.VolumeMount) []corev1.VolumeMount {
 	for i := range mounts {
 		if mounts[i].Name == "temporal-tls" {
-			mounts[i].Name = "/etc/temporal/tls"
+			mounts[i].MountPath = "/etc/temporal/tls"
 			return mounts
 		}
 	}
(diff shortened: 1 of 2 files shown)


jaypipes (approved):
👍 just fixes a typo from a previous PR. all good.

PR #527, opened 2026-08-14
Fix: Skip replica management when HPA is attached

### LINKED ISSUE #350: [Bug] HorizontalPodAutoscaler continuously rescales due to controller interference
# Bug: HorizontalPodAutoscaler continuously rescales due to controller interference

## Summary
When using `WorkerResourceTemplate` to configure HPA for a `TemporalWorkerDeployment`, the controller continuously resets the underlying Deployment's replica count to 1, directly conflicting with HPA scaling decisions. This results in a cascading rescale loop where the HPA scales up to `minReplicas` every ~10-15 seconds, only to be reset by the controller milliseconds later.

## Severity
**HIGH** - HPA is completely non-functional for TemporalWorkerDeployment resources.

---

## Environment
- **temporal-worker-controller chart version**: `0.24.1`
- **temporal-worker-controller container image version**: `v1.5.2`
- **Kubernetes version**: 1.33
- **Container platform**: EKS with Karpenter
- **Reproducer**: Minimal example provided below

---

### Observed Behavior
1. **Deployment is locked at 1 replica** despite HPA configuration specifying `minReplicas: 3`
2. **HPA generates continuous rescale events** — 11,087+ "SuccessfulRescale" events in 24 hours with message: "Current number of replicas below Spec.MinReplicas"
3. **Controller logs show active interference**:
   - Repeated `"msg":"scaling deployment"` entries every 10-15 seconds
   - Each entry explicitly sets `"replicas":1`
   - Followed by deployment error on ObjectReference handling
4. **HPA status confirms the conflict**:
   ```yaml
   status:
     currentReplicas: 1       # What's running
     desiredReplicas: 3       # What HPA wants
   ```

### Expected Behavior
When a `WorkerResourceTemplate` with HPA is created for a `TemporalWorkerDeployment`, the controller should:
1. Recognize that HPA is managing the Deployment
2. **NOT modify** the `spec.replicas` field (allow HPA full control)
3. Allow HPA to scale between `minReplicas` and `maxReplicas` based on metrics
4. Maintain steady state at the desired replica count

---

### Minimal TemporalWorkerDeployment
```yaml
metadata:
  name: temporal-worker-example
  namespace: default
spec:
  # NOTE: No replicas field — HPA should manage scaling
  rollout:
    strategy: AllAtOnce
  sunset:
    deleteDelay: 2h
    scaledownDelay: 15m
  template:
    metadata:
      labels:
        app: temporal-worker-example
    spec:
      containers:
        - name: temporal-worker-example
          image: myrepo/myapp:v1.0.0
          resources:
            limits:
              cpu: 1000m
              memory: 1Gi
            requests:
              cpu: 500m
              memory: 512Mi
      serviceAccountName: default
  workerOptions:
    connectionRef:
      name: temporal-connection
    temporalNamespace: default.local
```

### WorkerResourceTemplate with HPA (Desired Scaling Config)
```yaml
kind: WorkerResourceTemplate
metadata:
  name: temporal-worker-example-hpa
  namespace: default
spec:
  temporalWorkerDeploymentRef:
    name: temporal-worker-example
  template:
    apiVersion: autoscaling/v2
    kind: HorizontalPodAutoscaler
    spec:
      scaleTargetRef: {}  # Controller should populate this
      minReplicas: 3
      maxReplicas: 10
      behavior:
        scaleUp:
          stabilizationWindowSeconds: 30
          policies:
            - type: Pods
              value: 2
              periodSeconds: 60
            - type: Percent
              value: 50
              periodSeconds: 60
          selectPolicy: Max
      metrics:
        - type: Resource
          resource:
            name: cpu
            target:
              type: Utilization
              averageUtilization: 70
        - type: Resource
          resource:
            name: memory
            target:
              type: Utilization
              averageUtilization: 75
```

### Steps to Reproduce
1. Apply the above TemporalWorkerDeployment (without `spec.replicas`)
2. Apply the WorkerResourceTemplate with HPA configuration
3. Observe the underlying Deployment created by the controller
4. **Expected**: Deployment scales to 3+ replicas based on CPU/memory
5. **Actual**: Deployment stays at 1 replica, HPA generates continuous SuccessfulRescale events

### Validation
Check the generated HPA status:
```bash
kubectl get hpa -n default -l temporal.io/deployment-name=temporal-worker-example -o yaml
```

You'll see:
```yaml
status:
  currentReplicas: 1
  desiredReplicas: 3  # or higher based on metrics
  conditions:
    - reason: SucceededRescale
      message: the HPA controller was able to update the target scale to 3
```

But the actual Deployment will show only 1 replica running.

---

## Controller Log Evidence

From `temporal-worker-controller` logs, the interference is clear:

### Pattern: Controller resetting replicas
```json
{
  "level": "info",
  "msg": "scaling deployment",
  "controller": "temporalworkerdeployment",
  "TemporalWorkerDeployment": {
    "name": "temporal-worker-zili",
    "namespace": "byx-originacao"
  },
  "deployment": {
    "apiVersion": "apps/v1",
    "kind": "Deployment"
  },
  "deploymentError": "got runtime.Object without object metadata: &ObjectReference{...}",
  "replicas": 1
}
```

This repeats every 10-15 seconds, indicating the controller is:
1. Regularly reconciling the TemporalWorkerDeployment
2. Explicitly setting `replicas: 1` on the underlying Deployment
3. Failing due to an ObjectReference deserialization error (secondary issue)
4. But the `replicas: 1` field is still propagated despite the error

### Secondary errors in logs
```json
{
  "level": "info",
  "msg": "unknown field \"status.conditions\"",
  "controller": "temporalworkerdeployment"
}
```

This suggests the controller has issues with status field management, which may be related to the deployment error above.

---

## Root Cause Analysis

The controller's reconciliation logic appears to:
1. **Always set `spec.replicas` on the underlying Deployment** (defaulting to 1 or from TemporalWorkerDeployment spec)
2. **Not check if a WorkerResourceTemplate with HPA is managing this Deployment**
3. **Override HPA decisions on every reconciliation cycle**

This creates an irreconcilable conflict:
- **HPA**: "Target 3+ replicas based on metrics" → scales Deployment
- **Controller**: "TemporalWorkerDeployment says 1 replica" → resets Deployment
- **HPA**: Detects replicas < minReplicas → scales again
- **Loop**: Repeats indefinitely

---

## Questions for the Maintainers

1. Should the controller implement HPA-awareness (check for WorkerResourceTemplate presence and skip replica management)?
2. What is the intended workflow for autoscaling TemporalWorkerDeployments?
3. Can the ObjectReference deserialization error be addressed (secondary issue)?

---

### Option 1: Controller Detects HPA and Defers Replica Management (Recommended)
```go
if workerResourceTemplateExists(ctx, client, deployment) {
    // WorkerResourceTemplate is managing this deployment
    // Don't touch spec.replicas - let HPA manage it
    skip(deployment.Spec.Replicas)
} else {
    // Set replicas normally from TemporalWorkerDeployment spec
    deployment.Spec.Replicas = ptr.To(int32(twd.Spec.Replicas))
}
```

### Option 2: Explicit Validation
Reject TemporalWorkerDeployments that have both:
- No explicit `spec.replicas` AND
- An associated WorkerResourceTemplate with HPA

With a clear error: "Cannot use HPA-managed WorkerResourceTemplate without explicit replicas configuration"

### Option 3: User-facing Workaround Documentation
Document that HPA + TemporalWorkerDeployment currently requires:
- Explicit `spec.replicas` set to `minReplicas` value
- HPA configured to respect this as the baseline
- Understanding that controller will manage replicas, not HPA (defeats the purpose)

---

## Workarounds Attempted (All Failed)

1. ❌ **Omitting `spec.replicas`**: Controller defaults to 1, HPA can't override
2. ❌ **Setting `spec.replicas: 1`**: Controller locks it, HPA can't scale
3. ❌ **Setting `spec.replicas: 3`**: Controller overrides HPA, no dynamic scaling
4. ❌ **Setting `spec.replicas: null`**: Controller treats as 1

**Conclusion**: There is currently no working workaround. HPA + TemporalWorkerDeployment are fundamentally broken together.

---

## Impact

- **Affected Users**: Anyone attempting to use WorkerResourceTemplate with HPA for autoscaling
- **Scope**: All TemporalWorkerDeployment resources configured with HPA
- **Duration**: Deployment becomes unscalable; stuck at default replica count
- **Resource Waste**: Continuous HPA reconciliation cycles create unnecessary API load and logging overhead

---

## Additional Context

The HPA itself is functioning correctly:
- It successfully creates the HorizontalPodAutoscaler resource
- It correctly identifies when replicas are below minReplicas
- It successfully patches the Deployment to scale up
- **But**: Within milliseconds, the controller re-patches it back to 1

niyomukiza-mechack:
## What was changed

When an HPA exists via `WorkerResourceTemplate` the controller now skips enforcing the deployment's `spec.Replicas` of that version letting HPA manage the scaling.

## Changes
I added a wew helper function `hpaBuildIDSet()` in planner.go It detects HPA-kind WRTs and marks all active build IDs as HPA managed
I added `hpaBuildIDs` parameter on getScaleDeployments. It skips spec.Replicas enforcement for versions in this set
Target version : gated on `!hasHPA` for `spec.Replicas` enforcement, but preserves the 0 - >1 bootstrap since HPAs cannot scale a Deployment from zero
I added Logs  warning when spec.Replicas is set alongside an HPA, so a developer see the conflicting config instead of a silent fight

## Why?

A developer configured a WorkerResourceTemplate with an HPA (minReplicas: 3) on a WorkerDeployment that also had spec.replicas: 1
The controller enforced replicas: 1 every 15 seconds, the HPA scaled back to 3, creating an endless loop
The root cause was `getScaleDeployments` enforced a WD's `spec.Replicas` on every reconcile without checking whether an HPA is already managing that Deployment's replica count.

## Checklist

1. Closes #350

2. How was this tested:

Unit tests
Local e2e test (Kind cluster + Temporal Cloud)
Reproduced the bug and verified the fix on a Kind cluster
Before the fix, when I edited the WorkerDeployment's spec.Replicas to 1, the Controller locked replicas at 1 even thought I had an HPA with a minimum of 3 replicas and max 10. HPA was only used when WorkerDeployment's spec.Replicas was nil.

After this fix: Same spec.replicas: 1. Replicas stayed at 3 (HPA's value).

3. Any docs updates needed?

### DIFF
--- a/internal/planner/planner.go
+++ b/internal/planner/planner.go
@@ -157,8 +157,8 @@ func GeneratePlan(
 
 	// Add delete/scale operations based on version status
 	plan.DeleteDeployments = getDeleteDeployments(k8sState, status, spec, foundDeploymentInTemporal)
-	hpaBuildIDs := hpaBuildIDSet(wrts, k8sState)
-	plan.ScaleDeployments = getScaleDeployments(l, k8sState, status, spec, hpaBuildIDs)
+	hpaManaged := hasHPAScaler(wrts)
+	plan.ScaleDeployments = getScaleDeployments(l, k8sState, status, spec, hpaManaged)
 	plan.ShouldCreateDeployment = shouldCreateDeployment(status, maxVersionsIneligibleForDeletion)
 	plan.UpdateDeployments = getUpdateDeployments(k8sState, status, spec, connection)
 
@@ -736,18 +736,9 @@ func getDeleteDeployments(
 	return deleteDeployments
 }
 
-// hpaBuildIDSet returns a set of build IDs whose versions are managed by a
-// HorizontalPodAutoscaler through a WorkerResourceTemplate. When a version's build ID is in
-// this set, the controller must not manage that version's replica count, the HPA manages the scaling.
-//
-// A WRT renders one copy of its resource per active build ID (a build ID with a live Deployment in
-// k8sState), so an HPA-kind WRT marks every currently-active build ID.
-func hpaBuildIDSet(wrts []temporaliov1alpha1.WorkerResourceTemplate, k8sState *k8s.DeploymentState) map[string]bool {
-	hpaBuildIDs := make(map[string]bool)
-	if k8sState == nil {
-		return hpaBuildIDs
-	}
-
+// hasHPAScaler checks whether any WorkerResourceTemplate attaches a HorizontalPodAutoscaler to
+// this WorkerDeployment.When an HPA— the HPA manages the scaling not the Controller.
+func hasHPAScaler(wrts []temporaliov1alpha1.WorkerResourceTemplate) bool {
 	for i := range wrts {
 		// Parse kind from the WRT's spec.template
 		var templateMeta struct {
@@ -758,16 +749,11 @@ func hpaBuildIDSet(wrts []temporaliov1alpha1.WorkerResourceTemplate, k8sState *k
 		if err := json.Unmarshal(wrts[i].Spec.Template.Raw, &templateMeta); err != nil {
 			continue
 		}
-		if templateMeta.Kind != "HorizontalPodAutoscaler" {
-			continue
-		}
-
-		// This WRT is an HPA: it manages current active version's scaling
-		for buildID := range k8sState.Deployments {
-			hpaBuildIDs[buildID] = true
+		if templateMeta.Kind == "HorizontalPodAutoscaler" {
+			return true
 		}
 	}
-	return hpaBuildIDs
+	return false
 }
 
 // getScaleDeployments determines which deployments should be explicitly scaled and to what size.
@@ -778,7 +764,7 @@ func getScaleDeployments(
 	k8sState *k8s.DeploymentState,
 	status *temporaliov1alpha1.WorkerDeploymentStatus,
 	spec *temporaliov1alpha1.WorkerDeploymentSpec,
-	hpaBuildIDs map[string]bool,
+	hpaManaged bool,
 ) map[*corev1.ObjectReference]uint32 {
 	scaleDeployments := make(map[*corev1.ObjectReference]uint32)
 
@@ -786,10 +772,9 @@ func getScaleDeployments(
 	if status.CurrentVersion != nil && status.CurrentVersion.Deployment != nil {
 
 		// HPA exists — skip, don't touch replicas
-		if hpaBuildIDs[status.CurrentVersion.BuildID] {
+		if hpaManaged {
 			if spec.Replicas != nil {
-				l.Info("WorkerDeployment spec.Replicas is set but an HPA is managing this version's scaling; ignoring spec.Replicas",
-					"buildID", status.CurrentVersion.BuildID)
+				l.Info("HPA is managing this WorkerDeployment's scaling; ignoring spec.Replicas")
 			}
 		} else if spec.Replicas != nil {
 			// If spec.Replicas is non-nil, the controller is managing replicas instead of a scaler resource.
@@ -811,14 +796,12 @@ func getScaleDeployments(
 		status.TargetVersion.Deployment != nil {
 		if d, exists := k8sState.Deployments[status.TargetVersion.BuildID]; exists {
 
-			hasHPA := hpaBuildIDs[status.TargetVersion.BuildID]
-			if hasHPA && spec.Replicas != nil {
-				l.Info("WorkerDeployment spec.Replicas is set but an HPA is managing this version's scaling; ignoring spec.Replicas",
-					"buildID", status.TargetVersion.BuildID)
+			if hpaManaged && spec.Replicas != nil {
+				l.Info("HPA is managing this WorkerDeployment's scaling; ignoring spec.Replicas")
 			}
 
 			// Enforce spec.Replicas only when NO HPA manages this version — otherwise the HPA owns it.
-			enforceReplicas := spec.Replicas != nil && !hasHPA
+			enforceReplicas := spec.Replicas != nil && !hpaManaged
 			// Bootstrap 0 -> 1 regardless of HPA: an HPA cannot scale a Deployment up from zero, so if the
 			// Target Version was scaled to zero by Sunset Policy and nil replicas means a scaler owns it,
 			// the controller must bring it back to 1 for the HPA to take over.
--- a/internal/planner/planner.go
+++ b/internal/planner/planner.go
@@ -157,7 +157,8 @@ func GeneratePlan(
 
 	// Add delete/scale operations based on version status
 	plan.DeleteDeployments = getDeleteDeployments(k8sState, status, spec, foundDeploymentInTemporal)
-	plan.ScaleDeployments = getScaleDeployments(k8sState, status, spec)
+	hpaBuildIDs := hpaBuildIDSet(wrts, k8sState)
+	plan.ScaleDeployments = getScaleDeployments(l, k8sState, status, spec, hpaBuildIDs)
 	plan.ShouldCreateDeployment = shouldCreateDeployment(status, maxVersionsIneligibleForDeletion)
 	plan.UpdateDeployments = getUpdateDeployments(k8sState, status, spec, connection)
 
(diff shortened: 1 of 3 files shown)


internal/planner/planner.go
```diff
@@ -735,42 +736,96 @@ func getDeleteDeployments(
 	return deleteDeployments
 }
 
+// hpaBuildIDSet returns a set of build IDs whose versions are managed by a
```
jaypipes: As of v1.8.1 of temporal-worker-controller, there is no ability to have a `WorkerResourceTemplate` *per worker deployment version* (i.e. "buildID"), so I don't see the need to return a map of buildIDs that are managed by an HPA. Just return true or false if an HPA is managing the scale for this WD.

internal/planner/planner.go
```diff
@@ -735,42 +736,96 @@ func getDeleteDeployments(
...
 	scaleDeployments := make(map[*corev1.ObjectReference]uint32)
 
 	// Scale the current version if needed
 	if status.CurrentVersion != nil && status.CurrentVersion.Deployment != nil {
-		// If spec.Replicas is non-nil, the controller is managing replicas instead of a scaler resource.
-		// Scale the Current Version per the WorkerDeploymentSpec.Replicas value.
-		if spec.Replicas != nil {
+
+		// HPA exists — skip, don't touch replicas
+		if hpaBuildIDs[status.CurrentVersion.BuildID] {
```
jaypipes: See note above. This can just be a single boolean flag indicating whether HPA is managing the scaling or not for the WD as a whole.

internal/planner/planner.go
```diff
@@ -735,42 +736,96 @@ func getDeleteDeployments(
...
+			if spec.Replicas != nil {
+				l.Info("WorkerDeployment spec.Replicas is set but an HPA is managing this version's scaling; ignoring spec.Replicas",
```
```suggestion
				l.Info("HPA is managing this WorkerDeployment's scaling; ignoring spec.Replicas",
```

internal/planner/planner.go
```diff
@@ -735,42 +736,96 @@ func getDeleteDeployments(
...
 
-			// If the Target Version is an already-existing Deployment that was scaled to zero by the controller
-			// due to Sunset Policy, and the TWD has nil replicas because a scaler is managing the replicas, then
-			// no one will scale the Target Version back up, so we need to scale it back to 1 replica, which is what
-			// would happen if the Deployment was being created from scratch with nil replicas.
-			if spec.Replicas != nil || (spec.Replicas == nil && d.Spec.Replicas != nil && *d.Spec.Replicas == 0) {
-				replicas := int32(1) // just scale up to 1 if we are in the spec.Replicas == nil && d.Spec.Replicas == 0 case.
+			hasHPA := hpaBuildIDs[status.TargetVersion.BuildID]
+			if hasHPA && spec.Replicas != nil {
+				l.Info("WorkerDeployment spec.Replicas is set but an HPA is managing this version's scaling; ignoring spec.Replicas",
```
```suggestion
				l.Info("HPA is managing this WorkerDeployment's scaling; ignoring spec.Replicas",
```

internal/planner/planner.go
niyomukiza-mechack: Hmh that makes sense, single boolean is much simpler

carlydf:
> A developer configured a WorkerResourceTemplate with an HPA (minReplicas: 3) on a WorkerDeployment that also had spec.replicas: 1
> The controller enforced replicas: 1 every 15 seconds, the HPA scaled back to 3, creating an endless loop
The root cause was getScaleDeployments enforced a WD's spec.Replicas on every reconcile without checking whether an HPA is already managing that Deployment's replica count.

The behavior that you describe above is intentional and exactly what is [documented](https://github.com/temporalio/temporal-worker-controller/blob/main/api/v1alpha1/workerdeployment_types.go#L54-L57) on the `workerDeploymentSpec.replicas` field. We chose this because it is exactly how Kubernetes treats the `spec.replicas` field on the `Deployment` resource. See https://kubernetes.io/docs/concepts/workloads/autoscaling/horizontal-pod-autoscale/#migrating-deployments-and-statefulsets-to-horizontal-autoscaling:
> When an HPA is enabled, it is recommended that the value of spec.replicas of the Deployment and / or StatefulSet be removed from their [manifest(s)](https://kubernetes.io/docs/reference/glossary/?all=true#term-manifest). If this isn't done, any time a change to that object is applied, for example via kubectl apply -f deployment.yaml, this will instruct Kubernetes to scale the current number of Pods to the value of the spec.replicas key. This may not be desired and could be troublesome when an HPA is active, resulting in thrashing or flapping behavior.

carlydf:
Also, if we decide to merge this PR, it should not "close #350"

The problem reported in #350 says:
> ❌ Setting `spec.replicas: null`: Controller treats as 1

which is different from the behavior that this PR describes and changes. 

(To be clear, I have not been able to reproduce the behavior described in #350, and the reporter hasn't gotten back to us with instructions to repro either, so I am not convinced we need to fix anything for #350)

PR #528, opened 2026-08-17
feat: add payload encoding to gate workflow

### LINKED ISSUE #414: [Feature Request] support json/protobuf payloads for workflow gate

Users using protobuf messages for workflow inputs want to be able to use protobuf messages as inputs to the workflow gate. json/protobuf is supported by things like the `temporal` cli and ui, and should also be supported by the worker controller.

Add an additional optional field called `ProtoMessageType` to `GateWorkflowConfig` (and the WorkerDeployment CRD). When this field is populated alongside an input payload, the client should send the payload as a json/protobuf payload, instead of a json/plain payload, like this

```go
var arg interface{}
if len(wf.input) > 0 {
    if wf.protoMessageType != "" {
        arg = converter.NewRawValue(&commonpb.Payload{
            Metadata: map[string][]byte{
                converter.MetadataEncoding:    []byte(converter.MetadataEncodingProtoJSON),
                converter.MetadataMessageType: []byte(wf.protoMessageType),
            },
            Data: wf.input,
        })
    } else {
        arg = json.RawMessage(wf.input)
    }
}
```

### Additional context

We are happy to help contribute this if we can agree on the shape of the change

zainawaisn:
## What was changed

Adds the option to specify what encoding to use on the payload for the gate workflow. The existing functionality only supports `json/plain`, as that is the default the Temporal SDK uses when encoding is unspecified. The CRD now can take two additional fields under the gate:

```
gate:
  workflowType: "HealthCheck"
  encoding: json/protobuf # new
  messageType: my.package.DeployRequest   # new
  input:
    service: checkout
    replicas: 3
```

`encoding` can take the value of: `json/plain`, `json/protobuf`, `binary/plain`, and `binary/protobuf`. If using `binary/plain` or `binary/protobuf` encodings, then inline input cannot be used (a Secret or ConfigMapKey must be used instead). If encoding is omitted, `json/plain` will be used, preserving existing functionality and CRDs.

These reflect the values of the Temporal SDK encoding enum. The only one excluded is `binary/null`, which is the encoding label the SDK uses when the payload is non-existent. As omitting the input fields on the CRD means there is no payload, this encoding is redundant and therefore not used.

`messageType` is the fully qualified protobuf message name. Not needed for Temporal workers, but is used for tools that see the payload (Temporal UI for example). This field is required for the `binary/protobuf` encoding and optional for `json/protobuf`. Disallowed for all other encodings, as it is unnecessary.

The CRD templates also drop workloadRef and add schedulingGroup under the pod spec. This is unrelated pod-schema churn from the k8s.io/api v0.36.3 bump in #521, which did not regenerate manifests.

## Why?

Addresses #414. Protobuf messages can now be used by specifying the `json/protobuf` encoding. Adds support for the other Temporal SDK encoding types as well. `messageType` was added to be more in line with the payload labels the Temporal SDK already uses, and to enable visibility on certain payload types on the UI.

## Checklist

1. Closes #414 

2. How was this tested:
`make test-unit` passes (includes envtest). New coverage:

- **Payload construction** (`internal/controller/execplan_test.go`) — 4 tests on
  `gateWorkflowArg`: unset encoding still produces `json/plain` with the input verbatim, each
  enum value reaches the payload untouched, and `messageType` is recorded when set / its key
  omitted when not. Asserted through the SDK's default data converter, so these check the
  payload a worker actually receives.
- **Field propagation** — 2 tests covering spec → `WorkflowConfig` → `startWorkflowConfig`,
  where a missed assignment would silently drop the encoding.
- **Validation** — 8 new CEL specs against a real API server via envtest, and 14 gate cases in
  the webhook table, covering the `messageType` rules plus the binary-encoding input-source rule
  (Secret accepted; inline `input` or no input rejected).

Also: mutation-tested the new webhook rule (forcing it off makes the rejection cases fail, so
they aren't vacuous); `make manifests generate` leaves no further diff.

4. Any docs updates needed?
changes in docs/configuration.md


api/v1alpha1/workerdeployment_cel_validation_test.go
```diff
@@ -115,4 +116,100 @@ var _ = Describe("WorkerDeployment CRD CEL validation", func() {
...
+		g.InputFrom = &GateInputSource{
+			SecretKeyRef: &corev1.SecretKeySelector{
+				LocalObjectReference: corev1.LocalObjectReference{Name: "gate-input"},
+				Key:                  "request.bin",
+			},
+		}
+	}
+
+	const messageTypeRequiredErr = "gate.messageType is required when gate.encoding is binary/protobuf"
+	const messageTypeNotAllowedErr = "gate.messageType may only be set when gate.encoding is json/protobuf or binary/protobuf"
```
jaypipes: kinda odd to have these strings defined as constants in the test file instead of in `api/v1alpha1/workerdeployment_webhook.go` where they are hard-coded strings in the construction of the errors :) Might as well make package-level variables called `ErrMessageTypeRequired` and `ErrMessageTypeNotAllowed` in `api/v1alpha1/workerdeployment_webhook.go` and then you can simply assert below whether the returned error is that exported variable or not.

internal/controller/execplan.go
```diff
@@ -129,9 +131,17 @@ func (r *WorkerDeploymentReconciler) startTestWorkflows(ctx context.Context, l l
...
+			if wf.messageType != "" {
+				gl = gl.WithValues("messageType", wf.messageType)
+			}
 			if wf.isInputSecret {
 				// Don't log the actual input if it came from a Secret
-				l.Info("starting gate workflow",
+				gl.Info("starting gate workflow",
 					"workflowType", wf.workflowType,
 					"taskQueue", wf.taskQueue,
 					"buildID", wf.buildID,
```
jaypipes: I suppose you could have moved these three key/value pairs up to the `gl := l.WithValues(...)` line above.

internal/controller/execplan.go
```diff
@@ -151,7 +161,7 @@ func (r *WorkerDeploymentReconciler) startTestWorkflows(ctx context.Context, l l
 				}
 
 				// Log the input keys for non-secret sources (inline or ConfigMap)
-				l.Info("starting gate workflow",
+				gl.Info("starting gate workflow",
 					"workflowType", wf.workflowType,
 					"taskQueue", wf.taskQueue,
 					"buildID", wf.buildID,
```
jaypipes: ditto.

internal/controller/execplan.go
jaypipes: ditto.

jaypipes (approved):
Honestly, I think this is a great PR and though I went through it with a fine-tooth review comb, was only able to come up with a single very small nit/suggestion.

Well played @zainawaisn.

internal/controller/execplan.go
zainawaisn: done

internal/controller/execplan.go
zainawaisn: done.

internal/controller/execplan.go
zainawaisn: done.

Issue #465, opened 2026-07-27
chart: add extraVolumes / extraVolumeMounts hooks to the manager Deployment


nishijun:
Tracking issue for the chart-side piece of the Connection cleanup series
(#449 / #311), split out at @jaypipes' suggestion.

## Problem

Once the controller can take credentials from something other than a Kubernetes
Secret, the file has to reach the manager pod somehow. Today the chart offers no
way to attach one: `helm/temporal-worker-controller/values.yaml` (v1.8.0) has no
`extraVolumes` / `extraVolumeMounts` (or equivalent) keys, and the manager
Deployment template mounts only the webhook certificate.

That blocks the most useful delivery mechanism for the file-based path — a CSI
provider such as the GCP Secret Manager CSI driver or spiffe-helper's shared
volume — since neither can be attached without editing the rendered manifest.

## Proposal

Expose pass-through `extraVolumes` and `extraVolumeMounts` values on the manager
Deployment, so operators can attach an arbitrary volume (CSI, projected token,
emptyDir shared with a sidecar, etc.) without forking the chart.

## Related

- #449 — file-based / external credentials (API key)
- #311 — file-based mTLS credentials
- #452 — first PR of the Connection cleanup series

PR #532, opened 2026-08-19
helm: support extraEnv, extraVolumes and extraVolumeMounts on the manager

aldredb:
Let operators mount a private CA bundle into the controller so it can reach a Temporal
endpoint whose certificate is signed by an internal authority.

- add `extraEnv`, `extraVolumes` and `extraVolumeMounts` to the manager pod
- all three default to `[]`, leaving rendered output unchanged when unset
- without them the controller fails with `x509: certificate signed by unknown authority`,
  as the chart hardcodes `volumes:` to the webhook cert and exposes no env knob
- resolves the existing asymmetry where `kubeRBACProxy.volumeMounts` is settable but no
  supported option could supply the volume

Issue #465


jaypipes (approved):
@aldredb this looks awesome, thank you so much! :)

The JSONSchema for the values.yaml file has gotten out of sync with new additions over time and I've [filed a new issue to automate keeping that in sync](https://github.com/temporalio/temporal-worker-controller/issues/536). If you're interested in working on that, let me know!

helm/temporal-worker-controller/values.yaml
```diff
@@ -185,6 +185,40 @@ nodeSelector: {}
...
+extraVolumes: []
+
+# extraVolumeMounts adds volume mounts to the manager container. Volumes must be
+# supplied via extraVolumes above.
+#
+# extraVolumeMounts:
+#   - name: custom-ca
+#     mountPath: /etc/ssl/custom-ca
+#     readOnly: true
+extraVolumeMounts: []
```
jlegrone, line 220: There should be corresponding changes in `values.schema.json`

aldredb: Added in 6764fbc.

PR #544, opened 2026-08-21
fix(controller): delete rendered autoscalers at sunset scale-down

### LINKED ISSUE #529: Remove KEDA ScaledObject WRTs from drained versions
> We measured ~120 autoscaler raises against 122 controller re-zeroes in one 30-minute window on two drained versions. 

Did you experience this with KEDA ScaledObject or with HPA?

HPA is not expected to scale back up versions that have been explicitly scaled to zero by the controller (which the controller does after version is drained + sunset policy scaledown delay). See https://kubernetes.io/docs/concepts/workloads/autoscaling/horizontal-pod-autoscale/#implicit-maintenance-mode-deactivation for how that works with HPA.

Based on some brief agent conversation, it seems like KEDA does not handle this the same way. So maybe you experienced this with KEDA? I think to solve this issue, the worker controller needs to pause the KEDA ScaledObject right before scaling down a Deployment due to sunset policy, and remove the pause annotation if there is a roll back and that drained version becomes the target version again. See https://keda.sh/docs/2.20/concepts/scaling-deployments/#pausing-autoscaling

_Originally posted by @carlydf in https://github.com/temporalio/temporal-worker-controller/pull/498#discussion_r3797984759_

tomba7:
Fixes #529 (which is a sub-issue of #439)

This issue was discussed during the TWC community meeting on Aug 19 2026 where various ways to address the issue was debated. For details on the those, please refer to the comments on #529 (will add shortly) which also explains why the following fix was recommended

## What was changed

The controller now deletes a version's rendered autoscaler as soon as it begins pinning that version's replicas to zero for sunset, instead of keeping it until the versioned k8s Deployment is deleted.

- k8s.HasScaleTarget is a kind-agnostic check for a scaleTargetRef anywhere under spec
- getSunsetScaleDownBuildIDs gets drained versions past `scaledownDelay`, matching the condition the existing scale-to-zero path uses and are the candidates for autoscaler deletion
- getDeleteWorkerResources now deletes those autoscalers even though the k8s Deployment is still live
- getWorkerResourceApplies now filters out the re-render so the autoscaler stays gone

Only WRTs declaring a scaleTargetRef are affected. Every other rendered resource is cleaned up with the Deployment as before.

Lifecycle for a drained version:
1. drained: no open workflows pinned to it
2. drained + `scaledownDelay`: controller pins replicas to 0 and deletes the autoscaler
3. drained + `scaledownDelay` + `deleteDelay`: k8s Deployment deleted

## Why?

Both the controller and KEDA write the replica count which creates a scaling loop (0 -> N -> 0 -> ...)

- After scaledownDelay the controller holds a drained version at 0. KEDA restores minReplicaCount on its next poll, the controller re-zeroes, and this repeats for the whole deleteDelay window, 24h by default.
- Measured over 30 minutes on two drained versions (reported in #439): 120 keda-operator raises against 122 controller re-zeroes, each flap creating, scheduling and deleting a real pod.

## Checklist

1. How was this tested:

   Unit tests 
   - `go test ./internal/planner/... ./internal/k8s/...`
   - `TestHasScaleTarget`
   - `TestGetSunsetScaleDownBuildIDs`
   - `TestGetWorkerResourceApplies_SunsetBuildIDs`
   - `TestGetDeleteWorkerResources_SunsetBuildIDs`

   Manual testing on a local cluster
   - kind cluster with KEDA installed, controller built and deployed from this branch
   - a `WorkerDeployment` plus a WRT rendering a KEDA `ScaledObject` with
     `minReplicaCount: 1`, and short `scaledownDelay` + `deleteDelay`
   - rolled a new version so the previous one drains
   - before the fix: once the controller zeroed the drained version, KEDA
     raised it back within one `pollingInterval` and the controller re-zeroed which created
     a continuous pod create/delete cycle lasting the whole `deleteDelay` window
   - after the fix: the drained version's `ScaledObject` is deleted at `scaledownDelay`,
     replicas stay at 0, and no further flapping happens. After `deleteDelay` the k8s Deployment
     was deleted 

2. Any docs updates needed?

   Potentially `docs/worker-resource-templates.md` but will iterate based on review feedback on this PR.

### DIFF
--- a/internal/k8s/workerresourcetemplates.go
+++ b/internal/k8s/workerresourcetemplates.go
@@ -309,6 +309,49 @@ func isEmptyMap(v interface{}) bool {
 	return ok && len(m) == 0
 }
 
+// HasScaleTarget reports whether a WRT's raw template declares a scaleTargetRef, the
+// kind-agnostic signal for a resource that drives the versioned Deployment's replica
+// count. It searches the same spec subtree injectScaleTargetRefRecursive injects into
+func HasScaleTarget(raw []byte) bool {
+	if len(raw) == 0 {
+		return false
+	}
+	var obj map[string]interface{}
+	if err := json.Unmarshal(raw, &obj); err != nil {
+		return false
+	}
+	spec, ok := obj["spec"].(map[string]interface{})
+	if !ok {
+		return false
+	}
+	return hasScaleTargetRef(spec)
+}
+
+// hasScaleTargetRef walks and reports whether a scaleTargetRef key
+// exists anywhere in the tree
+func hasScaleTargetRef(obj map[string]interface{}) bool {
+	for k, v := range obj {
+		if k == "scaleTargetRef" {
+			return true
+		}
+		switch nested := v.(type) {
+		case map[string]interface{}:
+			if hasScaleTargetRef(nested) {
+				return true
+			}
+		case []interface{}:
+			for _, item := range nested {
+				if nestedItem, ok := item.(map[string]interface{}); ok {
+					if hasScaleTargetRef(nestedItem) {
+						return true
+					}
+				}
+			}
+		}
+	}
+	return false
+}
+
 // buildScaleTargetRef constructs the scaleTargetRef map pointing at the versioned Deployment.
 func buildScaleTargetRef(deploymentName string) map[string]interface{} {
 	return map[string]interface{}{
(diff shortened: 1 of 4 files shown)


jaypipes (approved):
@tomba7 nice work on this. honestly, can't see anything wrong with it at all.

PR #546, opened 2026-08-24
Support a trusted CA for API-key and no-credentials Connections

aldredb:
Let API-key and no-credentials Connections trust a private CA, matching what mutual-TLS Connections already get via MutualTLSSecretRef's own ca.crt key.

- add `ConnectionTLSConfig.CACertSecretRef`, applicable when the auth mode is API key or no credentials
- extract the additive CA-pool logic already used for mTLS (`fetchClientUsingMTLSSecret`) into a shared `TLSCertPool` helper and reuse it across auth modes
- error out (not silently fall back to system-only trust) when the referenced secret has no `ca.crt` key
- include the new field in `ClientPoolKey` so the client cache invalidates correctly when the CA changes

Without this, the only way to reach a Temporal server on a privately-signed certificate over API-key auth is a process-wide `SSL_CERT_FILE` override on the controller — which *replaces* rather than extends the system trust store, and affects every Connection the controller manages, not just the one that needs it.


tomba7:
Hi @aldredb thank you for this PR. We'll get it reviewed as soon as we can. Do you have an issue open for this by any chance?

api/v1alpha1/connection_types.go
```diff
@@ -38,10 +38,18 @@ type ConnectionTLSConfig struct {
 	// +optional
 	// +kubebuilder:validation:Pattern=`^[a-zA-Z0-9.-]+$`
 	ServerName string `json:"serverName,omitempty"`
+
+	// CACertSecretRef references a Secret (key "ca.crt") whose certificate is appended to
+	// the system trust store for this connection, independent of AuthMode.
```
jaypipes: I don't think "independent of AuthMode" is true in the comment above?

If the auth mode is mTLS, CACertSecretRef should not be set, according to the validation rule below...

internal/controller/clientpool/clientpool.go
```diff
@@ -261,6 +290,29 @@ func (cp *ClientPool) ParseClientSecret(
 		}
 	}
 
+	// TLS.CACertSecretRef is independent of AuthMode, so it applies here regardless of
```
jaypipes: > TLS.CACertSecretRef is independent of AuthMode

Again, I don't think that's true, as you allude to in the following line:

> AuthModeTLS ignores it ... and the two are mutually exclusive by CEL validation

I think a more accurate comment might be:

> TLS.CACertSecretRef is applicable when AuthMode is either `AuthModeAPIKey` or `AuthModeNoCredentials`

api/v1alpha1/connection_types.go
```diff
@@ -124,6 +132,15 @@ func (s ConnectionSpec) TLSServerName() string {
 	return s.TLS.ServerName
 }
 
+// CACertSecretName returns the name of the Secret referenced by TLS.CACertSecretRef, or an
```
jaypipes: To be consistent with the other helper (`TLSServerName()`), this should be named `TLSCACertSecretName()`.

api/v1alpha1/connection_types.go
```diff
@@ -38,10 +38,18 @@ type ConnectionTLSConfig struct {
...
+	// the system trust store for this connection, independent of AuthMode.
+	// The Secret must be of type Opaque or
+	// kubernetes.io/tls and exist in the same Kubernetes namespace as the Connection.
+	// +optional
+	CACertSecretRef *SecretReference `json:"caCertSecretRef,omitempty"`
 }
 
 // ConnectionSpec defines the desired state of Connection
 // +kubebuilder:validation:XValidation:rule="!(has(self.mutualTLSSecretRef) && has(self.apiKeySecretRef))",message="Only one of mutualTLSSecretRef or apiKeySecretRef may be set"
+// +kubebuilder:validation:XValidation:rule="!(has(self.mutualTLSSecretRef) && has(self.tls) && has(self.tls.caCertSecretRef))",message="tls.caCertSecretRef cannot be combined with mutualTLSSecretRef; bundle the CA into that secret's own ca.crt key instead"
```
jaypipes, line 54: these CEL validation rules are getting a bit too hairy for my taste...

internal/controller/clientpool/clientpool.go
```diff
@@ -23,11 +23,12 @@ import (
...
-	HostPort      string
-	TLSServerName string
-	Namespace     string            // Temporal namespace
-	SecretName    string            // Include secret name in key to invalidate cache when the secret name changes
-	AuthMode      v1alpha1.AuthMode // Include auth mode in key to invalidate cache when the auth mode changes for the secret
+	HostPort         string
+	TLSServerName    string
+	Namespace        string            // Temporal namespace
+	SecretName       string            // Include secret name in key to invalidate cache when the secret name changes
+	CACertSecretName string            // Include CA secret name in key to invalidate cache when TLS.CACertSecretRef changes
```
jaypipes: Should be `TLSCACertSecretName` to be consistent with `TLSServerName`

internal/controller/clientpool/clientpool.go
```diff
@@ -244,6 +255,24 @@ func (cp *ClientPool) fetchClientUsingNoCredentials(opts NewClientOptions) (*sdk
 	return &clientOpts, &key, &auth, nil
 }
 
+// mergeCACert returns the system CA pool with caCert appended, so a connection can trust a
```
jaypipes: Recommend renaming this method to `TLSCertPool()` to make it clear the method returns a CertPool.

api/v1alpha1/connection_types.go
```diff
@@ -38,10 +38,18 @@ type ConnectionTLSConfig struct {
 	// +optional
 	// +kubebuilder:validation:Pattern=`^[a-zA-Z0-9.-]+$`
 	ServerName string `json:"serverName,omitempty"`
+
+	// CACertSecretRef references a Secret (key "ca.crt") whose certificate is appended to
```
jaypipes:
```suggestion
	// CACertSecretRef references a Secret with a required "ca.crt" key whose certificate is appended to
```

jaypipes (changes requested):
Thanks @aldredb! Mostly minor suggestions and renames, please see inline!

api/v1alpha1/connection_types.go
aldredb: I'm following the existing validation rule pattern in this PR. May I know what is the action item here?

aldredb:
@jaypipes thanks for the review — addressed in a762022:

- renamed `CACertSecretName` → `TLSCACertSecretName` and `mergeCACert` → `TLSCertPool` for consistency with `TLSServerName`
- replaced the "independent of AuthMode" wording with the applicable auth modes, in both the field doc and `ParseClientSecret`
- applied the `ca.crt` doc suggestion

I left a question on the CEL validation thread. Ready for another look when you get a chance.

api/v1alpha1/connection_types.go
jaypipes: Oh, no, nothing you need to address in this PR, @aldredb :) I was just lamenting the length of the validation expression...

jaypipes (approved):
👍 thanks @aldredb!

Shivs11 (approved):
did a quick pass, it looks g to me

Issue #456, opened 2026-07-23
[Feature Request] Support ClusterConnection CRD


brandonbetsson:
### Is your feature request related to a problem? Please describe.

Today, Temporal Worker Controller requires a connection object to exist in each namespace where a WorkerDeployment is created. In a multi-namespace setup, this means duplicating the same Temporal connection configuration many times.

My use case is to define one shared connection centrally for all namespaces, rather than managing identical connection resources in every namespace.

### Describe the solution you'd like

I would like support for a single global or cluster-scoped connection resource, similar in concept to `VaultAuthGlobal`.

A possible design could be a new resource such as:

`Kind: ClusterConnection`

Then `WorkerDeployment` could reference either a namespaced connection or a global/cluster-scoped connection. `WorkerDeployment` could support a `connectionRef` that includes both `name` and `kind`, where `kind` is either:

- `Connection`
- `ClusterConnection`

Example API shape:

```
spec:
  connectionRef:
    name: shared-temporal-prod
    kind: ClusterConnection
```

Example `ClusterConnection`:

```
apiVersion: temporal.io/v1alpha1
kind: ClusterConnection
metadata:
  name: shared-temporal-prod
spec:
  hostPort: "production.abc123.tmprl.cloud:7233"
  # Optional: override the TLS server name used for certificate verification
  tls:
    serverName: "production.abc123.tmprl.cloud"
  mutualTLSSecretRef: 
    name: temporal-cloud-mtls  # Optional: for mTLS
    namespace: temporal-system
```

Example `WorkerDeployment` using `ClusterConnection`:

```
apiVersion: temporal.io/v1alpha1
kind: WorkerDeployment
metadata:
  name: payments-worker
  namespace: payments
spec:
  connectionRef:
    name: shared-temporal-prod
    kind: ClusterConnection
    # use kind: Connection for a namespace-scoped connection
    # or omit kind if you want Connection to remain the default
  taskQueue: payments
  buildId: payments-worker-v1
  replicas: 2
```

This would preserve backward compatibility while allowing platform teams to define one shared Temporal connection centrally and reuse it across namespaces.

### Additional context

Even with a ClusterConnection, worker pods still cannot directly mount or reference secrets from another namespace. So while this would solve the duplicated connection object problem, it would not fully address cross-namespace credential reuse on its own.

jaypipes:
[...]

PR #549, opened 2026-08-24
FEATURE: Add ClusterConnection cluster-scoped resource

niyomukiza-mechack:
## What was changed

- `api/v1alpha1/clusterconnection_types.go`: Added a new cluster-scoped type reusing
  ConnectionSpec and ConnectionStatus
- `workerdeployment_types.go`: added optional Kind field to ConnectionReference
  (default: Connection, backward compatible)
- `worker_controller.go` :
     - Added a new function `resolveConnection()` helper for kind-aware fetch,
           generalized `ensureConnectionFinalizer` to client.Object, and cluster-wide
     `removeConnectionFinalizerIfUnused` (cross-namespace list + kind-aware match), and 
     `findTWDsUsingClusterConnection` mapper. Also added ClusterConnection watch + RBAC
    - Migration path: When a WorkerDeployment's connectionRef changes (name or kind), the controller now releases     the finalizer from the previously referenced connection if no other WorkerDeployment still uses it.
    - Added ObservedConnectionRef to WorkerDeploymentStatus to detect the change, sameConnectionRef() with kind normalization, and refactored the fetch/release into getConnectionByRef() and releaseConnectionFinalizerIfUnused(). This also closes the pre-existing namespaced Connection name-change leak.

## Why?

Context: https://github.com/temporalio/temporal-worker-controller/issues/456
Today, TWC requires a Connection resource in every namespace. When there's many namespaces, there's a lot of duplication for Connection Resource. `ClusterConnection` lets you define the connection once, cluster-wide, and have WorkerDeployments in any namespace reference it.  The credential Secret still lives per-namespace since a pod can only mount a Secret from its own namespace.

## Checklist

1. Closes 

2. How was this tested:

Unit tests in `internal/controller/clusterconnection_test.go`

Local Kind cluster:
- Created 5 tenant namespaces, one ClusterConnection, one Secret per namespace, one WorkerDeployment per namespace all referencing the shared connection
- All 5 reconciled and connected through the single ClusterConnection
- Deleted one tenant (4 remaining) and ClusterConnection finalizer correctly kept
- Confirmed kubectl delete clusterconnection blocks while any WD references it (finalizer protection works)

##Migration path testing (the kind-change scenario):
- Created WDs pointing at namespaced Connections, then migrated to the ClusterConnection via kubectl patch
- Confirmed the new ClusterConnection gets the finalizer correctly
- Confirmed the old namespaced Connection's finalizer is released automatically, so it becomes freely deletable
- Confirmed the safety case: migrating a WD away from a shared ClusterConnection correctly keeps the finalizer while a WD in another namespace still references it

3. Any docs updates needed?


helm/temporal-worker-controller/templates/rbac.yaml
```diff
@@ -110,6 +110,7 @@ rules:
   - apiGroups:
       - temporal.io
     resources:
+      - clusterconnections
```
tomba7: Are we sure `clusterconnections` belong here under the `Role`? Since its cluster scoped it seems like it should be associated only with a `ClusterRole`?

api/v1alpha1/workerdeployment_types.go
```diff
@@ -223,6 +236,12 @@ type WorkerDeploymentStatus struct {
 	// Conditions represent the latest available observations of the WorkerDeployment's current state.
 	// +optional
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+	// ObservedConnectionRef records the connectionRef the controller last
```
jaypipes: This field is not needed. See comment below about how you can fetch the previous desired state (spec) for the `WorkerDeployment` CR and determine if the connection kind has changed.

api/v1alpha1/workerdeployment_types.go
```diff
@@ -12,17 +12,30 @@ import (
...
 	// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`
 	Name string `json:"name"`
+	// Kind is the type of connection resource.
+	// ClusterConnection selects a cluster-scoped ClusterConnection.
+	// Omitting this field preserves the pre-existing behavior of referencing a
+	// namespaced Connection.
+	// +optional
+	// +kubebuilder:default=Connection
+	// +kubebuilder:validation:Enum=Connection;ClusterConnection
+	Kind string `json:"kind,omitempty"`
```
jaypipes: As of August, 2026 we only have a single API version (`v1alpha1`). In the future, however, we will likely have other API versions and potentially other API groups.

Best to think long-term and let's call this field `ObjectRef` and make it a pointer to a [`corev1.TypedObjectReference`](https://pkg.go.dev/k8s.io/kubernetes/pkg/apis/core#TypedObjectReference).

TypedObjectReference contains all the type information we might need in the future (for both cluster and namespace-scoped resources).

Another benefit to using TypedObjectReference here is that you can make the `Name` field optional and put a validation rule in place that says either `Name` or `ObjectRef` must be populated. And you can then remove some of the messy conditionals below in `releaseConnectionFinalizerIfUnused` where you are checking the `Name` field and trying to determine if the name refers to a Connection or a ClusterConnection...

internal/controller/worker_controller.go
```diff
@@ -778,74 +788,156 @@ func (r *WorkerDeploymentReconciler) recordWarningAndSetBlocked(
...
+	return conn.Spec, &conn, nil
+}
+
+// resolveConnection fetches the connection resource referenced by the
+// WorkerDeployment's current connectionRef.
+func (r *WorkerDeploymentReconciler) resolveConnection(
+	ctx context.Context,
+	workerDeploy *temporaliov1alpha1.WorkerDeployment,
+) (temporaliov1alpha1.ConnectionSpec, client.Object, error) {
+	return r.getConnectionByRef(ctx, workerDeploy.Spec.WorkerOptions.ConnectionRef, workerDeploy.Namespace)
```
jaypipes: Is there a need for a `resolveConnection` helper method that is one line that calls the `getConnectionByRef` helper method? If not, recommend removing this level of indirection.

internal/controller/worker_controller.go
```diff
@@ -228,13 +227,25 @@ func (r *WorkerDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
-	if err := r.ensureConnectionFinalizer(ctx, l, &connection); err != nil {
+	connection := temporaliov1alpha1.Connection{Spec: connSpec}
+
+	// Ensure the finalizer is on the connection object so it cannot be deleted
+	// while this WD still references it.
+	if err := r.ensureConnectionFinalizer(ctx, l, connObj); err != nil {
 		return ctrl.Result{}, err
 	}
 
+	// If connectionRef changed since we last finalized (name or kind), release the
```
```suggestion
	// If connectionRef changed since we last reconciled (name or kind), release the
```

internal/controller/worker_controller.go
```diff
@@ -228,13 +227,25 @@ func (r *WorkerDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+	// finalizer from the previously-referenced connection if no other WD uses it.
+	// The new connection is already protected by ensureConnectionFinalizer above,
+	// so the WD is never left unprotected.
+	current := workerDeploy.Spec.WorkerOptions.ConnectionRef
```
jaypipes: Note that you can avoid the below code with the following guard:

```go
// If the desired state of the resource changed, check that any changes
// to the referenced connection type are handled.
if workerDeploy.Generation != workerDeploy.Status.ObservedGeneration {

}
```

internal/controller/worker_controller.go
```diff
@@ -341,6 +352,7 @@ func (r *WorkerDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	}
 	// Preserve conditions that were set during this reconciliation
 	status.Conditions = workerDeploy.Status.Conditions
+	status.ObservedConnectionRef = workerDeploy.Spec.WorkerOptions.ConnectionRef.DeepCopy()
```
jaypipes, line 364: 👍

internal/controller/worker_controller.go
```diff
@@ -778,74 +788,156 @@ func (r *WorkerDeploymentReconciler) recordWarningAndSetBlocked(
...
+		controllerutil.RemoveFinalizer(connObj, finalizerName)
+		if err := r.Update(ctx, connObj); err != nil {
+			return fmt.Errorf("unable to remove finalizer from connection %q: %w", ref.Name, err)
 		}
 	}
 
 	return nil
 }
 
+// removeConnectionFinalizerIfUnused releases the finalizer from the WD's current
```
jaypipes: I don't think there's a good reason to have this helper method that is just a one-liner that calls another (very similarly-named) helper method. I would prefer to just call `releaseConnectionFinalizerIfUnused` directly.

jaypipes (changes requested):
Thanks @niyomukiza-mechack! I've left some thoughts for you inline. I think we want to think about the structure of the `ConnectionRef` object in a bit of a longer-term view.

helm/temporal-worker-controller/templates/rbac.yaml
niyomukiza-mechack: Good catch! Should be fixed. `clusterconnections` was landing in the namespaced Role because `hack/sync-rbac-rules.py` filtered at block granularity, and `controller-gen` merged `clusterconnections` into the same block as the namespaced connections

api/v1alpha1/workerdeployment_types.go
niyomukiza-mechack: Super helpful and totally makes sense!

internal/controller/worker_controller.go
niyomukiza-mechack: removed

hack/sync-rbac-rules.py
```diff
@@ -65,20 +70,29 @@ def filter_namespaced(rules_text):
     if current:
         blocks.append("".join(current))
 
-    filtered = []
+        filtered = []
```
jaypipes: AFAICT, the above line is incorrect. `filtered` should be initialized one indentation less.

internal/controller/worker_controller.go
```diff
@@ -778,68 +791,132 @@ func (r *WorkerDeploymentReconciler) recordWarningAndSetBlocked(
...
 	}
 
-	// Check if any other WD (not the one being deleted) references this connection
 	for i := range wds.Items {
 		wd := &wds.Items[i]
-		if wd.Name == deletingWD.Name {
+		// Skip self by namespace and name: under a cluster-wide list, two WDs in
+		// different namespaces can share the same name, so name alone is not a
+		// unique identity.
+		if wd.Namespace == selfNamespace && wd.Name == selfName {
```
jaypipes, line 920: Seems to me this is a latent bug in the existing codebase, no? If so, good on you for catching it.

internal/controller/worker_controller.go
```diff
@@ -778,68 +791,132 @@ func (r *WorkerDeploymentReconciler) recordWarningAndSetBlocked(
...
+func connectionRefName(ref temporaliov1alpha1.ConnectionReference) string {
+	if ref.ObjectRef != nil {
+		return ref.ObjectRef.Name
+	}
+	return ref.Name
+}
+
+// sameConnectionRef reports whether two connectionRefs resolve to the same connection.
+func sameConnectionRef(a, b temporaliov1alpha1.ConnectionReference) bool {
+	return connectionRefName(a) == connectionRefName(b) && connectionRefIsCluster(a) == connectionRefIsCluster(b)
```
eniko-dif, line 839: Should namespace be checked here too? If neither of them are cluster connections, they might be in different namespaces with the same name.

internal/controller/worker_controller.go
```diff
@@ -944,6 +1023,29 @@ func (r *WorkerDeploymentReconciler) findTWDsUsingConnection(ctx context.Context
 	return requests
 }
 
+func (r *WorkerDeploymentReconciler) findTWDsUsingClusterConnection(
```
eniko-dif, line 1060: Might worth mentioning that depending on the number of WDs, this can put more load on the controller.

internal/controller/worker_controller.go
niyomukiza-mechack: Yes. I caught it because I accidentally introduced the same bug myself and took me a while to figure out what was happening😄

internal/controller/worker_controller.go
niyomukiza-mechack: Good question! So `sameConnectionRef` is only used in one place: comparing observed (from this WD's status) vs current (from this WD's spec) during migration detection. So both refs are on the same WorkerDeployment, so they always resolve in the same namespace. 
The migration we want to detect here is when a WD's connectionRef changes. whether that's the kind `Connection` to  `ClusterConnection` or the name.

internal/controller/worker_controller.go
niyomukiza-mechack: Good callout!

jaypipes (approved):
👍 thank you @niyomukiza-mechack :)

PR #553, opened 2026-08-27
fix(controller): prune worker deployment version before deleting its k8s Deployment

### LINKED ISSUE #543: [Bug] A failed version prune in deleteDrainedVersions orphans the Temporal version record permanently (no requeue)
## Problem

`deleteDrainedVersions` (added in #498, shipped in v1.9.0) makes exactly one attempt to prune the Temporal-side version record, and a failed attempt orphans that record permanently.

`executePlan` runs `executeK8sOperations` first, which deletes the k8s Deployment, then calls `deleteDrainedVersions`:

```go
_, err := depHandle.DeleteVersion(ctx, sdkclient.WorkerDeploymentDeleteVersionOptions{...})
if err != nil {
    if errors.As(err, &notFound) {
        continue
    }
    l.Info("could not delete worker deployment version, may require manual cleanup", "buildID", buildID, "error", err)
}
```

The failure is logged at `info`, the function returns nothing, and no error propagates, so there is no requeue. On the next reconcile the Deployment no longer exists, so the version is absent from `k8sState.Deployments` and can never become a `DeleteDeployments` candidate again. `getDeleteDeployments` only considers versions whose Deployment is still present, and the comment on `EligibleForDeletion` says as much. From that point the record is unreachable by the controller, short of deleting the whole WorkerDeployment.

The message even says "may require manual cleanup", but nothing surfaces it: no event, no condition, no metric. The only way to notice is grepping controller logs at `info` level.

## Two failure classes observed in production

Both on v1.9.0, across two clusters:

1. **`cannot be deleted since it has active pollers`.** Pollers linger server-side for `matching.PollerHistoryTTL` (5m default) after pods stop. Any `deleteDelay` shorter than the TTL guarantees this: at `deleteDelay: 1m` we measured **50 failures out of 97 prunes (52%)**. Raising `deleteDelay` to 8m removed the deterministic case, dropping two clusters to 7% and 13%, but did not eliminate it. I have not been able to attribute the residual failures to a mechanism; see #439, where the obvious suspect (autoscaler flap restarting pods during the window) did not survive checking.

2. **`stream terminated by RST_STREAM with error code: CANCEL`.** A transport error on the `DeleteVersion` call, observed during retirement bursts when several versions are pruned within seconds of each other. Unrelated to autoscaling, and no amount of delay tuning prevents it.

The residual failures concentrate in bursts, where a dozen or more versions sunset within a couple of minutes: 3 of 13 in one, 3 of 17 in another.

The default `deleteDelay` of 24h hides class 1 from most users, since the poller cache has long since expired. Anyone who shortens `deleteDelay`, which is reasonable for tighter retirement, hits it immediately. Class 2 can hit anyone.

## Impact

Orphaned records accumulate against `matching.maxVersionsInDeployment` and `matching.maxVersionsInTaskQueue`. In our environment the pre-existing backlog is roughly 400 per cluster, all with no k8s Deployment and therefore unreachable by the controller. Clearing them needs an out-of-band sweep calling `DeleteVersion` directly.

## Suggested fixes

Any one of these would close it:

1. **Return an error from `deleteDrainedVersions` so the reconcile requeues.** Pollers age out within `PollerHistoryTTL` and transport errors are transient, so a later attempt succeeds. The WorkerDeployment-deletion path in the same file already does exactly this, with `return fmt.Errorf("unable to delete version %s (will retry): %w", ...)` and a comment noting that pollers linger for `PollerHistoryTTL` "so a later attempt succeeds". The sunset path is inconsistent with it.
2. **Attempt `DeleteVersion` before deleting the Deployment**, so a failure leaves the Deployment in place and the version stays a candidate on the next reconcile.
3. **Record the failure in status** (for example a `pendingVersionDeletion` list) so a later reconcile can retry without needing the Deployment as its handle.

Option 1 looks smallest and matches existing behaviour elsewhere in the codebase.

## Observability gap

Independently of the fix, a failed prune deserves more visibility than an `info` log: an event on the WorkerDeployment, a status condition, or a counter metric. As it stands a cluster can leak version records for weeks with nothing to indicate it.

tomba7:
Fixes #543

## What was changed
- deleteDrainedVersions now runs before executeK8sOperations and returns the subset of plan.DeleteDeployments whose record the server had just confirmed gone.
- executePlan narrows the plan to that subset, so a version's k8s Deployment is only pruned once the version itself is deleted from the server side. 

## Why?
- The controller only tried to delete the version on the server once, and in that same pass it deleted the k8s Deployment, which was the only thing that would have let it try again
- A version appears in status.DeprecatedVersions only while its k8s Deployment exists, and getDeleteDeployments only picks versions that still have one. 
- So once the k8s Deployment was gone, the version on the server was invisible to the controller going forward
- Any transient failure orphaned the record permanently and is logged at info with no event or condition.

## Checklist

1. How was this tested:
   `go test ./internal/controller/...`
   * TestExecutePlan_VersionDeletionFails_KeepsDeploymentAndRetriesNextCycle
	   * The old ordering fails this test
   * TestExecutePlan_VersionAlreadyDeletedOnServer_DeletesDeployment

   Full CI equivalent run locally, all green i.e. fmt-imports, tidy, lint-code, go vet,
   lint-actions, test-unit, test-integration

2. Any docs updates needed?
   No


internal/controller/execplan.go
```diff
@@ -653,14 +653,21 @@ func (r *WorkerDeploymentReconciler) executeWRTOperations(
 func (r *WorkerDeploymentReconciler) deleteDrainedVersions(
 	ctx context.Context,
 	l logr.Logger,
+	workerDeploy *temporaliov1alpha1.WorkerDeployment,
 	depHandle sdkclient.WorkerDeploymentHandle,
 	p *plan,
-) {
+) []*appsv1.Deployment {
```
jaypipes: Please update the docstring for this method to indicate what is being returned now.

Alternately, instead of returning a slice of `*appsv1.Deployment`s, you could simply mutate `p.DeleteDeployments` since it seems that's what the call to `deleteDrainedVersions()` is doing in this patch?

```go
    p.DeleteDeployments = r.deleteDrainedVersions(ctx, l, workerDeploy, deploymentHandler, p)
```

If you do remove the returned slice, make sure to update the docstring to say that `p` is mutated to remove Kubernetes Deployments that should not be deleted due to the Temporal server-side WDV record not being found or not being deleted.

internal/controller/execplan.go
```diff
@@ -653,14 +653,21 @@ func (r *WorkerDeploymentReconciler) executeWRTOperations(
...
 	identity := getControllerIdentity()
+	markedForDeletion := make([]*appsv1.Deployment, 0, len(p.DeleteDeployments))
 	for _, d := range p.DeleteDeployments {
 		buildID, ok := d.GetLabels()[k8s.BuildIDLabel]
 		if !ok {
 			l.Info("deployment has no build ID label, skipping Temporal server-side version cleanup", "deployment", d.Name)
+			markedForDeletion = append(markedForDeletion, d)
```
jaypipes: If the Kubernetes Deployment has no buildID label, I don't think we want to mark it for deletion. I mean, the only way this code path would even execute is if the user has manually removed the buildID label from a Kubernetes Deployment, but if they have done so for some reason, we should probably be safe and assume that they are manually cleaning things up.

internal/controller/execplan.go
```diff
@@ -653,14 +653,21 @@ func (r *WorkerDeploymentReconciler) executeWRTOperations(
...
+			continue
+		}
+		if isVersionNotRegistered(workerDeploy, buildID) {
+			markedForDeletion = append(markedForDeletion, d)
 			continue
 		}
```
jaypipes, line 712: Note that as of @Shivs11's recent [commit](https://github.com/temporalio/temporal-worker-controller/commit/6a222eef10064aad5973c86ce08457676c95e136), we are [returning an error if the server returns NotRegistered status for a particular buildID](https://github.com/temporalio/temporal-worker-controller/commit/6a222eef10064aad5973c86ce08457676c95e136#diff-7dc7e72c757762856373da9b184293940b8ed6b895757b655c6da538dfeb2e78R185). So, I'm not sure whether this block of code should exist... @Shivs11 what do you think?

jaypipes (changes requested):
Thanks @tomba7! A few things inline to respond to :)

internal/controller/execplan.go
tomba7: Updated both the docstring and the signature accordingly

internal/controller/execplan.go
tomba7: Makes sense. I've described the scenario with a comment right above the logger.

internal/controller/execplan.go
tomba7: Since getDeleteDeployments still nominates NotRegistered versions for deletion, and they still reach this loop, it seems like it makes sense to keep this check. But, I haven't read through #556's implementation closely, so I'd
want your input to confirm whether it can be safely removed here. @jaypipes @Shivs11

jaypipes (approved):
This is good as-is. The NotRegistered block probably won't ever be executed because of Shivam's recent commit, but it's no biggie. :)

tomba7:
Thank you getting this merged @jaypipes Much appreciated

PR #554, opened 2026-08-28
Make controller more resilient to inconsistent Temporal state, delay requeue after Temporal ResourceExhausted, improve logs

### LINKED ISSUE #555: [bug] work around inconsistent worker deployment system workflow state
When gathering the latest observed state of a Temporal Worker Deployment, we [attempt to determine](https://github.com/temporalio/temporal-worker-controller/blob/252364f6e7c473d79fbdb2a094cf04e579912ac9/internal/temporal/worker_deployment.go#L159-L209) the Worker Deployment Version's status by examining the `WorkerDeploymentVersionSummary` struct returned from a call to `DescribeWorkerDeployment` API. We look at the `RoutingConfig` struct associated with the WorkerDeployment as well as the `DrainageInfo` field of the `WorkerDeploymentVersionSummary` struct to determine the resulting status of the WDV:

 ```go
	// Determine summary status
	drainageStatus := summary.DrainageInfo.GetStatus()
	if routingConfig.CurrentDeploymentVersion != nil &&
		summary.DeploymentVersion.DeploymentName == routingConfig.CurrentDeploymentVersion.DeploymentName &&
		summary.DeploymentVersion.BuildId == routingConfig.CurrentDeploymentVersion.BuildId {
		out.Status = temporaliov1alpha1.VersionStatusCurrent
	} else if routingConfig.RampingDeploymentVersion != nil &&
		summary.DeploymentVersion.DeploymentName == routingConfig.RampingDeploymentVersion.DeploymentName &&
		summary.DeploymentVersion.BuildId == routingConfig.RampingDeploymentVersion.BuildId {
		out.Status = temporaliov1alpha1.VersionStatusRamping
	} else if drainageStatus == enumspb.VERSION_DRAINAGE_STATUS_DRAINING {
		out.Status = temporaliov1alpha1.VersionStatusDraining
	} else if drainageStatus == enumspb.VERSION_DRAINAGE_STATUS_DRAINED {
		out.Status = temporaliov1alpha1.VersionStatusDrained

		// Extract DrainedSince directly from the summary's drainage info,
		// avoiding a per-summary DescribeVersion call.
		if summary.DrainageInfo != nil && summary.DrainageInfo.LastChangedTime != nil {
			drainedSince := summary.DrainageInfo.LastChangedTime.AsTime()
			out.DrainedSince = &drainedSince
		}
	} else {
		out.Status = temporaliov1alpha1.VersionStatusInactive
		// get unversioned poller info to decide whether to fast-track rollout
		if summary.DeploymentVersion.BuildId == targetBuildID &&
			routingConfig.CurrentDeploymentVersion == nil &&
			strategy == temporaliov1alpha1.UpdateProgressive {
			var err error
			var desc temporalClient.WorkerDeploymentVersionDescription
			describeVersion := func() error {
				desc, err = depHandle.DescribeVersion(
					ctx,
					temporalClient.WorkerDeploymentDescribeVersionOptions{
						BuildID: summary.DeploymentVersion.BuildId,
					},
				)
				return err
			}
			// At first, summary is found in DeploymentInfo.VersionSummaries but not ready for describe, so we have
			// to describe with backoff.
			//
			// Note: We can only check whether the task queues that we know of have unversioned pollers.
			//       If, later on, a poll request arrives tying a new task queue to the target summary, we
			//       don't know whether that task queue has unversioned pollers.
			if err = withBackoff(10*time.Second, 1*time.Second, describeVersion); err == nil { //revive:disable-line:max-control-nesting
				out.AllTaskQueuesHaveUnversionedPoller = allTaskQueuesHaveUnversionedPoller(ctx, client, desc.Info.TaskQueuesInfos)
			}
			// NOTE(jaypipes): We swallow any non-nil error here. Should we
			// at least log the error?
		}
	}
```

Unfortunately, there is a race condition in Temporal OSS server that introduces a small time window where the `RoutingConfig` field's contents may be inconsistent with the `VersionInfoSummaries` field contents. This race condition is exacerbated (the window of time where the inconsistency can occur is lengthened) in environments containing many thousands of pollers.

This race condition results in `RoutingConfig.CurrentBuildId` containing the correct current version for the Worker Deployment but the `VersionInfoSummary` for a demoted version contains the wrong `DrainageInfo` `Status` value.

This causes the following fallthrough condition to be executed:

```go
	} else {
		out.Status = temporaliov1alpha1.VersionStatusInactive
...
	}
```

Which incorrectly sets the status of the version to Inactive, which in turn erroneously [causes TWC to set the Kubernetes Deployment's `spec.replicas` field to 0](https://github.com/temporalio/temporal-worker-controller/blob/252364f6e7c473d79fbdb2a094cf04e579912ac9/internal/planner/planner.go#L847), which causes the premature deletion of the WDV, which in turn ends up stranding any open workflows that workers in that WDV were processing.

The fix is to be aware of the potential inconsistency in the returned response from `DescribeWorkerDeployment` and work around this bug in the server.

carlydf:
## What was changed
- Make controller more resilient to inconsistent Temporal state (actually check version status instead of just deducing it from routing config)
- delay requeue after Temporal ResourceExhausted
- improve "scaling" log by including build id tag

## Why?
- Server race condition where async SetCurrent could still be in progress when Describe request is served so routing config is out of sync with version summaries.
- no need to requeue so fast
- want to know what is being scaled

Closes #555

### DIFF
--- a/api/v1alpha1/workerdeployment_types.go
+++ b/api/v1alpha1/workerdeployment_types.go
@@ -151,8 +151,14 @@ type VersionStatus string
 
 const (
 	// VersionStatusNotRegistered indicates that the version is not registered
-	// with Temporal for any worker deployment.
+	// with Temporal for any worker deployment. It may be not yet created, or
+	// may have been deleted.
 	VersionStatusNotRegistered VersionStatus = "NotRegistered"
+
+	// VersionStatusCreated indicates that the version has been created explicitly.
+	// When the first poll requests for this version come in, the version status
+	// will change to Inactive.
+	VersionStatusCreated VersionStatus = "Created"
 
 	// VersionStatusInactive indicates that the version is registered in a Temporal
 	// worker deployment, but has not been set to current or ramping.
--- a/internal/controller/execplan.go
+++ b/internal/controller/execplan.go
@@ -99,8 +99,8 @@ func (r *WorkerDeploymentReconciler) executeK8sOperations(ctx context.Context, l
 	for d, replicas := range p.ScaleDeployments {
 		buildID := buildIDForDeployment(workerDeploy, d)
 		l.Info(
-			fmt.Sprintf("scaling deployment %q for Build ID %q to %d replicas", d.Name, buildID, replicas),
-			"deployment", d.Name,
+			"scaling deployment",
+			"deployment", d,
 			"buildID", buildID,
 			"replicas", replicas,
 		)
@@ -115,8 +115,8 @@ func (r *WorkerDeploymentReconciler) executeK8sOperations(ctx context.Context, l
 		if err := r.Client.SubResource("scale").Update(ctx, dep, client.WithSubResourceBody(scale)); err != nil {
 			l.Error(
 				err,
-				fmt.Sprintf("unable to scale deployment %q for Build ID %q to %d replicas", d.Name, buildID, replicas),
-				"deployment", d.Name,
+				"unable to scale deployment",
+				"deployment", d,
 				"buildID", buildID,
 				"replicas", replicas,
 			)
--- a/internal/controller/execplan.go
+++ b/internal/controller/execplan.go
@@ -143,6 +143,7 @@ func (r *WorkerDeploymentReconciler) executeK8sOperations(ctx context.Context, l
 func buildIDForDeployment(workerDeploy *temporaliov1alpha1.WorkerDeployment, deployment *corev1.ObjectReference) string {
 	matches := func(versionDeployment *corev1.ObjectReference) bool {
 		return versionDeployment != nil &&
+			deployment != nil &&
 			versionDeployment.Namespace == deployment.Namespace &&
 			versionDeployment.Name == deployment.Name
 	}
--- a/internal/controller/execplan.go
+++ b/internal/controller/execplan.go
@@ -97,7 +97,13 @@ func (r *WorkerDeploymentReconciler) executeK8sOperations(ctx context.Context, l
 
 	// Scale deployments
 	for d, replicas := range p.ScaleDeployments {
-		l.Info("scaling deployment", "deployment", d, "replicas", replicas)
+		buildID := buildIDForDeployment(workerDeploy, d)
+		l.Info(
+			fmt.Sprintf("scaling deployment %q for Build ID %q to %d replicas", d.Name, buildID, replicas),
+			"deployment", d.Name,
+			"buildID", buildID,
+			"replicas", replicas,
+		)
 		dep := &appsv1.Deployment{ObjectMeta: metav1.ObjectMeta{
 			Namespace:       d.Namespace,
 			Name:            d.Name,
@@ -107,9 +113,15 @@ func (r *WorkerDeploymentReconciler) executeK8sOperations(ctx context.Context, l
 
 		scale := &autoscalingv1.Scale{Spec: autoscalingv1.ScaleSpec{Replicas: int32(replicas)}}
 		if err := r.Client.SubResource("scale").Update(ctx, dep, client.WithSubResourceBody(scale)); err != nil {
-			l.Error(err, "unable to scale deployment", "deployment", d, "replicas", replicas)
+			l.Error(
+				err,
+				fmt.Sprintf("unable to scale deployment %q for Build ID %q to %d replicas", d.Name, buildID, replicas),
+				"deployment", d.Name,
+				"buildID", buildID,
+				"replicas", replicas,
+			)
 			r.Recorder.Eventf(workerDeploy, corev1.EventTypeWarning, ReasonDeploymentScaleFailed,
-				"Failed to scale Deployment %q to %d replicas: %v", d.Name, replicas, err)
+				"Failed to scale Deployment %q for Build ID %q to %d replicas: %v", d.Name, buildID, replicas, err)
 			return deletedWorkerResources, fmt.Errorf("unable to scale deployment: %w", err)
 		}
 	}
@@ -126,6 +138,27 @@ func (r *WorkerDeploymentReconciler) executeK8sOperations(ctx context.Context, l
 	}
 
 	return deletedWorkerResources, nil
+}
+
+func buildIDForDeployment(workerDeploy *temporaliov1alpha1.WorkerDeployment, deployment *corev1.ObjectReference) string {
+	matches := func(versionDeployment *corev1.ObjectReference) bool {
+		return versionDeployment != nil &&
+			versionDeployment.Namespace == deployment.Namespace &&
+			versionDeployment.Name == deployment.Name
+	}
+
+	if matches(workerDeploy.Status.TargetVersion.Deployment) {
+		return workerDeploy.Status.TargetVersion.BuildID
+	}
+	if workerDeploy.Status.CurrentVersion != nil && matches(workerDeploy.Status.CurrentVersion.Deployment) {
+		return workerDeploy.Status.CurrentVersion.BuildID
+	}
+	for _, version := range workerDeploy.Status.DeprecatedVersions {
+		if version != nil && matches(version.Deployment) {
+			return version.BuildID
+		}
+	}
+	return "unknown"
 }
 
 func (r *WorkerDeploymentReconciler) startTestWorkflows(ctx context.Context, l logr.Logger, workerDeploy *temporaliov1alpha1.WorkerDeployment, temporalClient sdkclient.Client, p *plan) error {
--- a/internal/controller/worker_controller.go
+++ b/internal/controller/worker_controller.go
@@ -323,8 +323,8 @@ func (r *WorkerDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		if errors.As(err, &rateLimitErr) {
 			r.recordWarningAndSetBlocked(ctx, &workerDeploy,
 				temporaliov1alpha1.ReasonTemporalStateFetchFailed,
-				fmt.Sprintf("Rate limited fetching worker deployment state: %v", err),
-				fmt.Sprintf("Rate limited by Temporal server: %v", err))
+				fmt.Sprintf("Got ResourceExhausted error fetching worker deployment state from Temporal server: %v", err),
+				fmt.Sprintf("Got ResourceExhausted error fetching worker deployment state from Temporal server: %v", err))
 			return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
 		}
 		r.recordWarningAndSetBlocked(ctx, &workerDeploy,
@@ -364,6 +364,14 @@ func (r *WorkerDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	if err := r.executePlan(ctx, l, &workerDeploy, temporalClient, plan); err != nil {
 		if isAccessDeniedErr(err) {
 			r.TemporalClientPool.EvictClient(clientPoolKey)
+		}
+		var rateLimitErr *serviceerror.ResourceExhausted
+		if errors.As(err, &rateLimitErr) {
+			r.recordWarningAndSetBlocked(ctx, &workerDeploy,
+				ReasonPlanExecutionFailed,
+				fmt.Sprintf("Got ResourceExhausted error executing plan: %v", err),
+				fmt.Sprintf("Got ResourceExhausted error executing plan: %v", err))
+			return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
 		}
 		r.recordWarningAndSetBlocked(ctx, &workerDeploy,
 			ReasonPlanExecutionFailed,
--- a/internal/controller/worker_controller.go
+++ b/internal/controller/worker_controller.go
@@ -307,6 +307,7 @@ func (r *WorkerDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	// Fetch Temporal worker deployment state
 	temporalState, err := temporal.GetWorkerDeploymentState(
 		ctx,
+		l,
 		temporalClient,
 		workerDeploymentName,
 		workerDeploy.Spec.WorkerOptions.TemporalNamespace,
--- a/internal/temporal/worker_deployment.go
+++ b/internal/temporal/worker_deployment.go
@@ -166,6 +166,7 @@ func versionInfoFromVersionSummary(
 	case enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_CREATED:
 		out.Status = temporaliov1alpha1.VersionStatusCreated
 	case enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_INACTIVE:
+		out.Status = temporaliov1alpha1.VersionStatusInactive
 		// get unversioned poller info to decide whether to fast-track rollout
 		if summary.DeploymentVersion.BuildId == targetBuildID &&
 			routingConfig.CurrentDeploymentVersion == nil &&
--- a/internal/temporal/worker_deployment.go
+++ b/internal/temporal/worker_deployment.go
@@ -140,6 +140,17 @@ func GetWorkerDeploymentState(
 	return state, nil
 }
 
+// versionStatusMap translates between the temporal SDK VersionStatus and the
+// TWC k8s API VersionStatus
+var versionStatusMap = map[enumspb.WorkerDeploymentVersionStatus]temporaliov1alpha1.VersionStatus{
+	enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_CURRENT:     temporaliov1alpha1.VersionStatusCurrent,
+	enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_DRAINING:    temporaliov1alpha1.VersionStatusDraining,
+	enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_DRAINED:     temporaliov1alpha1.VersionStatusDrained,
+	enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_INACTIVE:    temporaliov1alpha1.VersionStatusInactive,
+	enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_RAMPING:     temporaliov1alpha1.VersionStatusRamping,
+	enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_UNSPECIFIED: temporaliov1alpha1.VersionStatusNotRegistered,
+}
+
 // versionInfoFromVersionSummary returns a VersionInfo constructed from the
 // supplied Temporal VersionSummary message.
 //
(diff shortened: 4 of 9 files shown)


jaypipes:
@carlydf I have created issue #555 for tracking this, marked it urgent and targeted it for Tuesday's release. Please see the integration test failures. There's something majorly borked and seems to be a problem with the server, not TWC.

internal/temporal/worker_deployment.go
```diff
@@ -206,6 +193,42 @@ func versionInfoFromVersionSummary(
 			// NOTE(jaypipes): We swallow any non-nil error here. Should we
 			// at least log the error?
 		}
+	case enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_RAMPING:
```
Shivs11: nit: could be awesome to just have a small comment here explaining that we are comparing each version summary here with the routingConfig's view, since there could be inconsistent state.

internal/temporal/worker_deployment.go
```diff
@@ -157,28 +160,12 @@ func versionInfoFromVersionSummary(
...
-			out.DrainedSince = &drainedSince
-		}
-	} else {
-		out.Status = temporaliov1alpha1.VersionStatusInactive
+	switch summary.GetStatus() {
+	case enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_UNSPECIFIED:
+		out.Status = temporaliov1alpha1.VersionStatusNotRegistered
+	case enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_CREATED:
+		out.Status = temporaliov1alpha1.VersionStatusCreated
+	case enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_INACTIVE:
```
Shivs11: I think this is the reason we could be having integration test failures on this PR -- we do not seem to be having the inactive version mapped out to the "out.status" if the version is truly inactive, no?

Shivs11: I got this because I am seeing: " expected target version status to be 'Inactive', got ''" error strings in the integration test suite

internal/temporal/worker_deployment.go
```diff
@@ -206,6 +193,42 @@ func versionInfoFromVersionSummary(
...
+		if routingConfig.RampingDeploymentVersion != nil &&
+			summary.DeploymentVersion.DeploymentName == routingConfig.RampingDeploymentVersion.DeploymentName &&
+			summary.DeploymentVersion.BuildId == routingConfig.RampingDeploymentVersion.BuildId {
+			out.Status = temporaliov1alpha1.VersionStatusRamping
+		} else {
+			l.Error(
+				errors.New("worker deployment version status conflicts with routing config"),
+				"version reports Ramping but routing config identifies a different Ramping version; trusting routing config",
+				"buildID", summary.DeploymentVersion.BuildId,
+				"routingConfigBuildID", routingConfig.RampingDeploymentVersion.BuildId,
```
Shivs11: just thinking this through out loud: so say we land in an inconsistent state where you have a secondary version being the "stale current" version. In that world, we would come here, log the error message. 

However, for this version, we would not assign a specific status then, no? I wonder if we should return an error in this reconciliation loop rather than worry about "what status this version should be" from the controller pov, since tbh the server is giving it weird state

internal/controller/execplan.go
```diff
@@ -128,6 +140,27 @@ func (r *WorkerDeploymentReconciler) executeK8sOperations(ctx context.Context, l
 	return deletedWorkerResources, nil
 }
 
+func buildIDForDeployment(workerDeploy *temporaliov1alpha1.WorkerDeployment, deployment *corev1.ObjectReference) string {
+	matches := func(versionDeployment *corev1.ObjectReference) bool {
+		return versionDeployment != nil &&
```
Shivs11, line 145: sorry to be pedantic here, i don't mean to but we are never ever gonna have versionDeployment.Namespace/Name (and the same for deployment.xxx) ever be nil, yeah?

Shivs11 (commented):
a few comments

api/v1alpha1/workerdeployment_types.go
```diff
@@ -151,9 +151,15 @@ type VersionStatus string
 
 const (
 	// VersionStatusNotRegistered indicates that the version is not registered
-	// with Temporal for any worker deployment.
+	// with Temporal for any worker deployment. It may be not yet created, or
+	// may have been deleted.
 	VersionStatusNotRegistered VersionStatus = "NotRegistered"
 
+	// VersionStatusCreated indicates that the version has been created explicitly.
```
jaypipes, line 158: "created explicitly" might be taken to mean "created by calling `temporal worker deployment create-version`", just FYI :)

api/v1alpha1/workerdeployment_types.go
```diff
@@ -151,9 +151,15 @@ type VersionStatus string
...
+	// When the first poll requests for this version come in, the version status
+	// will change to Inactive.
```
jaypipes, line 160: Why is this?

internal/controller/execplan.go
```diff
@@ -97,7 +97,13 @@ func (r *WorkerDeploymentReconciler) executeK8sOperations(ctx context.Context, l
 
 	// Scale deployments
 	for d, replicas := range p.ScaleDeployments {
-		l.Info("scaling deployment", "deployment", d, "replicas", replicas)
+		buildID := buildIDForDeployment(workerDeploy, d)
+		l.Info(
+			fmt.Sprintf("scaling deployment %q for Build ID %q to %d replicas", d.Name, buildID, replicas),
```
jaypipes: I know you want to add this information to the log line to assist in debugging, but this isn't appropriate, IMHO. We should continue to use the best practice of using the log key/value pairs (but add the `buildID` key/value pair as you've done below)

internal/controller/execplan.go
jaypipes: Both [Namespace and Name are strings](https://github.com/kubernetes/api/blob/4ec81822ffec45f9987712a2bb9a701d8ff9e1c0/core/v1/types.go#L7805-L7812), not pointers to string. So no, that's not possible.

That said, `deployment` can be nil and probably should have a nil guard here.

internal/controller/worker_controller.go
```diff
@@ -365,6 +366,14 @@ func (r *WorkerDeploymentReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		if isAccessDeniedErr(err) {
 			r.TemporalClientPool.EvictClient(clientPoolKey)
 		}
+		var rateLimitErr *serviceerror.ResourceExhausted
+		if errors.As(err, &rateLimitErr) {
+			r.recordWarningAndSetBlocked(ctx, &workerDeploy,
```
jaypipes, line 371: Why do we want to set the worker deployment to blocked on a resource exhausted error executing the plan?

internal/controller/worker_controller.go
jaypipes, line 356: Do we want to check for resource exhausted here as well?

internal/temporal/worker_deployment.go
```diff
@@ -206,6 +193,42 @@ func versionInfoFromVersionSummary(
 			// NOTE(jaypipes): We swallow any non-nil error here. Should we
 			// at least log the error?
 		}
+	case enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_RAMPING:
+		if routingConfig.RampingDeploymentVersion != nil &&
+			summary.DeploymentVersion.DeploymentName == routingConfig.RampingDeploymentVersion.DeploymentName &&
+			summary.DeploymentVersion.BuildId == routingConfig.RampingDeploymentVersion.BuildId {
+			out.Status = temporaliov1alpha1.VersionStatusRamping
+		} else {
+			l.Error(
```
jaypipes: This is not an error in TWC, though, so I think this should be, at most, a warning. Errors should be for when TWC is not behaving as expected. But, in this case, it's temporal OSS server that has returned inconsistent data. We should just log an info or warning record but make sure to set `out.Status` to `VersionStatusRamping` otherwise we will be returning a VersionInfo struct with an empty `Status` value.

I think this switch code block should be rewritten to something more like this:

```go
// versionStatusMap translates between the temporal SDK VersionStatus and the
// TWC k8s API VersionStatus
var versionStatusMap map[enumspb.WorkerDeploymentVersionStatus]temporaliov1alpha1.VersionStatus{
    enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_CURRENT: temporaliov1alpha1.VersionStatusCurrent,
    enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_DRAINING: temporaliov1alpha1.VersionStatusDraining,
    enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_DRAINED: temporaliov1alpha1.VersionStatusDrained,
    enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_INACTIVE: temporaliov1alpha1.VersionStatusInactive,
    enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_RAMPING: temporaliov1alpha1.VersionStatusRamping,
    enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_UNSPECIFIED: temporaliov1alpha1.VersionStatusNotRegistered,
}
...
func versionInfoFromVersionSummary(
...
    apiVersionStatus, ok := versionStatus[summary.GetStatus()]
    if !ok {
        l.Error(fmt.Errorf("unknown worker version status %s", summary.GetStatus())
        return nil
    }
    out.Status = apiVersionStatus

    sumDeployName := summary.DeploymentVersion.DeploymentName
    sumBuildID := summary.DeploymentVersion.BuildId

    // There may be inconsistencies in the data returned from temporal server. Check
    // for these inconsistencies and warn if found.
    if apiVersionStatus == temporaliov1alpha1.VersionStatusCurrent {
        if routingConfig.CurrentDeploymentVersion != nil {
            rcDeployName := routingConfig.CurrentDeploymentVersion.DeploymentName
            rcBuildID := routingConfig.CurrentDeploymentVersion.BuildId
            if sumDeploymentName != rcDeployName || sumBuildID != rcBuildID {
                l.Warn(
				"version reports Current but routing config identifies a different Current version; trusting routing config",
				"buildID", sumBuildID,
				"routingConfigBuildID", rcBuildId,
			)
		}
    } else if apiVersionStatus == temporaliov1alpha1.VersionStatusRamping {
        if routingConfig.RampingDeploymentVersion != nil {
            rcDeployName := routingConfig.RampingDeploymentVersion.DeploymentName
            rcBuildID := routingConfig.RampingDeploymentVersion.BuildId
            if sumDeploymentName != rcDeployName || sumBuildID != rcBuildID {
                l.Warn(
				"version reports Ramping but routing config identifies a different Ramping version; trusting routing config",
				"buildID", sumBuildID,
				"routingConfigBuildID", rcBuildId,
			)
		}
    }
...
```

api/v1alpha1/workerdeployment_types.go
carlydf: that is what this status means!

api/v1alpha1/workerdeployment_types.go
carlydf: This matches exactly what the "created" status is in the UI.

I guess... how should TWC handle a version that it sees is "Created" but no pollers yet. I think that it should still spin up a k8s Deployment, but we could also block and say "this is likely supposed to be managed by serverless and not TWC." In the previous code, we would have treated "created" the same as "inactive" because "inactive" was the "else" condition


Statuses from API proto:
```
const (
	WORKER_DEPLOYMENT_VERSION_STATUS_UNSPECIFIED WorkerDeploymentVersionStatus = 0
	// The Worker Deployment Version has been created inside the Worker Deployment but is not used by any
	// workflow executions. These Versions can still have workflows if they have an explicit Versioning Override targeting
	// this Version. Such Versioning Override could be set at workflow start time, or at a later time via `UpdateWorkflowExecutionOptions`.
	WORKER_DEPLOYMENT_VERSION_STATUS_INACTIVE WorkerDeploymentVersionStatus = 1
	// The Worker Deployment Version is the current version of the Worker Deployment. All new workflow executions
	// and tasks of existing unversioned or AutoUpgrade workflows are routed to this version.
	WORKER_DEPLOYMENT_VERSION_STATUS_CURRENT WorkerDeploymentVersionStatus = 2
	// The Worker Deployment Version is the ramping version of the Worker Deployment. A subset of new Pinned workflow executions are
	// routed to this version. Moreover, a portion of existing unversioned or AutoUpgrade workflow executions are also routed to this version.
	WORKER_DEPLOYMENT_VERSION_STATUS_RAMPING WorkerDeploymentVersionStatus = 3
	// The Worker Deployment Version is not used by new workflows but is still used by
	// open pinned workflows. The version cannot be decommissioned safely.
	WORKER_DEPLOYMENT_VERSION_STATUS_DRAINING WorkerDeploymentVersionStatus = 4
	// The Worker Deployment Version is not used by new or open workflows, but might be still needed by
	// Queries sent to closed workflows. The version can be decommissioned safely if user does
	// not query closed workflows. If the user does query closed workflows for some time x after
	// workflows are closed, they should decommission the version after it has been drained for that duration.
	WORKER_DEPLOYMENT_VERSION_STATUS_DRAINED WorkerDeploymentVersionStatus = 5
	// The Worker Deployment Version is created by user (via `CreateWorkerDeploymentVersion` API)
	// but server has not seen any poller for it yet.
	WORKER_DEPLOYMENT_VERSION_STATUS_CREATED WorkerDeploymentVersionStatus = 6
)
```

internal/controller/execplan.go
carlydf: ok yeah that occurred to me too, I just wasn't sure where to land on the tradeoff.

are you saying if a user strips the key/value pair when they ingest the logs, that is simply on them?

internal/controller/execplan.go
carlydf: added `deployment != nil` guard!

internal/controller/worker_controller.go
carlydf: `generatePlan` calls `k8sclient.List`, but does not call any temporal APIs, so we don't get it from there. it would not hurt to add it in case some future change makes genplan call temporal? Architecturally i'd prefer if we maintain forever that genplan does not call temporal though, fwiw.

I'm happy to add the resource exhausted handle if you think that's good insurance though

internal/controller/worker_controller.go
carlydf: hm let's think: "set blocked" means we do:

```
	r.setCondition(workerDeploy, temporaliov1alpha1.ConditionProgressing, metav1.ConditionFalse, reason, conditionMessage)
	r.setCondition(workerDeploy, temporaliov1alpha1.ConditionReady, metav1.ConditionFalse, reason, conditionMessage)
```

To me it makes sense to say "Progressing = False", because we are waiting 30s for resource exhausted to hopefully go away. Do you think that makes sense?

Tbh, the best user experience here would be for us to look at the exact error, and if it talks about a task queue limit, or number of versions limit, or number of deployments limit, we could elaborate more on what action the user should take for each one. 

But in the meantime, I think changing the text from "rate limited" (sometimes not true) to "resource exhausted" (the umbrella error type server is currently using for both rate limited and size limited) is an improvement since at least it doesnt misrepresent

internal/temporal/worker_deployment.go
carlydf: oops :) done. nice our tests work :)

api/v1alpha1/workerdeployment_types.go
jaypipes: ack, ok that makes sense.

internal/controller/execplan.go
jaypipes: > are you saying if a user strips the key/value pair when they ingest the logs, that is simply on them?

Yes, that is exactly what I'm saying :)

internal/controller/worker_controller.go
jaypipes: nope, all good. your explanation makes sense to me.

internal/controller/worker_controller.go
jaypipes: > hm let's think: "set blocked" means we do:
> 
> ```
> 	r.setCondition(workerDeploy, temporaliov1alpha1.ConditionProgressing, metav1.ConditionFalse, reason, conditionMessage)
> 	r.setCondition(workerDeploy, temporaliov1alpha1.ConditionReady, metav1.ConditionFalse, reason, conditionMessage)
> ```
> 
> To me it makes sense to say "Progressing = False", because we are waiting 30s for resource exhausted to hopefully go away. Do you think that makes sense?

Yes.

> But in the meantime, I think changing the text from "rate limited" (sometimes not true) to "resource exhausted" (the umbrella error type server is currently using for both rate limited and size limited) is an improvement since at least it doesnt misrepresent

Agreed.

I guess I saw "SetBlocked" and assumed that set the WD into some sort of terminal state... but that's not the case, it just updates those two conditions, as you note above. OK, never mind then!

internal/temporal/worker_deployment.go
carlydf: I just pushed a refactor to do this instead of the switch

internal/temporal/worker_deployment.go
carlydf: I think Jay's refactor suggestion fixes this issue

internal/temporal/worker_deployment.go
carlydf: added 
```
	// There may be inconsistencies in the data returned from temporal server. Check
	// for these inconsistencies and warn if found.
```

internal/controller/execplan.go
carlydf: pushed the removal of those changes!

internal/temporal/worker_deployment.go
```diff
@@ -138,12 +140,24 @@ func GetWorkerDeploymentState(
...
 
+// versionStatusMap translates between the temporal SDK VersionStatus and the
+// TWC k8s API VersionStatus
+var versionStatusMap = map[enumspb.WorkerDeploymentVersionStatus]temporaliov1alpha1.VersionStatus{
+	enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_CURRENT:     temporaliov1alpha1.VersionStatusCurrent,
+	enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_DRAINING:    temporaliov1alpha1.VersionStatusDraining,
+	enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_DRAINED:     temporaliov1alpha1.VersionStatusDrained,
+	enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_INACTIVE:    temporaliov1alpha1.VersionStatusInactive,
+	enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_RAMPING:     temporaliov1alpha1.VersionStatusRamping,
+	enumspb.WORKER_DEPLOYMENT_VERSION_STATUS_UNSPECIFIED: temporaliov1alpha1.VersionStatusNotRegistered,
```
jaypipes, line 152: I didn't have the CREATED status in there. I guess for completeness we need to add it?

carlydf: added

jaypipes (approved):
thank you @carlydf !

PR #556, opened 2026-08-28
fix(controller): DescribeVersion on NotRegistered Versions to prevent pre-mature scaledown of said versions 

### LINKED ISSUE #550: [Bug] scale down K8s deployments if the version is NotRegistered + NotFound
- I think we should only scale the k8s deployments down of NotRegistered versions if they truly have no workloads being processed on them.

### Describe the bug
- Today, the worker controller gets ready to scale down a k8s deployment that it has spun up on this condition (https://github.com/temporalio/temporal-worker-controller/blob/252364f6e7c473d79fbdb2a094cf04e579912ac9/internal/planner/planner.go#L764) 
- However, you could have the server be in a weird state divergence world where a version, that is processing some Pinned workloads, is not present in the worker deployment summaries list.
- In that world, the worker-controller would hit this condition and get ready to scale the k8s deployment for this version down since it thinks that the version has been deleted.
- I think we should have an additional DescribeVersion RPC call being made when we hit this special case so that we can truly understand if the version is missing (which would be the inference on a NotFound response from the RPC method) or not. 

### Environment/Versions

- OS and processor: [e.g. M1 Mac, x86 Windows, Linux]
- Temporal Version: [e.g. 1.14.0?] and/or SDK version
- Are you using Docker or Kubernetes or building Temporal from source?

Shivs11:
## Summary

- describe Kubernetes-backed Worker Deployment Versions that are absent from deployment summaries before treating them as `NotRegistered`
- add an integration regression test using a real pinned workflow while filtering the version from deployment summaries

## Behavioral impact

- Versions present in `VersionSummaries` retain their existing status precedence and rollout behavior.
- A Kubernetes build ID missing from `VersionSummaries` now triggers one `DescribeVersion` call:
  - success restores the version to controller state and prevents unsafe deletion
  - `NotFound` preserves the existing `NotRegistered` cleanup path
  - any other error stops reconciliation instead of continuing with incomplete state and potentially deleting the deployment
- A successfully described inactive target version can use the existing progressive-rollout unversioned-poller check.
- No planner, CRD, or public API behavior changes.

## Testing

- `go test ./internal/temporal ./internal/controller ./internal/planner -count=1`
- `go test -v -tags test_dep ./internal -run 'TestIntegration/missing-summary-version-is-described-before-deletion' -count=1`
- `go vet ./internal/temporal ./internal/controller ./internal/planner`
- `git diff --check`

Closes #550

### DIFF
--- a/internal/temporal/worker_deployment.go
+++ b/internal/temporal/worker_deployment.go
@@ -139,8 +139,8 @@ func GetWorkerDeploymentState(
 
 	// A version could be missing from the VersionSummaries in the odd event that there is
 	// state divergence between the deployment workflow's local state and the actual versions
-	// that are present. For versions that are known to TWC but absent from the Worker Deployment 
-	// description's version summaries, double-check their state before allowing them to map to 
+	// that are present. For versions that are known to TWC but absent from the Worker Deployment
+	// description's version summaries, double-check their state before allowing them to map to
 	// NotRegistered and get scaled down.
 	for buildID := range k8sDeployments {
 		if _, exists := state.Versions[buildID]; exists {
--- a/internal/temporal/worker_deployment.go
+++ b/internal/temporal/worker_deployment.go
@@ -181,6 +181,9 @@ func GetWorkerDeploymentState(
 				LastCurrentTime:   info.GetLastCurrentTime(),
 			},
 		)
+		if versionInfo == nil || versionInfo.Status == temporaliov1alpha1.VersionStatusNotRegistered {
+			return nil, fmt.Errorf("describe worker deployment version for buildID %q returned no registered status", buildID)
+		}
 		state.Versions[buildID] = versionInfo
 	}
 
--- a/internal/temporal/worker_deployment.go
+++ b/internal/temporal/worker_deployment.go
@@ -137,6 +137,52 @@ func GetWorkerDeploymentState(
 		state.Versions[version.DeploymentVersion.BuildId] = versionInfo
 	}
 
+	// A version can continue processing Pinned workloads even when it is absent
+	// from VersionSummaries. Confirm missing Kubernetes versions individually
+	// before allowing them to map to NotRegistered.
+	for buildID := range k8sDeployments {
+		if _, exists := state.Versions[buildID]; exists {
+			continue
+		}
+
+		desc, err := client.WorkflowService().DescribeWorkerDeploymentVersion(
+			ctx,
+			&workflowservice.DescribeWorkerDeploymentVersionRequest{
+				Namespace: namespace,
+				DeploymentVersion: &deploymentpb.WorkerDeploymentVersion{
+					DeploymentName: workerDeploymentName,
+					BuildId:        buildID,
+				},
+			},
+		)
+		if err != nil {
+			var notFound *serviceerror.NotFound
+			if errors.As(err, &notFound) {
+				continue
+			}
+			return nil, fmt.Errorf("unable to describe worker deployment version for buildID %q: %w", buildID, err)
+		}
+
+		info := desc.GetWorkerDeploymentVersionInfo()
+		if info == nil || info.GetDeploymentVersion() == nil {
+			return nil, fmt.Errorf("describe worker deployment version for buildID %q returned no version info", buildID)
+		}
+
+		versionInfo := versionInfoFromVersionSummary(
+			ctx, l, client, targetBuildID, strategy, depHandle, routingConfig,
+			&deploymentpb.WorkerDeploymentInfo_WorkerDeploymentVersionSummary{
+				DeploymentVersion: info.GetDeploymentVersion(),
+				Status:            info.GetStatus(),
+				DrainageInfo:      info.GetDrainageInfo(),
+				LastCurrentTime:   info.GetLastCurrentTime(),
+			},
+		)
+		if versionInfo == nil || versionInfo.Status == temporaliov1alpha1.VersionStatusNotRegistered {
+			return nil, fmt.Errorf("describe worker deployment version for buildID %q returned no registered status", buildID)
+		}
+		state.Versions[buildID] = versionInfo
+	}
+
 	return state, nil
 }
 
--- a/internal/temporal/worker_deployment.go
+++ b/internal/temporal/worker_deployment.go
@@ -137,9 +137,10 @@ func GetWorkerDeploymentState(
 		state.Versions[version.DeploymentVersion.BuildId] = versionInfo
 	}
 
-	// A version can continue processing Pinned workloads even when it is absent
-	// from VersionSummaries. Confirm missing Kubernetes versions individually
-	// before allowing them to map to NotRegistered.
+	// A version could be missing from the VersionSummaries in the odd event that there is
+	// state divergence between the deployment workflow's local state and the actual versions
+	// that are present. Confirm if those missing Kubernetes versions have any existing workloads
+	// running before allowing them to map to NotRegistered.
 	for buildID := range k8sDeployments {
 		if _, exists := state.Versions[buildID]; exists {
 			continue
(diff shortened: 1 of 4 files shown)


internal/tests/internal/not_registered_integration_test.go
```diff
@@ -0,0 +1,296 @@
...
+
+	filtered := proto.Clone(resp).(*workflowservice.DescribeWorkerDeploymentResponse)
+	info := filtered.GetWorkerDeploymentInfo()
+	if info == nil {
+		return filtered, nil
+	}
+
+	summaries := info.VersionSummaries[:0]
+	omitted := false
+	for _, summary := range info.VersionSummaries {
```
Shivs11, line 67: basically mimicking the removal of a version summary from the deployment summary list

internal/temporal/worker_deployment.go
```diff
@@ -135,9 +135,76 @@ func GetWorkerDeploymentState(
...
+
+	if routingConfig.CurrentDeploymentVersion != nil &&
+		deploymentName == routingConfig.CurrentDeploymentVersion.DeploymentName &&
+		buildID == routingConfig.CurrentDeploymentVersion.BuildId {
+		out.Status = temporaliov1alpha1.VersionStatusCurrent
+	} else if routingConfig.RampingDeploymentVersion != nil &&
+		deploymentName == routingConfig.RampingDeploymentVersion.DeploymentName &&
+		buildID == routingConfig.RampingDeploymentVersion.BuildId {
+		out.Status = temporaliov1alpha1.VersionStatusRamping
+	}
```
jaypipes: This is the same block of code that @carlydf 's [PR reworks](https://github.com/temporalio/temporal-worker-controller/pull/554), so perhaps we should wait on that and rework this?

internal/temporal/worker_deployment.go
```diff
@@ -151,68 +218,91 @@ func versionInfoFromVersionSummary(
...
+			versionInfo.Status = temporaliov1alpha1.VersionStatusDrained
+
+			// Extract DrainedSince directly from the summary's drainage info,
+			// avoiding a per-summary DescribeVersion call.
+			if summary.DrainageInfo != nil && summary.DrainageInfo.LastChangedTime != nil {
+				drainedSince := summary.DrainageInfo.LastChangedTime.AsTime()
+				versionInfo.DrainedSince = &drainedSince
+			}
+		}
 	}
```
jaypipes: same comment as above... should probably rebase this patch onto Carly's PR 554

Shivs11:
going to hold off on this one for now, and wait for @carlydf's PR to go through!

internal/temporal/worker_deployment.go
```diff
@@ -137,9 +137,76 @@ func GetWorkerDeploymentState(
...
+	} else if routingConfig.RampingDeploymentVersion != nil &&
+		deploymentName == routingConfig.RampingDeploymentVersion.DeploymentName &&
+		buildID == routingConfig.RampingDeploymentVersion.BuildId {
+		out.Status = temporaliov1alpha1.VersionStatusRamping
+	}
+
+	return out
+}
+
+func shouldCheckUnversionedPollers(
```
carlydf: nit: for me, moving this logic into a `shouldCheckUnversionedPollers` helper makes `versionInfoFromVersionSummary` a bit harder to follow because the important `Inactive` and "first progressive rollout" conditions are hidden behind the helper. It also for me makes it a bit harder to focus on the actual logic changes in the PR

Could we leave the existing condition inline and keep this PR focused on confirming versions missing from summaries?

I'm going to push a commit focussed on simplifying the diff of this PR (essentially addressing my own comment), but feel free to revert and address it differently, or disagree ❤️

carlydf: ohh I now see why you did that. I will massage differently

internal/temporal/worker_deployment.go
```diff
@@ -181,6 +181,9 @@ func GetWorkerDeploymentState(
 				LastCurrentTime:   info.GetLastCurrentTime(),
 			},
 		)
+		if versionInfo == nil || versionInfo.Status == temporaliov1alpha1.VersionStatusNotRegistered {
+			return nil, fmt.Errorf("describe worker deployment version for buildID %q returned no registered status", buildID)
+		}
```
Shivs11, line 187: i think this is just defensive programming, since we do have code in the server which can set the status to unregistered if something goes wrong while calculating the drainage status of the version itself

in that case, the right thing to do is to error out and not just scale the version down.

internal/temporal/worker_deployment.go
```diff
@@ -137,6 +137,56 @@ func GetWorkerDeploymentState(
 		state.Versions[version.DeploymentVersion.BuildId] = versionInfo
 	}
 
+	// A version could be missing from the VersionSummaries in the odd event that there is
+	// state divergence between the deployment workflow's local state and the actual versions
+	// that are present. Confirm if those missing Kubernetes versions have any existing workloads
+	// running before allowing them to map to NotRegistered.
```
carlydf:
```suggestion
	// that are present. For versions that are known to TWC but absent from the Worker Deployment 
	// description's version summaries, double-check their state before allowing them to map to 
	// NotRegistered and get scaled down.
```

internal/tests/internal/not_registered_integration_test.go
carlydf: nice

carlydf (approved):

internal/temporal/worker_deployment.go
carlydf: yes I agree. if the version was found, but somehow returned NotRegistered later, we should not scale down.

internal/temporal/worker_deployment.go
```diff
@@ -137,6 +137,57 @@ func GetWorkerDeploymentState(
...
 
+	// A version could be missing from the VersionSummaries in the odd event that there is
+	// state divergence between the deployment workflow's local state and the actual versions
+	// that are present. For versions that are known to TWC but absent from the Worker Deployment
+	// description's version summaries, double-check their state before allowing them to map to
+	// NotRegistered and get scaled down.
+	for buildID := range k8sDeployments {
+		if _, exists := state.Versions[buildID]; exists {
+			continue
+		}
```
jaypipes, line 148: would be great to put a log line in here:

```go
l.Info(
    "warning: Kubernetes Deployment not found in WorkerDeployment VersionSummaries. "+
        "Possible inconsistent Temporal server state.",
    "buildID", buildID,
)
```

internal/temporal/worker_deployment.go
```diff
@@ -137,6 +137,57 @@ func GetWorkerDeploymentState(
...
+			ctx, l, client, targetBuildID, strategy, depHandle, routingConfig,
+			&deploymentpb.WorkerDeploymentInfo_WorkerDeploymentVersionSummary{
+				DeploymentVersion: info.GetDeploymentVersion(),
+				Status:            info.GetStatus(),
+				DrainageInfo:      info.GetDrainageInfo(),
+				LastCurrentTime:   info.GetLastCurrentTime(),
+			},
+		)
+		if versionInfo == nil || versionInfo.Status == temporaliov1alpha1.VersionStatusNotRegistered {
+			return nil, fmt.Errorf("describe worker deployment version for buildID %q returned no registered status", buildID)
```
jaypipes, line 186:
```suggestion
			return nil, fmt.Errorf("describe worker deployment version for buildID %q returned NotRegistered status", buildID)
```

internal/tests/internal/not_registered_integration_test.go
```diff
@@ -0,0 +1,296 @@
...
+	// DescribeVersion fallback, the first one classifies v1 as NotRegistered and
+	// deletes its Kubernetes Deployment even though its pinned workflow is still open.
+	eventually(t, 30*time.Second, time.Second, func() error {
+		if filteringService.filteredCalls.Load() < 2 {
+			return fmt.Errorf("waiting for reconciliations with the filtered version summary")
+		}
+
+		var deployment appsv1.Deployment
+		if err := k8sClient.Get(ctx, types.NamespacedName{Name: deploymentNameV1, Namespace: namespace}, &deployment); err != nil {
+			return fmt.Errorf("v1 Deployment was deleted: %w", err)
```
jaypipes, line 276: Technically, this should check whether `err` was NotFound...

Issue #415, opened 2026-07-01
[Bug] Reconciliation loop running too hot


gleason-m:
### Describe the bug

The reconciliation loop runs multiple times per second per TWD. This shows count of "Running Reconcile loop" logs during the upgrade, rollback, and roll-forward.

<img width="1811" height="708" alt="Image" src="https://github.com/user-attachments/assets/c6c8b4ac-fc75-4aba-a559-6c054ed48774" />

I believe it is related to the non-deterministic map iteration in mapToStatus: https://github.com/temporalio/temporal-worker-controller/blob/a2e3ccdfeadb7a769f74e5cff7ba1cfa4928969d/internal/controller/state_mapper.go#L57

My thinking is this map causes the reconciler's final write to output the deprecatedVersions array in a non-deterministic order and triggers reconciliation again even if no status actually changes (just ordering changes) https://github.com/temporalio/temporal-worker-controller/blob/a2e3ccdfeadb7a769f74e5cff7ba1cfa4928969d/internal/controller/worker_controller.go#L399

Here can see the TWD version climbing several times per second:
```bash
gleasonm@symbiote-dev:~$ kubectl get twd $NAME -n $NS -w -o custom-columns='RV:.metadata.resourceVersion,GEN:.metadata.generation,TARGET:.status.targetVersion.status' | while IFS= read -r line; do printf '%s  %s\n' "$(date '+%H:%M:%S.%3N')" "$line"; done
20:56:33.807  RV          GEN   TARGET
20:56:33.811  401336452   264   Current
20:56:33.830  401336457   264   Current
20:56:33.893  401336459   264   Current
20:56:33.950  401336461   264   Current
20:56:34.009  401336463   264   Current
20:56:34.077  401336466   264   Current
20:56:34.135  401336469   264   Current
20:56:34.188  401336472   264   Current
20:56:34.248  401336474   264   Current
20:56:34.310  401336479   264   Current
20:56:34.368  401336483   264   Current
20:56:34.431  401336486   264   Current
20:56:34.497  401336489   264   Current
20:56:34.550  401336492   264   Current
20:56:34.616  401336497   264   Current
20:56:34.671  401336499   264   Current
20:56:34.743  401336502   264   Current
20:56:34.826  401336506   264   Current
20:56:34.960  401336509   264   Current
20:56:35.016  401336512   264   Current
```

and here can see there is no diff when I sort by buildID:
```bash
[...]

PR #557, opened 2026-08-29
fix(controller): only send status updates when the status changed

tomba7:
Part 2 of #415

## What was changed

`Reconcile` now snapshots the status it read from the API server and compares the recomputed status against it before writing. If it is semantically identical, the status is not sent.

## Why?

Every reconcile rebuilds the whole status and writes it unconditionally, then requeues after 10s (requeue interval). For a settled WorkerDeployment the recomputed status at each reconcile pass is identical and each WorkerDeployment sends 6 status updates a minute that the API server has to process and then discard. This gets amplified at scale (100 WDs = 10 write RPS which carry no new information)

The scope of this fix is the status write at the end of `Reconcile`, as it runs for every WorkerDeployment on every pass and carries effectively all of the redundant volume. The other status writes are left alone deliberately.  For example, the blocked conditions path reflects error scenarios where the controller cannot make progress and are expected to be written.

## Unit Testing
   - Verifies that once the rollout settles, a reconcile that recomputes the same status does not send it back to the API server.
   - Verifies that once the rollout settles, a change that makes the status differ again is still written to the API server.
   - `go test ./...` passes.

### DIFF
--- a/internal/controller/reconciler_events_test.go
+++ b/internal/controller/reconciler_events_test.go
@@ -192,6 +192,25 @@ func assertNoEventEmitted(t *testing.T, events []string, reason string) {
 	}
 }
 
+// countWDStatusWrites returns interceptor funcs that count status writes issued for a
+// WorkerDeployment, passing each one through to the fake client.
+func countWDStatusWrites(count *int) interceptor.Funcs {
+	return interceptor.Funcs{
+		SubResourceUpdate: func(
+			ctx context.Context,
+			c client.Client,
+			subResourceName string,
+			obj client.Object,
+			opts ...client.SubResourceUpdateOption,
+		) error {
+			if _, ok := obj.(*temporaliov1alpha1.WorkerDeployment); ok && subResourceName == "status" {
+				*count++
+			}
+			return c.SubResource(subResourceName).Update(ctx, obj, opts...)
+		},
+	}
+}
+
 // ─── Stub types ──────────────────────────────────────────────────────────────
 
 // stubWDHandle implements sdkclient.WorkerDeploymentHandle with configurable per-method errors.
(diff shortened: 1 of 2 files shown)


internal/controller/reconciler_events_test.go
```diff
@@ -680,6 +699,76 @@ func TestReconcile_DescribeWorkerDeploymentNotFound(t *testing.T) {
...
+
+	writes := 0
+	r, _ := newTestReconcilerWithInterceptors([]client.Object{twd, tc}, countWDStatusWrites(&writes))
+	r.TemporalClientPool.SetClientForTesting(
+		noCredsPoolKey(tc.Spec.HostPort, twd.Spec.WorkerOptions.TemporalNamespace),
+		newStubTemporalClient(nil),
+	)
+
+	req := ctrl.Request{NamespacedName: types.NamespacedName{Name: twd.Name, Namespace: twd.Namespace}}
+	// Reconcile enough times to reach steady state (it settles after two reconciles)
```
jaypipes: How do we know that it settles after two reconciles? :) Please update the code comment with the explanation.

jaypipes (approved):
👍 Feel free to address the one little nit about a comment in the test.

internal/controller/reconciler_events_test.go
```diff
@@ -736,6 +759,78 @@ func TestReconcile_DescribeWorkerDeploymentNotFound(t *testing.T) {
...
+	r, _ := newTestReconcilerWithInterceptors([]client.Object{twd, tc}, countWDStatusWrites(&writes))
+	r.TemporalClientPool.SetClientForTesting(
+		noCredsPoolKey(tc.Spec.HostPort, twd.Spec.WorkerOptions.TemporalNamespace),
+		newStubTemporalClient(nil),
+	)
+
+	req := ctrl.Request{NamespacedName: types.NamespacedName{Name: twd.Name, Namespace: twd.Namespace}}
+	// Reconcile enough times to reach steady state. The first pass computes the status before creating
+	// the k8s Deployment, then the second pass adds the k8s Deployment reference to the status once it
+	// exists. On the third pass there is nothing new to write.
```
jaypipes, line 781: 👍

jaypipes (approved):
molto bene!

PR #558, opened 2026-08-29
helm: document extraEnv proxy usage and validate EnvVar shape

### LINKED ISSUE #510: Support injecting extra environment variables (e.g. HTTP proxy) into the controller container

When deploying the temporal-worker-controller in a Kubernetes cluster behind a corporate proxy, there is currently no way to inject environment variables such as `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY` into the manager container via the Helm chart.

The Go runtime, gRPC (used by the Temporal SDK), and net/http (used by controller-runtime / client-go) all natively respect the standard proxy environment variables. However, the Helm chart does not expose an extraEnv (or similar) field in values.yaml, making it impossible to pass these variables without forking or patching the chart.

Add an `extraEnv` field to values.yaml that allows users to inject arbitrary environment variables into the manager container. This is a common pattern in Helm charts.

Example Usage : 

```yaml
extraEnv:
  - name: HTTP_PROXY
    value: "http://proxy.corp.example.com:8080"
  - name: HTTPS_PROXY
    value: "http://proxy.corp.example.com:8080"
  - name: NO_PROXY
    value: "localhost,127.0.0.1,.svc,.cluster.local,10.0.0.0/8"
```

kriyanshii:
extraEnv already exists; document the corporate-proxy and valueFrom cases, require a name in the chart schema, and cover both in Helm CI.

## What was changed
`extraEnv` already exists on the manager container (added in #532). This PR documents the corporate-proxy and `valueFrom` cases from #510, requires each entry to have a non-empty `name` in `values.schema.json`, and adds Helm CI coverage that extraEnv renders (literal + secret ref) and that a nameless entry is rejected.

## Why?
Operators behind a corporate proxy need to pass `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY` into the manager. The Go runtime, gRPC, and net/http already honor those variables, but #510 stayed open because the chart only documented `SSL_CERT_FILE` and did not validate or test the EnvVar shape.

## Checklist

1. Closes #510
2. How was this tested:
- `helm template test-release helm/temporal-worker-controller | grep HTTP_PROXY` prints nothing (empty extraEnv does not leak proxy vars)
- `helm template` with `extraEnv` `HTTP_PROXY` + secret-backed `API_TOKEN` renders both on the manager container after the chart-owned env vars
- `helm template` with `extraEnv[0].value` and no `name` fails schema validation (`missing property 'name'`)
- `helm lint --strict helm/temporal-worker-controller` passes
3. Any docs updates needed?
No external docs. `values.yaml` and `values.schema.json` now include the proxy and `valueFrom` examples.

### DIFF
--- a/.github/workflows/helm-validate.yml
+++ b/.github/workflows/helm-validate.yml
@@ -62,6 +62,29 @@ jobs:
           helm template test-release helm/temporal-worker-controller \
             --set 'rbac.restrictWatchNamespaces={ns-a,ns-b}'
 
+      - name: Template with extra manager environment variables
+        run: |
+          set -euo pipefail
+          rendered=$(helm template test-release helm/temporal-worker-controller \
+            --set-string 'extraEnv[0].name=HTTP_PROXY' \
+            --set-string 'extraEnv[0].value=http://proxy.example.com:8080' \
+            --set-string 'extraEnv[1].name=API_TOKEN' \
+            --set-string 'extraEnv[1].valueFrom.secretKeyRef.name=controller-secrets' \
+            --set-string 'extraEnv[1].valueFrom.secretKeyRef.key=api-token')
+          echo "$rendered" | grep -q 'name: HTTP_PROXY'
+          echo "$rendered" | grep -q 'value: http://proxy.example.com:8080'
+          echo "$rendered" | grep -q 'name: API_TOKEN'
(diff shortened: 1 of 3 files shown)


jaypipes (approved):
great stuff, thank you @kriyanshii :)

Issue #536, opened 2026-08-19
Automate syncing of Helm `values.schema.json` file


jaypipes:
The `helm/temporal-worker-controller/values.schema.json` file is not kept in sync with new changes to the Helm Chart's `values.yaml` file.

We should add some automation that keeps it in sync. It's a little tricky because the JSONSchema generation from the `helm schema` CLI command is kinda funky and requires a bit of massaging.

kriyanshii:
@jaypipes is this open?

jaypipes:
> [@jaypipes](https://github.com/jaypipes) is this open?

@kriyanshii yes it is! :)

kriyanshii:
please assign this to me.

jaypipes:
@kriyanshii done!

Note that you may want to break this into two discrete tasks due to the funkiness with the `helm schema` command.

I'd suggest first creating a step in the [`helm-lint` GH Actions workflow job](https://github.com/temporalio/temporal-worker-controller/blob/main/.github/workflows/helm-validate.yml#L12) that runs a Bash script that checks if there were non-comment changes to the `helm/temporal-worker-controller/values.yaml` file in the pull request but there were no changes to the `values.schema.json` file, returning a non-zero exit code if so.

Then, in a followup PR you can work on a script/tool that examines the values.yaml file and automatically generates the `values.schema.json` file contents. You can use the `helm schema` CLI command or not, depending on your tolerance for its quirks.

PR #561, opened 2026-08-31
helm: fail CI when values.yaml changes without a schema update

### LINKED ISSUE #562: Detect drift in `values.schema.json` file
Create a step in the [helm-lint GH Actions workflow job](https://github.com/temporalio/temporal-worker-controller/blob/main/.github/workflows/helm-validate.yml#L12) that runs a Bash script that checks if there were non-comment changes to the helm/temporal-worker-controller/values.yaml file in the pull request but there were no changes to the values.schema.json file, returning a non-zero exit code if so.

kriyanshii:
Catch PRs that change Helm values but forget values.schema.json, as the first step of #536.

## What was changed
Added a `helm-lint` CI step that runs `hack/check-helm-values-schema.sh`. The script fails when `helm/temporal-worker-controller/values.yaml` has non-comment changes in the PR and `values.schema.json` was not updated. Comment-only or whitespace-only edits are ignored.

This is the first of the two steps from #536 (detect drift). Auto-generating `values.schema.json` can follow in a later PR.

## Why?
`values.schema.json` has been falling out of date when Helm values change. A CI check makes that mismatch visible on the PR instead of relying on reviewers to notice.

## Checklist

1. Closes #562

2. How was this tested:
   - `./hack/check-helm-values-schema.sh` on this branch skips (no `values.yaml` change)
   - Adding `newKey: true` to `values.yaml` without touching the schema fails the script
   - Comment-only `values.yaml` changes still pass
   - Updating both files passes

3. Any docs updates needed?
   No. Schema generation is still manual; this PR only adds the check.


jaypipes:
@kriyanshii awesome, thanks so much for the PR! I created a sub-issue (#562) specifically for the drift-detection and editing this PR summary to close out the sub-issue instead of the parent.

hack/check-helm-values-schema.sh
```diff
@@ -0,0 +1,83 @@
...
+# CI (pull_request): compare against the PR base SHA.
+# CI (push): compare against the previous commit on the branch.
+# Local: compare against the merge-base with origin/main (or main).
+set -euo pipefail
+
+ROOT="$(git rev-parse --show-toplevel)"
+VALUES="helm/temporal-worker-controller/values.yaml"
+SCHEMA="helm/temporal-worker-controller/values.schema.json"
+
+resolve_base() {
```
jaypipes: Instead of this `resolve_base` function, consider simplifying this script by having a single `GIT_REF` environment variable, defaulting to `HEAD`, and have the callsite in the GH action workflow set this `GIT_REF` variable appropriately. You can then make the call signature of this script more user-friendly and update the above docstring accordingly to something like this:

```sh
# check-helm-values.sh compares the contents of the helm/temporal-worker-controller/values.yaml file
# between the current working tree and the Git ref pointed to by the GIT_REF environment variable.
# If any non-comment changes have been made to the values.yaml file, the values.json.schema file
# is checked for a non-empty diff. If the diff is NOT empty, the script returns 0 (success), otherwise
# the script returns a non-0 exit code.
# 
# GIT_REF defaults to HEAD, meaning running this script against a dirty working tree effectively
# checks that the committer has remembered to update the JSONSchema of the values.yaml file
# if they have made any changes to the values.yaml contents.

GIT_REF=${GIT_REF:-$(git rev-parse --verify --quiet HEAD)}

base=$GIT_REF
...
```

.github/workflows/helm-validate.yml
```diff
@@ -24,6 +26,12 @@ jobs:
       - name: Lint chart
         run: helm lint --strict helm/temporal-worker-controller
 
+      - name: Check values.schema.json stays in sync with values.yaml
+        env:
+          PR_BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          PUSH_BEFORE_SHA: ${{ github.event.before }}
```
jaypipes: See my comment below. You should be able to simplify the above to just this:

```yaml
      - name: Check values.schema.json stays in sync with values.yaml
        env:
          GIT_REF: ${{ github.event_name == 'pull_request' && github.event.pull_request.base.sha || github.event.before }}
```

jaypipes (changes requested):

hack/check-helm-values-schema.sh
kriyanshii: done

.github/workflows/helm-validate.yml
kriyanshii: check now

jaypipes (approved):
👍 thank you @kriyanshii!

PR #564, opened 2026-09-01
introduce WorkerDeploymentSpec.Deployment

jaypipes:
Adds a new `WorkerDeploymentSpec.Deployment` field of type
`*appsv1.DeploymentSpec` that allows users to override the desired
configuration of the Kubernetes Deployment created by Temporal Worker
Controller associated with a WorkerDeploymentVersion (Build ID).

Marks the existing `WorkerDeploymentSpec.Template`, `Replicas`,
`MinReadySeconds` and `ProgressDeadlineSeconds` fields as deprecated and
to use the corresponding fields on `WorkerDeploymentSpec.Deployment` to
control those aspects of the per-WDV Kubernetes Deployment.

Issue #496

Co-authored-by: @otan


api/v1alpha1/workerdeployment_types.go
carlydf:
```
type SettingsForEachVersion {
  replicas: 5
  MaxUnavailable: 2
  MaxSurge: 3
}
```

```
type SettingsForEachVersion {
  replicas: 5
  MaxUnavailable: 2
  MaxSurge: 3
  byStatusSettings:
  - target:
    replicas: 5
  - draining:
    replicas: 1
}
```

jaypipes:
@otan Hi Oliver! We've had some weird issues with the CLA assistant thing... any chance you can re-sign the CLA with your Github account? Thank ou!

otan:
you have my sword

internal/planner/planner.go
```diff
@@ -682,6 +683,43 @@ func updateDeploymentWithPodTemplateSpec(
...
+	k8sState *k8s.DeploymentState,
+	spec *temporaliov1alpha1.WorkerDeploymentSpec,
+) *appsv1.Deployment {
+	existingDeployment, exists := k8sState.Deployments[buildID]
+	if !exists {
+		return nil
+	}
+
+	desired := appsv1.DeploymentStrategy{
+		Type: appsv1.RollingUpdateDeploymentStrategyType,
```
Shivs11, line 710: maybe my understanding of this is flawed, but here goes: I see that the default type here, for the strategy, is a rolling update. However, say tomorrow some user wanted to have this set to Recreate. 

- Would it be possible for them to set this? Are we saying that for worker-controller managed versions, we would never have this strat at all? 
- Moreover, without this PR, if someone today were to manually update their deployments and set this "Recreate" strategy, the worker-controller would not care. However, after this PR goes in, the controller would overwrite this field. Do we want this?

Shivs11 (commented):
just one comment

api/v1alpha1/workerdeployment_webhook.go
```diff
@@ -109,6 +120,21 @@ func validateRolloutStrategy(s RolloutStrategy) []*field.Error {
 		}
 	}
 
+	if isExplicitlyZeroIntOrString(s.MaxUnavailable) &&
+		isExplicitlyZeroIntOrString(s.MaxSurge) {
+		allErrs = append(
+			allErrs,
+			field.Invalid(
+				field.NewPath("spec.rollout.maxUnvailable"),
```
Shivs11, line 128: typo here -- should be maxUnavailable

api/v1alpha1/workerdeployment_types.go
```diff
@@ -480,6 +481,48 @@ type RolloutStrategy struct {
 	// - "Progressive"
 	Strategy DefaultVersionUpdateStrategy `json:"strategy"`
 
+	// The maximum number of pods that can be unavailable during an in-place
```
Shivs11, line 484: one second....would this also bring changes to the now deprecated previous version of the CRD? (TWD)

internal/planner/planner_test.go
```diff
@@ -25,6 +25,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/intstr"
```
Shivs11, line 28: I think each of the files that have been touched have unit tests, but not seeing an e2e integration test which could maybe test out that updating a WD would update it's existing child deployment with the intended numbers

helm/temporal-worker-controller-crds/templates/temporal.io_clusterconnections.yaml
```diff
@@ -71,6 +79,9 @@ spec:
             x-kubernetes-validations:
             - message: Only one of mutualTLSSecretRef or apiKeySecretRef may be set
               rule: '!(has(self.mutualTLSSecretRef) && has(self.apiKeySecretRef))'
+            - message: tls.caCertSecretRef cannot be combined with mutualTLSSecretRef;
+                bundle the CA into that secret's own ca.crt key instead
+              rule: '!(has(self.mutualTLSSecretRef) && has(self.tls) && has(self.tls.caCertSecretRef))'
```
Shivs11, line 84: isn't this an unrelated change to the crds?

api/v1alpha1/workerdeployment_types.go
carlydf: I don't think it should?

api/v1alpha1/workerdeployment_types.go
carlydf, line 476: I think the behavior and semantics of these fields make sense, but I’m concerned about putting them under `RolloutStrategy` (we discussed this in yesterday's meeting and @jlegrone shares this concern).

`spec.rollout` describes how to roll out a target version (how to direct Temporal traffic over the rollout time period). These new fields instead describe availability and temporary capacity while replacing pods belonging to one version, which is independent: a pod replacement can happen without a Temporal version rollout. I think putting these settings under `RolloutStrategy` conflates the too in a way that isn't helpful to users (doesn't add simplicity to their lives).

Could we model the user-visible behavior instead? something like

```yaml
spec:
   workerVersionSettings:
     podReplacement:
       maxUnavailable: 5%
       maxSurge: 0
   rollout:
     strategy: Progressive
```

`workerVersionSettings` avoids the term `perVersion` while describing the scope: these settings apply to each WDV individually. `podReplacement` describes the behavior users care about without committing us to k8s Deployment as the backing k8s resource.

This also leaves a reasonable place for other version-scoped settings in the future (ie. replica-count or build-ID/version-status-specific overrides for certain settings) without putting unrelated settings under Temporal rollout configuration. Not saying that we add any of those now (or ever unless there is compelling reason), but I think we should put these new version-specific pod-replacement settings in a place where we would be comfortable adding similarly scoped settings in the future.

Other than the placement of the new fields, the semantics of how the fields are used looks good to me! I think it's fine to say that only rolling updates within a version's replicas are allowed, and not leave room for other Deployment update strategies

carlydf (commented):
Other than the placement of the new fields, the semantics of how the fields are used looks good to me! I think it's fine to say that only rolling updates within a version's replicas are allowed, and not leave room for other Deployment update strategies

helm/temporal-worker-controller-crds/templates/temporal.io_clusterconnections.yaml
jaypipes: Yeah, I'm actually not sure where this came from. I will undo it.

api/v1alpha1/workerdeployment_types.go
eniko-dif: I think I'd prefer keeping these options under rollout. My reasoning is that `workerVersionSettings` and `podReplacement` are things users would have to look up (even if the naming is intuitive) to understand the exact meaning. If the options stay under rollout, I feel like it's more "obvious" what it refers to since those options are related to deployment strategy.

api/v1alpha1/workerdeployment_types.go
jaypipes: correct, we explicitly don't make changes to the deprecated CRDs.

api/v1alpha1/workerdeployment_webhook.go
jaypipes: thanks! fixed.

internal/planner/planner.go
jaypipes: > maybe my understanding of this is flawed, but here goes: I see that the default type here, for the strategy, is a rolling update. However, say tomorrow some user wanted to have this set to Recreate.

Not the default type... it's the *only* strategy type supported.

> * Would it be possible for them to set this? Are we saying that for worker-controller managed versions, we would never have this strat at all?

I don't really see a use case for Recreate strategy when it comes to TWC.

The Kubernetes Deployment's `spec.strategy.type` only influences how the Kubernetes Deployment controller handles mutations to the Kubernetes Deployment's `spec.template.spec` field. 

TWC is the only thing that creates the Kubernetes Deployments associated with each individual WorkerDeploymentVersion that comprise the "rainbow deployment" of the Temporal WorkerDeployment. TWC is the thing that manages the setting of those Kubernetes Deployments' `spec.template.spec` field.

The only code path in TWC that ever updates the the Kubernetes Deployment's `spec.template.spec` field [restricts the update to only the Kubernetes Deployment representing the current/target build ID](https://github.com/temporalio/temporal-worker-controller/blob/79ebf0b5029fd85806f44ca9480debcc456f625a/internal/planner/planner.go#L699-L704). TWC never updates non-current Kubernetes Deployment `spec.template.spec` field contents. This is deliberate, I believe: we did not want a change to, say, the cpu/memory resource in the WorkerDeployment's `spec.template.spec..resources` field to impact drained or draining build IDs, only the current build ID (and future target versions).

The [original feature request](https://github.com/temporalio/temporal-worker-controller/issues/496) for this PR describes how @otan's team occasionally performs manual `kubectl rollout restart` commands on the Kubernetes Deployment associated with the current buildID for two primary reasons:

* They have made non-breaking changes to the worker image referenced in the Kubernetes Deployment's `spec.template.spec.containers[*].image` field and pushed a new image with the same tag and want to have the Pods in the Kubernetes Deployment restart to pick up that new image.
* The rollout may have gotten "stuck" due to some reason and they want to "kick" the rollout into gear after making a change to a field in the Kubernetes Deployment's `spec.template.spec` to fix whatever caused the rollout to get stuck.

In both cases, the desire is to be able to have more fine-grained control over the Kubernetes Deployment's restart of the rollout in order to force Kubernetes Deployment controller to be more conservative with its strategy of killing existing Pods and starting new Pods. I don't see how the Recreate rollout deployment strategy would be something anyone would be looking for...

api/v1alpha1/workerdeployment_types.go
carlydf: I guess my point is that the pod replacement strategy of a specific k8s Deployment is actually totally orthogonal to how you roll out a new target version. So it should not be under rollout.strategy.

jaypipes:
@carlydf @otan @jlegrone @Shivs11 OK, I went back to the drawing board on this. I've added a `WorkerDeploymentSpec.Deployment` field that is of type `*appsv1.DeploymentSpec`. This allows users to configure all the Kubernetes Deployment settings they want to, including the deployment strategy, the rolling update `maxUnavailable`/`maxSurge` values or anything else. The `Replicas`, `MinReadySecond`, `Template` and `ProgressDeadlineSeconds` fields in `WorkerDeploymentSpec` have been deprecated.

PR #567, opened 2026-09-02
fix(controller): skip ClusterConnection watch when namespace-scoped

aldredb:
Skip the `ClusterConnection` watch when the controller is namespace-scoped.

`ClusterConnection` (added in #549) is the only cluster-scoped CRD in the set, but the watch is
registered unconditionally. An install with `rbac.restrictWatchNamespaces` gets a namespaced
`Role`, which can never authorize it, so the controller logs this on every retry, per namespace,
forever:

```
Failed to watch *v1alpha1.ClusterConnection: failed to list *v1alpha1.ClusterConnection:
clusterconnections.temporal.io is forbidden: User "system:serviceaccount:<ns>:<sa>" cannot
list resource "clusterconnections" in API group "temporal.io" at the cluster scope
```

Reconciliation is unaffected, so this is log noise and alert pressure rather than an outage.

### DIFF
--- a/api/v1alpha1/workerdeployment_types.go
+++ b/api/v1alpha1/workerdeployment_types.go
@@ -139,6 +139,11 @@ const (
 	// referenced Connection resource cannot be found.
 	ReasonConnectionNotFound = "ConnectionNotFound"
 
+	// ReasonClusterConnectionUnsupported is set on ConditionProgressing=False when a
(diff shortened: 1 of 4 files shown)


eniko-dif (approved):
Thanks for the fix, this looks good to me!

jaypipes (approved):
Awesome, thank you @aldredb and sorry for this oversight!

Issue #316, opened 2026-05-05
Clean up and standardize webhook configuration and upgrade documentation


anatolebeuzon:
### What are you really trying to do?

Upgrading from 1.4.0 to 1.5.2

### Describe the bug

If `values.yaml` looks like the following:

```yaml
  certmanager:
    enabled: false
```

then the container for 1.5.2 will be stuck in `ContainerCreating` with events:

```
Events:
  Type     Reason       Age                   From               Message
  ----     ------       ----                  ----               -------
  Normal   Scheduled    8m47s                 default-scheduler  ...
  Warning  FailedMount  33s (x12 over 8m47s)  kubelet            MountVolume.SetUp failed for volume "cert" : secret "webhook-server-cert" not found
```

I believe that's a side effect of https://github.com/temporalio/temporal-worker-controller/pull/217.

I thought a workaround would to install and enable certmanager, even if we're not using it:

```yaml
  certmanager:
    install: true
    enabled: true
```

but even this does not work as it doesn't install CRDs:

```
Error: UPGRADE FAILED: [resource mapping not found for name: "temporal-worker-controller-serving-cert" namespace: "cloud-agents" from "": no matches for kind "Certificate" in version "cert-manager.io/v1"
ensure CRDs are installed first, resource mapping not found for name: "temporal-worker-controller-selfsigned-issuer" namespace: "..." from "": no matches for kind "Issuer" in version "cert-manager.io/v1"
ensure CRDs are installed first]
```

I suppose there are two problems:
* the chart should probably not mount `webhook-server-cert` if `certmanager` is `false`
* if `certmanager.install` is `true`, then it should also install CRDs

jaypipes:
@niyomukiza-mechack I'd like you to take a look into this please. What @anatolebeuzon is running into is a poor user experience for upgrading TWC and for configuring and reasoning about the webhooks.

The root cause of the problem is related to the history of webhook handling in TWC, which I have summarized below with links to the relevant code and YAML files.

## v1.4.0

[...]

PR #574, opened 2026-09-08
fix: make webhook cert Secret name configurable

niyomukiza-mechack:
## What was changed

Replaced the hard-coded webhook-server-cert Secret name with a new configurable Helm value `webhook.certSecretName`. Previously, the Secret name was hard coded in both `templates/manager.yaml` and `templates/certmanager.yaml` so users who bring their own TLS certificate had to name their Secret exactly `webhook-server-cert`  no other name worked. Now they can set `webhook.certSecretName` to match whatever Secret they've created. The default is still `webhook-server-cert` so existing installations are unaffected.

Files changed:

`values.yaml` - added `webhook.certSecretName` with default value updated the caBundle comment to reference the value instead of the literal name
`templates/manager.yaml` - replaced hard coded secretName: `webhook-server-cert` with `{{ .Values.webhook.certSecretName }}`
`templates/certmanager.yaml` - same replacement
`values.schema.json` - added certSecretName property to the webhook schema

## Why?

Context: https://github.com/temporalio/temporal-worker-controller/issues/316

The webhook cert Secret name was hard coded as webhook-server-cert in two template files with no way to configure it. This caused a poor user experience: users who set `certmanager.enabled: false` and brought their own TLS certificate were forced to name their Secret exactly `webhook-server-cert` or the controller pod would get stuck in ContainerCreating forever. This is part of the larger webhook configuration cleanup tracked in #316.

## Checklist

1. Closes 

2. How was this tested:

Verified the new value flows through to both templates
Custom Secret name renders in both manager.yaml and certmanager.yaml and Default value preserves backward compatibility

3. Any docs updates needed?


tomba7 (approved):
LGTM

jaypipes (approved):
👍 yep, makes sense to me. thanks @niyomukiza-mechack :)

PR #575, opened 2026-09-08
Fix: remove cert-manager subchart dependency

### LINKED ISSUE #316: Clean up and standardize webhook configuration and upgrade documentation

niyomukiza-mechack:
## What was changed

Removed the embedded cert-manager subchart dependency from the TWC Helm chart.

Files changed:

Chart.yaml - removed the dependencies: block that declared cert-manager as a conditional subchart
Chart.lock - deleted (no more dependencies to lock)
values.yaml - removed certmanager.install and updated the enabled comment to say "installed independently" instead of "via the subchart above or separately"
Makefile - removed the helm dependency build target and its prerequisites from test-all and test-unit (no subchart to fetch anymore)
skaffold.yaml - removed certmanager.install: false from the TWC setValues (the value no longer exists)

## Why?

Cert-manager manages cluster-wide resources that should be installed once per cluster, not bundled inside another chart's Helm release.

## Checklist

1. Closes #316


helm/temporal-worker-controller/values.yaml
```diff
@@ -159,20 +159,14 @@ workerResourceTemplate:
       resources: ["horizontalpodautoscalers"]
 
 certmanager:
-  # install controls whether cert-manager is installed as a Helm subchart.
-  # Set to true if cert-manager is not already installed in the cluster.
-  # See https://cert-manager.io/docs/installation/
-  install: false
```
tomba7, line 165: Since this field is going away, it would be worth updating the following sites where its still referenced: 

[README.md:80](https://github.com/temporalio/temporal-worker-controller/blob/a5de404da666795f6b54ce4bb57675a395b8e5c2/README.md#L80)
[docs/worker-resource-templates.md:94](https://github.com/temporalio/temporal-worker-controller/blob/a5de404da666795f6b54ce4bb57675a395b8e5c2/docs/worker-resource-templates.md#L94)
[internal/demo/README.md:13](https://github.com/temporalio/temporal-worker-controller/blob/a5de404da666795f6b54ce4bb57675a395b8e5c2/internal/demo/README.md#L13)

tomba7 (approved):
The rest LGTM

jaypipes (commented):
agree with @tomba7 about updating the various docs and let's add a tracking GH issue for this work please :)

helm/temporal-worker-controller/values.yaml
niyomukiza-mechack: Updated! I removed certmanager.install references from `README.md`, `docs/worker-resource-templates.md`, and `internal/demo/README.md`. Thanks for catching those.

jaypipes (approved):
muy bueno.

eniko-dif (approved):
The PR looks good to me, my agent found some cleanup nits:

 1. CI workflows still run the now-dead helm dependency build (inconsistent with the Makefile cleanup). Four workflow files still execute helm repo add jetstack https://charts.jetstack.io && helm dependency build helm/temporal-worker-controller:
 - .github/workflows/helm-validate.yml:22 and :37                                                                                                                                 
 - .github/workflows/helm-image-check.yml:32                                                                                                                                      
 - .github/workflows/test-integration.yml:145                                                                                                                                     
 - .github/workflows/release.yml:143 ("Build chart dependencies" step)                                                                                                            
                                                                                                                                                                                   
2. Stale comment in `release.yml:126`. The version-bump step still says:                                                                                                           
                                                                                                                                                                                  
 │ # Use ^version: to only match the top-level chart version, not indented dependency versions (e.g. cert-manager) which would otherwise be overwritten.                          
                                                                                                                                                                                  
 cert-manager is no longer a dependency, so the parenthetical example is misleading. Either drop the (e.g. cert-manager) example or simplify the comment.                         
                                                                                                                                                                                  
 3. Stale error message in `api/v1alpha1/webhook_suite_test.go:176–178`. The test renders the chart via helm template and on failure tells users:                                   
                                                                                                                                                                                  "helm template failed — is helm installed and the chart's dependencies built? Run via 'make test-unit' (which provisions ./bin/helm and the chart deps)..."                    
                                                                                                                                                                                  
There are no longer any chart deps to build, and make test-unit no longer builds them. Worth tightening the message so future failures don't send developers chasing a           
 non-existent dependency-build step.                                                                                                                                              
                                                                                                                                                                                  
 4. Vestigial charts/ leftovers for existing developers. helm/**/charts/*.tgz is gitignored, and anyone who previously ran make test-unit has a local                             
 helm/temporal-worker-controller/charts/cert-manager-v1.20.1.tgz. I verified this leftover causes helm lint (and helm lint --strict) to fail with:                                
                                                                                                                                                                                  
 ```                                                                                                                                                                              
   [ERROR] ... chart metadata is missing these dependencies: cert-manager                                                                                                         
 ```

PR #577, opened 2026-09-09
fix(controller): prune superseded inactive worker versions

### LINKED ISSUE #596: Inactive WorkerDeploymentVersions with no open pinned workflow executions never cleaned up
A rollout superseded before its version becomes Current or Ramping stays Inactive forever.

The controller scales its Deployment to zero, but the controller only cleans up the Temporal-side WorkerDeploymentVersion and the k8s-side Deployment for Drained and NotRegistered versions, not Inactive.

We should add cleanup for Inactive versions but only after checking with the CountWorkflow visibility query that there are no open pinned workflow executions targeting that specific build ID (which could be the case if a user executed the workflow with versioning override pinning to that build ID.

OP: @mitchross

mitchross:
A rollout superseded before its version becomes Current or Ramping stays Inactive forever. The controller scales its Deployment to zero, but only Drained and NotRegistered versions enter deletion, leaking both the Kubernetes Deployment and Temporal version record. This is the Inactive case discussed in #498.

Extend the existing controller deletion path:
- Nominate superseded Inactive versions only after Kubernetes has observed zero desired/actual replicas and no terminating replicas; protect current and target versions.
- Count running pinned workflows for the exact deployment version using both legacy dot and current colon encodings. Errors, missing responses, and nonzero counts retain the version.
- Use normal DeleteVersion without skipping drainage. Temporal refusal retains the Deployment for retry; NotFound permits Kubernetes cleanup.
- Retain rendered resources when an Inactive version cannot be deleted.

No CRD or new retirement clock is introduced. The existing sunset delays are defined for Drained versions. Unused Inactive versions are removed after the checks above succeed.

The documented operational limit is explicit: stop sending pinned overrides to versions being retired. Visibility is eventually consistent and the count is not atomic with deletion. Temporal's existing Drained status [also does not account for newly pinned overrides](https://typescript.temporal.io/api/interfaces/proto.temporal.api.deployment.v1.IWorkerDeploymentVersionInfo#drainageinfo). This change does not claim to provide a server-side exclusion against concurrent workflow starts or override updates.

Fixes #596 

Validation:
- `go test ./... -count=1` with envtest assets and Helm dependencies passes.
- Targeted integration tests against Kubernetes envtest and real Temporal 1.31.2 pass: superseded never-active version removal, pinned override protection followed by removal after closure, and normal Drained sunset.
- The same unused-version integration test fails on the unmodified controller after 90 seconds because its Deployment remains.
- Unit coverage includes observed/terminating replicas, current/target guards, visibility failures, server refusal, NotFound retry, and retained rendered resources.

The complete `TestIntegration` suite passes (475 seconds), covering Manual, AllAtOnce, Progressive, rollback, version limits, resource templates, and deletion. No production deployment has occurred.


jaypipes:
@mitchross 

> A rollout superseded before its version becomes Current or Ramping stays Inactive forever. The controller scales its Deployment to zero, but only Drained and NotRegistered versions enter deletion, leaking both the Kubernetes Deployment and Temporal version record. 

A WorkerDeploymentVersion that was superseded before its version became Current was marked as Inactive by Temporal Worker Controller before [this fix](https://github.com/temporalio/temporal-worker-controller/pull/554/changes) was committed in TWC v1.10.0/v1.7.1.

Would you mind installing TWC v1.10.0 and verifying that you no longer see the behaviour this PR addresses?

mitchross:
> @mitchross
> 
> > A rollout superseded before its version becomes Current or Ramping stays Inactive forever. The controller scales its Deployment to zero, but only Drained and NotRegistered versions enter deletion, leaking both the Kubernetes Deployment and Temporal version record.
> 
> A WorkerDeploymentVersion that was superseded before its version became Current was marked as Inactive by Temporal Worker Controller before [this fix](https://github.com/temporalio/temporal-worker-controller/pull/554/changes) was committed in TWC v1.10.0/v1.7.1.
> 
> Would you mind installing TWC v1.10.0 and verifying that you no longer see the behaviour this PR addresses?

Absolutely, will report back!

mitchross:
Thanks for taking a look, @jaypipes! I asked GPT Astra to run the follow-up tests and leave this reply on my behalf.

I'm completely fine with your decision on whether this needs an upstream change, and whether you want to use or adapt this patch. I just wanted to share what the retest found in case you're still curious.

The test used the official **v1.10.1 controller image** and **Temporal 1.32.0** in a separate local Kubernetes cluster with real worker Pods. It started a worker but left it waiting for permission to receive normal jobs (`rollout.strategy: Manual`), then replaced it with a second version and made that second version Current. The first version was never Current or Ramping and had no workflows assigned to it.

- The controller stopped the first version's Pods, but its Kubernetes Deployment and Temporal version record remained after seven minutes. Temporal itself reported it as Inactive.
- A separate version that had previously been Current cleaned up successfully.
- Restarting stock didn't remove the abandoned records. Changing only the test controller to the existing v1.10.1 backport of this patch removed those same records within 29 seconds. The replacement workers continued working.

Here's the standalone reproduction, sample worker, setup commands, and test details:
https://gist.github.com/mitchross/ec514b49662c1700fb438287f75ce25d

The original cluster test was run successfully; the Gist packaging hasn't had a separate run from scratch yet. It uses public dependencies and doesn't need access to my cluster or application repositories.

Could this be a separate case from the one #554 fixes? Either way, thanks for reviewing it!

docs/concepts.md
```diff
@@ -125,6 +125,17 @@ Defines how Drained versions are cleaned up:
...
+Versions superseded before ever becoming Current or Ramping remain Inactive in Temporal;
+they never acquire a drainage timestamp. The controller retires these versions after
+their Deployment has fully scaled to zero, visibility reports no running pinned workflows,
+and Temporal accepts the normal version deletion request. The drainage-based sunset delays
+do not apply to these unused versions. API failures or active pollers defer deletion and are
+retried on subsequent reconciliations.
+
+Stop sending pinned version overrides to a version being retired. Visibility is eventually
+consistent, so this check cannot exclude concurrent workflow starts or override changes.
+Temporal's drained status has the [same limitation for newly pinned overrides](https://typescript.temporal.io/api/interfaces/proto.temporal.api.deployment.v1.IWorkerDeploymentVersionInfo#drainageinfo).
```
jaypipes: Remove the above last paragraph. It's too much sausagemaking I'm afraid and will likely just confuse readers.

docs/concepts.md
```diff
@@ -125,6 +125,17 @@ Defines how Drained versions are cleaned up:
 - **scaledownDelay**: How long to wait after a version has been Drained before scaling pods to zero
 - **deleteDelay**: How long to wait after a version has been Drained before deleting the Kubernetes `Deployment`
 
+Versions superseded before ever becoming Current or Ramping remain Inactive in Temporal;
```
jaypipes: Let's convert this into a callout...
```suggestion
> **NOTE**: Versions superseded before ever becoming Current or Ramping remain Inactive in Temporal;
```

docs/concepts.md
```diff
@@ -125,6 +125,17 @@ Defines how Drained versions are cleaned up:
...
+they never acquire a drainage timestamp. The controller retires these versions after
+their Deployment has fully scaled to zero, visibility reports no running pinned workflows,
+and Temporal accepts the normal version deletion request. The drainage-based sunset delays
+do not apply to these unused versions. API failures or active pollers defer deletion and are
+retried on subsequent reconciliations.
```
jaypipes:
```suggestion
do not apply to these unused versions.
```

(removed the last line because it's just sausagemaking and not particularly important to convey to the reader)

internal/controller/execplan.go
```diff
@@ -710,6 +719,21 @@ func (r *WorkerDeploymentReconciler) deleteDrainedVersions(
...
+			return v.BuildID == buildID && v.Status == temporaliov1alpha1.VersionStatusInactive
+		}) {
+			retainedInactiveBuilds[buildID] = true
+			// Visibility can contain either the legacy dot separator or the newer colon form.
+			legacyVersion := strings.ReplaceAll(p.WorkerDeploymentName+"."+buildID, "'", "''")
+			version := strings.ReplaceAll(p.WorkerDeploymentName+":"+buildID, "'", "''")
+			count, err := temporalClient.CountWorkflow(ctx, &workflowservice.CountWorkflowExecutionsRequest{
+				Query: fmt.Sprintf("TemporalWorkerDeploymentVersion IN ('%s', '%s') AND TemporalWorkflowVersioningBehavior = 'Pinned' AND ExecutionStatus = 'Running'", legacyVersion, version),
+			})
+			if err != nil || count == nil || count.Count != 0 {
```
jaypipes: Please make a little helper function `getOpenPinnedWorkflowExecutions()` to encapsulate this query.

internal/controller/execplan.go
```diff
@@ -662,13 +663,19 @@ func (r *WorkerDeploymentReconciler) executeWRTOperations(
...
-// deleteDrainedVersions prunes the Temporal server-side Worker Deployment Version
+// deleteDeprecatedVersions prunes the Temporal server-side Worker Deployment Version
 // record for each k8s Deployment in DeleteDeployments, before executeK8sOperations
 // deletes them. It is mutated to remove the k8s Deployments that should not be
 // deleted because their Temporal server-side WDV record could not be removed or
 // because the k8s Deployment has no build ID label. The removed k8s deployments
 // stay in the cluster so that a later reconcile retries the pruning.
 //
+// DeleteVersion does not check pinned execution visibility for Inactive versions,
+// which can receive workflows through VersioningOverride. Check visibility first.
```
jaypipes:
```suggestion
// Because DeleteVersion does not check to see if there are open workflows using
// pinned execution for Inactive versions, we query the Temporal visibility service
// for open pinned workflows before deleting Inactive versions.
```

internal/controller/execplan.go
```diff
@@ -710,6 +719,21 @@ func (r *WorkerDeploymentReconciler) deleteDrainedVersions(
...
 		}
+		if slices.ContainsFunc(workerDeploy.Status.DeprecatedVersions, func(v *temporaliov1alpha1.DeprecatedWorkerDeploymentVersion) bool {
+			return v.BuildID == buildID && v.Status == temporaliov1alpha1.VersionStatusInactive
+		}) {
+			retainedInactiveBuilds[buildID] = true
+			// Visibility can contain either the legacy dot separator or the newer colon form.
+			legacyVersion := strings.ReplaceAll(p.WorkerDeploymentName+"."+buildID, "'", "''")
+			version := strings.ReplaceAll(p.WorkerDeploymentName+":"+buildID, "'", "''")
+			count, err := temporalClient.CountWorkflow(ctx, &workflowservice.CountWorkflowExecutionsRequest{
+				Query: fmt.Sprintf("TemporalWorkerDeploymentVersion IN ('%s', '%s') AND TemporalWorkflowVersioningBehavior = 'Pinned' AND ExecutionStatus = 'Running'", legacyVersion, version),
```
jaypipes: Might be a little easier to read if you create a `qs` variable first and pass that:

```go
qs := fmt.Sprintf(
    "TemporalWorkerDeploymentVersion IN ('%s', '%s') "+
        "AND TemporalWorkflowVersioningBehavior = 'Pinned' "+
        "AND ExecutionStatus = 'Running'",
    legacyVersion, version,
)
req := &workflowservice.CountWorkflowExecutionsRequest{Query: qs}
count, err := temporalClient.CountWorkflow(ctx, req)
...
```

internal/controller/execplan.go
```diff
@@ -710,6 +719,21 @@ func (r *WorkerDeploymentReconciler) deleteDrainedVersions(
...
+			})
+			if err != nil || count == nil || count.Count != 0 {
+				l.Info("could not confirm inactive version has no running pinned workflows, keeping its Deployment", "buildID", buildID, "error", err)
```
jaypipes, line 727: This info line might be deceiving in the case when `count.Count != 0`. In that case, we *have confirmed* that there are open pinned workflow executions against this Inactive version and so the info message should reflect that instead of saying "could not confirm..."

internal/controller/execplan.go
```diff
@@ -687,15 +694,17 @@ func (r *WorkerDeploymentReconciler) executeWRTOperations(
...
 	ctx context.Context,
 	l logr.Logger,
 	workerDeploy *temporaliov1alpha1.WorkerDeployment,
 	depHandle sdkclient.WorkerDeploymentHandle,
+	temporalClient sdkclient.Client,
 	p *plan,
 ) {
 	identity := getControllerIdentity()
 	markedForDeletion := make([]*appsv1.Deployment, 0, len(p.DeleteDeployments))
+	retainedInactiveBuilds := make(map[string]bool)
```
jaypipes: Unless I'm mistaken, you don't need the above `retainedInactiveBuilds`  variable. The `continue` on line 734 will cause the Deployment to not be added to the `markedForDeletion` slice on line 755...

jaypipes (changes requested):
@mitchross thanks very much for this PR! Left a few inline suggestions for you, but nothing major :)

jaypipes:
@mitchross hey, just FYI, the TWC maintainers all hang out on the `#temporal-workers` channel in the Temporal Slack community. If you'd like a higher-bandwidth conversation about any of your PRs or issues, feel free to hop over there :)

mitchross:
> @mitchross hey, just FYI, the TWC maintainers all hang out on the `#temporal-workers` channel in the Temporal Slack community. If you'd like a higher-bandwidth conversation about any of your PRs or issues, feel free to hop over there :)

Thanks for the heads up. Getting some practice in at home as we start to ramp up Temporal in a cluster at $JOB. 

I think I hit on all the comments. In full transparency I leveraged AI, GPT 6. I purposely used the temporal skills , and temporal MCP docs server to ensure adherence.

internal/controller/execplan.go
```diff
@@ -710,6 +722,20 @@ func (r *WorkerDeploymentReconciler) deleteDrainedVersions(
 			markedForDeletion = append(markedForDeletion, d)
 			continue
 		}
+		if slices.ContainsFunc(workerDeploy.Status.DeprecatedVersions, func(v *temporaliov1alpha1.DeprecatedWorkerDeploymentVersion) bool {
+			return v.BuildID == buildID && v.Status == temporaliov1alpha1.VersionStatusInactive
+		}) {
+			retainedResourceBuilds[buildID] = true
```
jaypipes: Not sure if you caught my earlier comment about this, but I don't believe you need the `retainedResourceBuilds` variable at all. Pretty sure the existing `markedForDeletion` variable and behaviour is all that is needed.

jaypipes:
> > @mitchross hey, just FYI, the TWC maintainers all hang out on the `#temporal-workers` channel in the Temporal Slack community. If you'd like a higher-bandwidth conversation about any of your PRs or issues, feel free to hop over there :)
> 
> Thanks for the heads up. Getting some practice in at home as we start to ramp up Temporal in a cluster at $JOB.

Nice! :)

> I think I hit on all the comments. In full transparency I leveraged AI, GPT 6. I purposely used the temporal skills , and temporal MCP docs server to ensure adherence.

Totally cool. Thanks for the transparency, it is much appreciated!

internal/controller/execplan.go
```diff
@@ -755,15 +751,27 @@ func (r *WorkerDeploymentReconciler) deleteDeprecatedVersions(
...
+		}
+		if !slices.ContainsFunc(workerDeploy.Status.DeprecatedVersions, func(v *temporaliov1alpha1.DeprecatedWorkerDeploymentVersion) bool {
+			return v.BuildID == buildID && v.Status == temporaliov1alpha1.VersionStatusInactive
+		}) {
+			continue
+		}
+		p.DeleteWorkerResources = slices.DeleteFunc(p.DeleteWorkerResources, func(ref planner.WorkerResourceRef) bool {
+			return ref.BuildID == buildID
+		})
 	}
```
jaypipes: @mitchross I think maybe Claude is getting confused :) I don't believe this block of code is needed. Inactive build IDs that have open workflows are not added to the `markedForDeletion` slice (line 753) because of the `continue` statements on lines 728 and 732.

jaypipes (approved):
muy bueno, thank you so much @mitchross! :)

PR #582, opened 2026-09-09
Test: add E2E configuration matrix workflow for TWC

niyomukiza-mechack:
## What was changed

Added a new E2E configuration matrix workflow that deploys TWC to a Kind cluster with different Helm value configurations and verifies the controller actually starts and works.

Currently tests three modes: default (cluster-wide with cert-manager), namespace-scoped (restrictWatchNamespaces), and BYO cert (certmanager.enabled=false with a custom certSecretName).
Each mode verifies that controller pods reach Running state, there are no error-level log entries, and the webhook accepts a test WorkerDeployment.

This would have caught both the namespace-scoped ClusterConnection watch bug  and the hard-coded cert Secret name issue  before they shipped. 

The goal is to expand the matrix over time as we discover new customer deployment configurations - each new mode is one entry in the matrix, so adding coverage is lightweight.

## Checklist

Related to #584

1. Closes 

2. How was this tested:

3. Any docs updates needed?


.github/workflows/e2e-config.yml
```diff
@@ -0,0 +1,163 @@
...
+jobs:
+  e2e:
+    name: e2e-${{ matrix.mode.name }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    strategy:
+      fail-fast: false
+      matrix:
+        mode:
+          - name: default
```
zainawaisn, line 24: `webhook.enabled` is set to [false by default in values.yaml](https://github.com/temporalio/temporal-worker-controller/blob/84d717c31fd1e8899515a104fe63c791c41046ff/helm/temporal-worker-controller/values.yaml#L139) for WorkerDeployment webhooks.

You can either add a "--set webhook.enabled=true" arg to each individual helm-arg or add "--set webhook.enabled=true" to the shared `name: Install TWC` step.

```
- name: Install TWC
        run: |
          helm install temporal-worker-controller helm/temporal-worker-controller \
            --namespace temporal-system \
            --set image.repository=temporal-worker-controller \
            --set image.tag=e2e \
            --set image.pullPolicy=IfNotPresent \
            --set webhook.enabled=true \
            ${{ matrix.mode.helm-args }}
```

zainawaisn: As a follow-up to this, the byo-cert mode does not set `certmanager.caBundle`. When the webhooks are enabled and `cert-manager.installed`=false, byo-cert will need to assign the cert from "Generate BYO certificate" step to `certmanager.caBundle` in order to validate the self-signed cert.

niyomukiza-mechack: I agree but this was done on purpose. Getting this workflow green in CI took about 10 rounds of fixes due to the subchart dependency still being in Chart.yaml. Every change that touched the cert setup or install order broke something new, so I simplified things to get to green. No webhook.enabled=true, no caBundle, no custom cert Secret name. The plan was/is once the subchart removal https://github.com/temporalio/temporal-worker-controller/pull/575 merges, I'll rebase and enable everything. The webhook.enabled=true, caBundle for BYO, custom certSecretName, and remove the dependency build step. Sorry I should have mentioned this in the PR description

zainawaisn: No worries, this sounds good then. These items can be done as a follow-up after #575.

zainawaisn (approved):

PR #587, opened 2026-09-14
fix(controller): scale draining versions back up from 0 replicas

### LINKED ISSUE #573: [Bug] Draining versions scaled to 0 during unspecified-drainage window never scale back up

Keep pollers on Worker Deployment Versions that Temporal still reports as **Draining**, so PINNED workflow/activity tasks on those builds can complete and the version can reach Drained.

Sunset docs and the CRD comments say `scaledownDelay` / `deleteDelay` start only after DRAINED. That part is working. The failure is earlier: some draining versions already have 0 replicas, so they can never drain.

### Describe the bug

Two interacting behaviors in the planner:

1. **Unspecified drainage is treated as Inactive and scaled to 0 immediately.**
   After a version leaves Current/Ramping, Temporal often still has `DrainageStatus=UNSPECIFIED` until the drainage checker runs. `GetWorkerDeploymentState` maps that to `VersionStatusInactive`:

   https://github.com/temporalio/temporal-worker-controller/blob/v1.8.0/internal/temporal/worker_deployment.go

   ```go
   } else if drainageStatus == enumspb.VERSION_DRAINAGE_STATUS_DRAINING {
       versionInfo.Status = temporaliov1alpha1.VersionStatusDraining
   } else if drainageStatus == enumspb.VERSION_DRAINAGE_STATUS_DRAINED {
       versionInfo.Status = temporaliov1alpha1.VersionStatusDrained
       // ...
   } else {
       versionInfo.Status = temporaliov1alpha1.VersionStatusInactive
   }
   ```

   `getScaleDeployments` then scales non-target Inactive versions to 0 on the same reconcile:

   https://github.com/temporalio/temporal-worker-controller/blob/v1.8.0/internal/planner/planner.go

   ```go
   case temporaliov1alpha1.VersionStatusInactive:
       // Scale down inactive versions that are not the target
       // ...
       scaleDeployments[version.Deployment] = 0
   ```

2. **Draining has no scale case, so a version zeroed in that window never comes back.**
   Same function: Current/Target are scaled to `spec.replicas`, Drained is scaled to 0 after `scaledownDelay`, Inactive (non-target) is scaled to 0. There is no `VersionStatusDraining` arm. Whatever replica count the Deployment already has is left alone.

   So if the first post-Current reconcile saw unspecified drainage, replicas stay 0 even after Temporal later reports `DRAINING` and the CR status is `Draining`.

   PINNED tasks then `SCHEDULE_TO_START` / heartbeat-timeout with no pollers. The version never reaches Drained → sunset cannot delete it → drain deadlock. It also occupies a Worker Deployment version slot (we hit the 100-version cap with several of these stuck).

This is still the case on `main` as of 2026-09-08, not only v1.8.0.

Raising `scaledownDelay` does not help: these versions never entered the Drained scaledown path.

### Minimal Reproduction

Manual rollout, PINNED workflows, `spec.replicas` set (controller-managed, not HPA).

1. Version A is Current, with at least one PINNED workflow running (or sitting in a long `workflow.Sleep` / `condition` wait).
2. Promote version B to Current.
3. If the controller reconciles A while Temporal still has `DrainageStatus=UNSPECIFIED`, A is classified Inactive and scaled to 0.
4. Subsequent reconciles show A as Draining (Temporal drainage checker has now run). Replicas stay 0.
5. When the PINNED workflow needs a WFT or activity retry, nothing polls that version.

Contrast: if the first post-Current reconcile already sees `DRAINING`, replicas stay at `spec.replicas` and the version eventually drains. Same cluster, same `sunset.scaledownDelay: 1h`, both outcomes at once.

### Environment/Versions

- Controller: `temporalio/temporal-worker-controller:v1.8.0` (bug still present on `main`)
- Kubernetes (EKS)
- Temporal Cloud
- `WorkerDeployment` rollout strategy: `Manual`
- `spec.replicas: 2`
- `sunset.scaledownDelay: 1h`, `deleteDelay: 1h`
- Workflows: `versioningBehavior: PINNED`

### Additional context

Production snapshot (one WorkerDeployment, Temporal Cloud):

| Version | Temporal drainage | k8s replicas | PINNED Running workflows |
| --- | --- | --- | --- |
| A (left Current ~10 min after start) | draining, 5d | **0** | 4 |
| B (left Current ~12 min after start) | draining, 3d | **0** | 1 |
| C (left Current after days as Current) | draining | **2** | 8 |
| D (draining ~1h) | draining | **2** | 3 |

C/D show sunset-after-DRAINED is doing the right thing (keep replicas while Draining). A/B lost pollers minutes after leaving Current, which is the Inactive window, not the 1h scaledown clock.

Suggested fix (either is enough to stop the deadlock; both is better):

- Do not map unspecified drainage of a non-current/non-ramping version to Inactive. Treat it as Draining (or "drainage unknown — do not scale down") until Temporal reports DRAINING or DRAINED.
- Add a `VersionStatusDraining` arm that scales to `spec.replicas` (or at least ≥1), so a version zeroed in the race recovers.

Workaround that works today: `kubectl scale` the 0-replica Draining Deployment back up. Because Draining is a no-op in `getScaleDeployments`, the controller will not scale it back down. Not a substitute for the fix.

tomba7:
## What was changed

- getScaleDeployments currently handles Inactive, Ramping, Current and Drained, but leaves out Draining. This was exposed by issue 573.
- Added a case that handles draining versions and scales them back up from zero (to spec.Replicas when the controller manages replicas and to 1 when a scaler does)

## Why?

- A draining version at 0 replicas can never recover, whether it got there from a bug or via someone scaling it by hand. Since there are no pollers, the draining version never drains.

## Checklist

1. Closes #573 (item 2)

2. How was this tested:

   Unit tests:
   - Draining at 0, controller-managed: scaled back up to spec.Replicas.
   - Draining at 0, scaler-managed: scaled up to 1.
   - Draining at a non-zero count: left untouched.
   - Draining at 0 while the whole worker deployment is scaled to zero: left untouched.

   Manual testing on a kind cluster plus a Temporal Cloud namespace:
   - Reproduced the stuck state by promoting a new version with pinned
     workflows still open on the old one, then scaling the old version's
     Deployment to 0 with kubectl.
   - Without the fix the version stayed at 0 replicas and never drained. With
     the fix the controller scaled it back up within seconds and drainage
     resumed.


internal/planner/planner.go
```diff
@@ -856,6 +857,27 @@ func getScaleDeployments(
 					scaleDeployments[version.Deployment] = uint32(replicas)
 				}
 			}
+		case temporaliov1alpha1.VersionStatusDraining:
+			// A draining version with 0 replicas has no pollers, so it will never finish
+			// draining and will never be retired. We detect and repair this scenario
+			// here. Any other non-zero count still has pollers and will drain on its own.
```
jaypipes, line 863: I recognize that a draining version with 0 replicas has no pollers. This is absolutely true and what we handled in @carlydf 's [commit](https://github.com/temporalio/temporal-worker-controller/commit/3fdfa93d9d30dbfb83d523f30be24378cf442c3a). What I'm wondering, though, is after Carly's fix, how would we get into this situation here where a Draining WDV would have its Deployment.spec.replicas set to 0?

jaypipes (commented):
I'm fine with this code -- it certainly doesn't hurt anything. However, I question whether this code will ever actually execute since Carly's fix landed in TWC 1.10.0/1.7.1. I can't think of a way that the controller would ever set `Deployment.spec.replicas` to `0` for a Draining WDV after Carly's fix landed... @tomba7 thoughts?

internal/planner/planner.go
tomba7: @jaypipes Totally agree. However, I do think that any cluster that was on an older version prior to Carly's fix and had a WDV zeroed out will still have its k8s Deployment at 0 replicas, and nothing today would recover it. And if we end up here again due to another bug in the future, this provides a path to recovery.

jaypipes: @tomba7 ack. good point on the older TWC versions potentially leaving a k8s Deployment at 0 replicas...

jaypipes (approved):

PR #589, opened 2026-09-15
Add flags to configure webhook TLS cert directory/filenames

shivangi660:
## Summary
- `controller-runtime` defaults the webhook server's TLS cert location to `/tmp/k8s-webhook-server/serving-certs/{tls.crt,tls.key}`, hardcoded via the manager's default `webhook.Options` whenever `ctrl.Options{}` doesn't set `WebhookServer` explicitly (as `cmd/main.go` currently does not).
- Platforms that provision workload TLS material via an in-cluster PKI agent writing to a different fixed path/filename (e.g. a node-level identity provisioner) have no way to point this controller at that location without forking.
- Adds three flags (with env var fallback for the directory, mirroring the existing `--watch-namespaces`/`WATCH_NAMESPACES` pattern):
  - `--webhook-cert-dir` (env `WEBHOOK_CERT_DIR`, default: unset)
  - `--webhook-cert-name` (default: `tls.crt`)
  - `--webhook-key-name` (default: `tls.key`)
- When `--webhook-cert-dir`/`WEBHOOK_CERT_DIR` is unset, behavior is unchanged — controller-runtime's default path is used, so this is backward compatible.

## Test plan
- [x] `go build ./...` passes
- [x] `go vet ./cmd/...` passes


shivangi660:
Superseded by #590 — same change, rebased onto current main with signed commits.

Issue #542, opened 2026-08-20
flaky integration test: TestIntegration/7th-rollout-unblocked-after-pollers-die-version-deleted


jaypipes:
```
    --- FAIL: TestIntegration/7th-rollout-unblocked-after-pollers-die-version-deleted (32.83s)
```

We have a flaky integration test that periodically fails. A sampling of log records from the test output (de-duplicated):

```
2026-08-19T22:36:28Z	DEBUG	Running Reconcile loop	{"controller": "workerdeployment", "controllerGroup": "temporal.io", "controllerKind": "WorkerDeployment", "WorkerDeployment": {"name":"conditions-temporal-state-fetch-failed","namespace":"test-integration-20260819222923"}, "namespace": "test-integration-20260819222923", "name": "conditions-temporal-state-fetch-failed", "reconcileID": "e41b070e-2cfd-49cd-992b-aebfa64596cb"}
2026-08-19T22:36:28Z	DEBUG	events	Unable to get worker deployment state: unable to describe worker deployment test-integration-20260819222923/conditions-temporal-state-fetch-failed: Namespace does-not-exist is not found.	{"type": "Warning", "object": {"kind":"WorkerDeployment","namespace":"test-integration-20260819222923","name":"conditions-temporal-state-fetch-failed","uid":"e28c5c8f-baf1-4fa1-8b5c-dc306c8cac94","apiVersion":"temporal.io/v1alpha1","resourceVersion":"945"}, "reason": "TemporalStateFetchFailed"}
2026-08-19T22:36:28Z	ERROR	unable to create TemporalClient	{"controller": "workerdeployment", "controllerGroup": "temporal.io", "controllerKind": "WorkerDeployment", "WorkerDeployment": {"name":"conditions-client-creation-failed","namespace":"test-integration-20260819222923"}, "namespace": "test-integration-20260819222923", "name": "conditions-client-creation-failed", "reconcileID": "0f3a27cd-52ad-4d33-bb3c-4e04ecd32a74", "error": "failed reaching server: connection error: desc = \"transport: Error while dialing: dial tcp 127.0.0.1:1: connect: connection refused\""}
github.com/temporalio/temporal-worker-controller/internal/controller.(*WorkerDeploymentReconciler).Reconcile
	/home/runner/work/temporal-worker-controller/temporal-worker-controller/internal/controller/worker_controller.go:282
[...]

PR #590, opened 2026-09-15
Add flags to configure webhook TLS cert directory/filenames

### LINKED ISSUE #593: Manager crash-loops when webhook TLS cert is injected outside the default controller-runtime cert dir
## Problem

The manager's webhook server always used controller-runtime's hardcoded default cert directory (`/tmp/k8s-webhook-server/serving-certs`), with no flag/env override. Deployments that provision the webhook TLS cert via a mechanism other than the chart's own cert-manager `Issuer`/`Certificate` (e.g. a platform-managed PKI sidecar/init container that writes cert+key to some other pod path, then set `certmanager.enabled: false`) have no way to point the webhook server at where their cert actually landed. The webhook server then fails to load a cert at the default path on startup, and the manager pod crash-loops.

This is what `--webhook-cert-dir`/`--webhook-cert-name`/`--webhook-key-name` (see #590) fixes — makes the cert dir/filenames configurable via flag or `WEBHOOK_CERT_DIR` env, instead of hardcoded.

## Distinct from #591

#591 tracks a separate, still-open concern once #590 lands: even with a configurable cert dir, the `ValidatingWebhookConfiguration`'s `caBundle` still needs to be populated with the CA that signed that externally-injected cert (currently only automatic when `certmanager.enabled: true`; static `caBundle` value goes stale on rotation). #590 fixes the pod crash-loop; #591 is about the API server trusting the webhook's TLS cert once the pod is actually running.

## Fix

#590

shivangi660:
## Summary
controller-runtime defaults the webhook server to `/tmp/k8s-webhook-server/serving-certs/{tls.crt,tls.key}`, with no way to override it. Platforms that provision workload TLS material at a different path/filename (e.g. an in-cluster PKI agent that writes certs under a fixed non-default path) can't point this controller at their cert location without forking.

- Adds `--webhook-cert-dir` (env `WEBHOOK_CERT_DIR`), `--webhook-cert-name`, and `--webhook-key-name` flags.
- When `webhook-cert-dir` is unset, behavior is unchanged (controller-runtime's default path is used).

Fixes #593


jaypipes:
@shivangi660 is there a GH tracking issue that this PR addresses? I note that #591 is a separate concern to this, right?

shivangi660:
@jaypipes there wasn't one yet — opened #593 to track the crash-loop this PR fixes (manager hard-crashes on startup when the webhook TLS cert is injected outside controller-runtime's default `/tmp/k8s-webhook-server/serving-certs` path and `certmanager.enabled: false`).

And yes, #591 is separate: even once this PR lands, the `ValidatingWebhookConfiguration`'s `caBundle` still needs the CA that signed that externally-injected cert, which isn't wired up outside `certmanager.enabled: true`. #590 gets the pod running; #591 is about the API server trusting its webhook once it's up. Working on a follow-up PR for #591 now (CA-injector Job, similar to how ingress-nginx's admission webhook patches its own caBundle).

shivangi660:
Integration test failure is unrelated to this PR's diff (`cmd/main.go` only, adds webhook cert-dir flags). It's `TestIntegration/6th-rollout-unblocked-after-pollers-die-max-ctrlr-versions` failing on:

```
validation_helpers.go:493: eventually failed after 30s: expected deprecated version did not match actual: expected Deployment for build 'v0-dd8f' to have 0 replicas, but had 1
```

Same class of flake as #542 (grpc `connection refused`/`EOF` noise from the test's temporal-test-server during teardown, causing a reconcile to miss its 30s window). I don't have permission to re-run the workflow (not a repo admin) — could someone retrigger it?

jaypipes (approved):
👍

shivangi660:
> 👍

@jaypipes can you help with the merge as well

jaypipes:
> > 👍
> 
> @jaypipes can you help with the merge as well

@shivangi660 we are in the process of cutting the v1.11.0 release at the moment. Holding off merging this until after that is cut.

jaypipes:
Well, actually, this is a bug fix technically and isn't dangerous, so might as well get it into v1.11.0...

Issue #246, opened 2026-03-26
[Feature Request] Temporal Worker Controller as a library


eniko-dif:
### Is your feature request related to a problem? Please describe.

Companies operating a self-hosted Temporal server cluster (such as Datadog) often require more advanced configuration of the client used to connect to the server than what is currently supported by the TemporalConnection CRD. In addition to connection-level customization, there may be further setup or configuration needed to successfully run and integrate the controller within their environment.

### Describe the solution you'd like

To address this, the proposal is to expose a public interface that allows creating the controller with custom options. This would make it easier to configure connection details beyond what the CRD supports. It could also optionally allow providing a pre-configured Temporal client, giving full control over how the controller connects to the server.

However, the change should not impact the current behavior of the controller in any way, such that for Temporal Cloud users it works directly out-of-the-box.

PR #597, opened 2026-09-16
[Issue 246]: Introduce ConnectionProvider seam (step 1: controller-as-library)

eniko-dif:
## Summary

Related to issue https://github.com/temporalio/temporal-worker-controller/issues/246

This is **step 1** of making the worker controller usable as a library: it introduces a clean seam between the `WorkerDeployment` reconciler and the connection resources it references, so the OSS controller no longer hardcodes the `temporal.io/Connection` and `temporal.io/ClusterConnection` kinds.

A wrapper binary can now register additional connection kinds at process start — without upstream CRD changes — by appending its own `ConnectionProvider` to the startup-built slice. The reconciler resolves a `ConnectionReference` to a registered provider, which supplies the referenced object (for finalizer management), a cached SDK client, and a fingerprint (for drift detection).

### New seam: `internal/controller/connectionprovider`
- **`ConnectionProvider`** — the group-kind-level factory. One instance is registered per `GroupKind` at startup; the default registry ships `Connection` and `ClusterConnection`. Provides `Fetch`, `NewObject`, `IsClusterScoped`, and `GroupKind`.
- **`ResolvedConnection`** — the per-reconcile handle bound to a fetched object. Carries the behavior for that connection: `Object()`, `GetClient()`, `Fingerprint()`, `ApplyWorkerPodSpec()`, `Evict()`.
- **Error taxonomy** mapping to distinct status conditions: `*UnknownKindError` → `ReasonUnknownConnectionKind`, `*AuthError` → `ReasonAuthSecretInvalid`, `*DialError` → `ReasonTemporalClientCreationFailed`.
- **Identity by name + GroupKind** (not hardcoded kind names), applied consistently across the watch mapper, finalizer release, and `sameConnectionRef` — so a namespaced `Connection` \"foo\" and a cluster-scoped connection \"foo\" are distinct.

### Default provider: `internal/controller/clientpool/default_provider.go`
- `DefaultProvider` implements `ConnectionProvider` for the two built-in kinds, sharing a `ClientPool`.
- `defaultResolved` implements `ResolvedConnection`; the `ConnectionSpec` is opaque to the reconciler and never leaves this package.

### Reconciler wiring
- `WorkerDeploymentReconciler` now holds a startup-built `Providers []ConnectionProvider` (never mutated) instead of a concrete `ClientPool`. `main.go` wires `clientpool.NewDefaultProviders`.
- Watches are set up per registered kind (skipping cluster-scoped kinds when the controller is namespace-scoped), so wrapper-registered kinds get the watch for free.
- `connection_helpers.go` centralizes finalizer add/release logic, scoped correctly per kind.

### API
- `ConnectionReference` gains an `ObjectRef` (full type info: apiGroup + kind + name) alongside the shorthand `Name` form. CEL validation enforces exactly-one-of and `apiGroup == temporal.io`; kind is intentionally **not** validated at admission — an unregistered kind surfaces as `ReasonUnknownConnectionKind` on the next reconcile.

## Design notes

- **`Fingerprint` / `ApplyWorkerPodSpec` invariant:** `ApplyWorkerPodSpec` writes `Fingerprint()` into the `connection-spec-hash` annotation, so the two cannot drift apart — the planner compares `Fingerprint()` against that annotation to decide whether an in-place pod-spec update is needed. The default provider guarantees this by delegating to `Fingerprint()`; the contract is documented on both methods and locked by a test.
- **`PodSpecApplyOpts`:** the per-worker context (`TemporalNamespace`, `WorkerDeploymentName`, `BuildID`) is passed as a struct rather than positional `string`s, removing the risk of swapping indistinguishable arguments at call sites.
- Unknown kinds are rejected **before** the WD finalizer is added, so a bad `connectionRef` never wedges deletion.

## Why \"step 1\"

The concrete `ConnectionSpec` still lives in `api/v1alpha1` and the default provider is the only one shipped. This PR establishes the interface and the registration mechanism; subsequent steps will let a wrapper binary supply its own provider implementations and (eventually) its own connection CRDs without forking the controller.

## Test coverage

- `connectionprovider`: `LookupProvider` (unknown/found/empty), `RefGroupKind`, `RefName`, `UnknownKindError.Is`.
- `clientpool` (default provider): `Fetch` (namespaced/cluster-scoped/not-found), `GetClient` (auth/dial/cache-hit), `Fingerprint` stability, `ApplyWorkerPodSpec` (inject + idempotent + annotation-equals-fingerprint), `Evict`, `NewObject`, `IsClusterScoped`, `NewDefaultProviders`.
- Existing reconciler/planner/k8s tests updated for the new signatures.

`go build ./...`, `go vet`, and `gofmt` are clean; affected package tests pass.

## Checklist

- [x] Builds with `go build ./...`
- [x] `gofmt` clean
- [x] Tests added/updated for new behavior


internal/controller/connectionprovider/connectionprovider.go
```diff
@@ -0,0 +1,158 @@
...
+	Fetch(ctx context.Context, ref temporaliov1alpha1.ConnectionReference, k8sNamespace string) (ResolvedConnection, error)
+
+	// NewObject returns a zero-valued object of this kind, for setting up a
+	// watch in SetupWithManager.
+	NewObject() client.Object
+
+	// IsClusterScoped reports whether this kind is cluster-scoped. Used to
+	// scope the watch mapper: a namespaced kind lists WorkerDeployments in the
+	// connection's own namespace; a cluster-scoped kind lists across all.
+	IsClusterScoped() bool
```
jaypipes, line 46: I don't believe this interface method is necessary. It's possible to ask the discovery client whether a GroupKind is cluster-scoped or not:

```go
func kindIsNamespaced(
	disco discovery.CachedDiscoveryInterface, // See: https://github.com/kubernetes/client-go/blob/master/discovery/cached/disk/cached_discovery.go#L45
	gvr schema.GroupVersionResource, // See: https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime/schema#GroupVersionResource
) bool {
	apiResources, err := disco.ServerResourcesForGroupVersion(
		gvr.GroupVersion().String(),
	)
	if err != nil {
		return false
	}
	for _, apiResource := range apiResources.APIResources {
		if apiResource.Name == gvr.Resource {
			return apiResource.Namespaced
		}
	}
	return false
}
```

internal/controller/connectionprovider/connectionprovider.go
```diff
@@ -0,0 +1,158 @@
...
+
+	// ApplyWorkerPodSpec mutates the worker pod template for this kind and
+	// records Fingerprint() in the connection-spec-hash annotation. Must be
+	// idempotent so it serves both fresh-pod creation and in-place drift
+	// updates. The annotation MUST equal Fingerprint() so drift detection
+	// (which compares the two) is consistent; the default provider guarantees
+	// this by delegating to Fingerprint(). A provider whose kind owns pod config
+	// out of band implements the mutation as a no-op but must still write the
+	// fingerprint annotation.
+	ApplyWorkerPodSpec(podSpec *corev1.PodSpec, annotations map[string]string, opts PodSpecApplyOpts) error
```
jaypipes, line 77: Do we need to pass the `annotations` map? How about just passing the fingerprint that we will set on the PodSpec?

internal/controller/clusterconnection_test.go
```diff
@@ -20,6 +21,35 @@ import (
...
+	prov, err := connectionprovider.LookupProvider(r.Providers, ref)
+	if err != nil {
+		return nil, err
+	}
+	connection, err := prov.Fetch(ctx, ref, namespace)
+	if err != nil {
+		return nil, err
+	}
+	return connection.Object(), nil
+}
```
jaypipes, line 51: Any reason these methods are defined on WorkerDeploymentReconciler in the `clusterconnection_test.go` file instead of a non-test file like the newly-added `internal/controller/connection_helpers.go` file?

internal/controller/worker_controller.go
```diff
@@ -1066,57 +911,40 @@ func (r *WorkerDeploymentReconciler) reconcileRequestForWRT(ctx context.Context,
...
-		ref := twd.Spec.WorkerOptions.ConnectionRef
-		// Only namespaced Connection refs are driven by this (Connection) watch.
-		if !connectionRefIsCluster(ref) && connectionRefName(ref) == tc.GetName() {
-			requests = append(requests, reconcile.Request{
-				NamespacedName: types.NamespacedName{
-					Name:      twd.Name,
-					Namespace: twd.Namespace,
-				},
-			})
+// findTWDsUsingConnectionKind returns a mapper that enqueues every WorkerDeployment
```
jaypipes, line 914: This should probably be moved to the new `internal/controller/connection_helpers.go` file along with the other connection-related helpers.

internal/tests/go.mod
```diff
@@ -206,12 +206,14 @@ require (
...
 	modernc.org/memory v1.11.0 // indirect
-	modernc.org/sqlite v1.49.1 // indirect
+	modernc.org/sqlite v1.51.0 // indirect
 	sigs.k8s.io/json v0.0.0-20250730193827-2d320260d730 // indirect
 	sigs.k8s.io/randfill v1.0.0 // indirect
 	sigs.k8s.io/structured-merge-diff/v6 v6.4.0 // indirect
 	sigs.k8s.io/yaml v1.6.0 // indirect
 )
+
+replace github.com/temporalio/temporal-worker-controller => ../..
```
jaypipes, line 219: oops :)

internal/planner/planner.go
```diff
@@ -412,172 +415,65 @@ func getDeleteWorkerResources(
...
-	return mounts
-}
-
 // checkAndUpdateDeploymentPodTemplateSpec determines whether the Deployment for the given buildID is
 // out-of-date with respect to the user-provided pod template spec. This enables rolling updates when
 // the build ID is stable (e.g., using spec.workerOptions.buildID) but the pod spec has changed.
 // If an update is required, it rebuilds the deployment spec and returns a pointer to that Deployment.
 // If no update is needed or the Deployment does not exist, it returns nil.
 func checkAndUpdateDeploymentPodTemplateSpec(
+	ctx context.Context,
```
jaypipes, line 471: Any reason to add the context parameter here? I don't see it being used anywhere.

internal/controller/genstatus.go
```diff
@@ -72,7 +72,7 @@ func (r *WorkerDeploymentReconciler) generateStatus(
 			if prevStatusByWorkflowID[wf.WorkflowID] == wf.Status {
 				continue
 			}
-			r.Recorder.Eventf(workerDeploy, corev1.EventTypeWarning, ReasonGateWorkflowFailed,
+			r.Recorder.Eventf(workerDeploy, nil, corev1.EventTypeWarning, ReasonGateWorkflowFailed, ReasonGateWorkflowFailed,
```
jaypipes, line 75: Hmmm.... I'm wondering how this code was previously compiling. It seems we were not calling the `EventRecorder.Eventf` [interface method](https://github.com/kubernetes/client-go/blob/28076445520055420e3be4255b4cd27fd19df1f9/tools/internal/events/interfaces.go#L43) properly... there should have been a compile-time failure.

jaypipes (commented):
Phew, this is a lot to digest :) Some suggestions inline for you to take a look at. Will do a more in-depth review later on.

PR #606, opened 2026-09-18
fix(controller): back off failing drained version deletes per build ID

### LINKED ISSUE #583: [bug] deleteDrainedVersions retries a permanently-failing DeleteVersion forever, and its error bypasses the controller's own backoff
**Version:** `temporalio/temporal-worker-controller:v1.10.1` (chart `oci://docker.io/temporalio` 0.29.1), against Temporal server 1.31.2. Code references against `main` @ `a5de404`.

## Summary

When `DeleteWorkerDeploymentVersion` fails with anything other than `NotFound`, the controller keeps the version nominated for deletion and re-attempts on the next reconcile — with no attempt counter, no escalating interval, and no give-up. If the failure is permanent, this becomes an unbounded error source for as long as the CR exists.

Measured on one cluster: **577 failed attempts in 30 minutes** (19.2/min) across 6 WorkerDeployments, sustained for **17+ hours** without decay.

## Reproduction / observed state

Six WorkerDeployments, each stuck on exactly one drained build ID. Log line, once per deployment per reconcile:

```json
{"level":"info","msg":"could not delete worker deployment version, keeping its k8s Deployment to reconcile",
 "WorkerDeployment":{"name":"worker-a","namespace":"default"},
 "buildID":"0.738.0-aaaa","deployment":"worker-a-0-738-0-aaaa",
 "error":"something went wrong, please retry (36634b00)"}
```

Per-deployment counts over 30 min: 98, 97, 96, 96, 96, 95 — i.e. one attempt every ~19s each.

The masked UUID resolves in the server frontend log to:

```
msg        masked service failures
error      rpc error: code = Internal desc = activity error
operation  DeleteWorkerDeploymentVersion
logging-call-at  common/rpc/interceptor/mask_internal_error.go:111
```

The underlying reason the delete can never succeed is a server-side inconsistency: the version's own system workflow has **already completed its self-delete**, while the parent deployment workflow still lists the version as `drained`.

```
temporal worker deployment describe --name default/worker-a
  Version Summaries:
    0.745.0-bbbb  unspecified   4 hours ago
    0.738.0-aaaa  drained      17 hours ago   <-- controller retries this forever

temporal workflow describe \
  --workflow-id "temporal-sys-worker-deployment-version:default/worker-a:0.738.0-aaaa"
  Status                COMPLETED
  CloseTime             17 hours ago
  TemporalChangeVersion [..., "serialDelete-1", "delete-wci-1"]
```

All 6 are identical: version workflow `COMPLETED` carrying `serialDelete-1`, closed 15–17h ago. The parent deployment workflows are themselves **healthy** (7–12 history events, i.e. normally continuing-as-new), so this is not a wedged parent or an in-flight-update ceiling.

I appreciate that the inconsistency itself is a server bug and I'll raise it separately. This issue is about the controller's response to it.

## The controller-side problem

`internal/controller/execplan.go`, `deleteDrainedVersions`:

```go
_, err := depHandle.DeleteVersion(ctx, sdkclient.WorkerDeploymentDeleteVersionOptions{
    BuildID:  buildID,
    Identity: identity,
})
if err != nil {
    if !errors.As(err, &notFound) {
        l.Info("could not delete worker deployment version, keeping its k8s Deployment to reconcile",
            "buildID", buildID, "deployment", d.Name, "error", err)
        continue
    }
    ...
}
```

1. **Only `NotFound` is classified.** Every other error — including a permanent `Internal` — is implicitly treated as transient and retried indefinitely.
2. **`deleteDrainedVersions` returns nothing**, so the error never propagates to `executePlan`'s caller. The controller already has backoff for exactly this shape of problem — `worker_controller.go:358` and `:406` map `*serviceerror.ResourceExhausted` to `RequeueAfter: 30s` (added in #291 for #287) — but delete failures can never reach it.
3. Retries therefore land at the base reconcile cadence (`worker_controller.go:443`, `RequeueAfter: 10s`).

Worth noting for anyone fixing this: the existing `withBackoff(timeout, tick, fn)` helper (`internal/temporal/worker_deployment.go:305`) is a fixed-tick retry-until-deadline loop, not exponential — `withBackoff(10*time.Second, 1*time.Second, ...)` is up to 10 attempts 1s apart. Wrapping the delete call in it would multiply the request volume rather than reduce it. This needs state that survives across reconciles, per (deployment, buildID).

## Suggested fix

- Track consecutive delete failures per (deployment, buildID) and escalate the requeue interval.
- Either return the error from `deleteDrainedVersions` so the existing `ResourceExhausted`/requeue handling applies, or classify explicitly at the call site.
- After N consecutive failures, stop retrying and surface a condition or Event on the CR. A version whose workflow has already completed will never become deletable, so a permanently-nominated version should be visible in status rather than only in logs.

## Impact if unfixed

- Sustained failing RPC volume against the frontend (19.2/min here, scaling with the number of affected deployments).
- The affected version records can never be pruned, which feeds the per-deployment version-cap pressure that #377 is about.
- The corresponding versioned k8s Deployments are retained indefinitely (harmless at `replicas: 0`, but they accumulate).

## Note on #555

This is not #555. That one is a *transient* `RoutingConfig` vs `VersionSummaries` race with a bounded window. This inconsistency is persistent — 17 hours and counting — and the retry behaviour means the controller never stops hammering it.

tomba7:
Fixes #583

## What was changed

deleteDrainedVersions now rate limits the call to DeleteWorkerDeploymentVersion
per (worker deployment, build ID). Every failure doubles the delay, from 10s up
to a 30m cap, and is cleared on success.

## Why?

Any error other than NotFound was retried on every reconcile, with no attempt
counter and no backoff/delay. In the scenario the reporter hit, the retries never
decays and keeps going at the reconcile cadence for as long as the CR exists.

Per stuck version it translates to one attempt per reconcile, which is every ~10-19s
in practice indefinitely (8640 times a day if we assume a perfect reconcile cadence
of 10s). The 30m cap brings it down to about 55 a day.

This issue was triggered by a server side bug, but the controller still needs to be able to cope
with a delete that keeps on failing, and that is what this change fixes.

## Testing
- Unit Tests
	- TestExecutePlan_VersionDeletionFails_KeepsDeploymentAndRetriesAfterBackoff
	- TestVersionDeleteBackoffSchedule
- Live Cluster
	-  Verified locally on a kind cluster running Temporal OSS 1.29.1 in-cluster with an Envoy
	   fault proxy in front of the frontend aborting `DeleteWorkerDeploymentVersion`
           with gRPC `Internal`
	- The backoff escalated 10s, 20s, 40.1s, 80.1s, 160.4s while reconciles continued every
	  10s, and once the fault was cleared the next attempt succeeded and pruned both the
           server side version and its k8s Deployment.


jaypipes (approved):
Fantastic use of the `client-go/util/flowcontrol` library. Bravo @tomba7.

PR #611, opened 2026-09-19
feat(wrt): strip temporal_ prefix from HPA metric matchLabels when en…

### LINKED ISSUE #588: [Feature Request] Support `worker_version` selectors in `WorkerResourceTemplate`

We run a self-hosted Temporal Server and use `WorkerResourceTemplate` to create one HPA per worker version, using the Temporal Server's `approximate_backlog_count` metric through Datadog.

The Temporal Server already exports `worker_version` on the backlog metric. However, the controller [injects only `temporal_worker_deployment_name`, `temporal_worker_build_id`, and `temporal_namespace`](https://github.com/temporalio/temporal-worker-controller/blob/57df3a2d159153ff723b3db7d218816ba04be1ce/internal/k8s/workerresourcetemplates.go#L101). We currently rename the server's separate deployment/build labels in the collector to match those selectors.

Ideally, metric selector injection should be configurable, so Temporal Worker Controller's users can choose the label names and how their values are derived from the worker deployment, build ID, and Temporal namespace. Supporting `worker_version` is our immediate use case, but a general configuration mechanism would also accommodate other collector and metrics-adapter conventions.

For our setup, this should provide an opt-in way to generate a `worker_version` selector from the worker version associated with each rendered resource.

### Additional context

- Controller Helm chart: `0.29.1`; self-hosted Temporal Server: `1.31.0`.
- Temporal Server [constructs the deployment-version metric tag from the deployment name and build ID](https://github.com/temporalio/temporal/blob/v1.31.0/service/matching/physical_task_queue_key.go#L194).

pranavthakur0-0:
Fixes #588

## What was changed

- Added default-off controller flag `--wrt-hpa-match-labels-strip-temporal-prefix` and env `WRT_HPA_MATCH_LABELS_STRIP_TEMPORAL_PREFIX`.
- Added Helm value `workerResourceTemplate.hpaMatchLabelsStripTemporalPrefix` (schema + manager Deployment env).
- When enabled, auto-injected WorkerResourceTemplate HPA external metric `matchLabels` use `worker_deployment_name`, `worker_build_id`, and `namespace` instead of `temporal_worker_deployment_name`, `temporal_worker_build_id`, and `temporal_namespace`.
- Plumbed the setting through reconciler → planner → WRT renderer; webhook rejects the unprefixed keys when stripping is enabled (same controller-owned guarantee as `temporal_*`).
- Documented the Helm setting in `docs/worker-resource-templates.md`.
- Unit tests (env parsing, render/planner, webhook) and Helm validate workflow checks for the new env default and `--set true`.

## Why?

Self-hosted Temporal Server OpenMetrics use native label names (`worker_deployment_name`, etc.). Temporal Cloud metrics use the `temporal_*` prefix. Today the controller always injects `temporal_*`, which breaks HPA external metrics against self-hosted backends unless operators rename labels out of band. This implements the maintainer plan for #588: optional prefix stripping on the three hardcoded injected keys only, without changing KEDA triggers, pod selectors, or metric values.

## Testing

- `go test ./api/v1alpha1 ./internal/k8s ./internal/planner ./internal/controller`
- Helm validate CI / `helm template` asserts `WRT_HPA_MATCH_LABELS_STRIP_TEMPORAL_PREFIX` defaults to `false` and renders `true` when the value is enabled


jaypipes (approved):
I like it. Thank you @pranavthakur0-0!

jaypipes (approved):
👍

pranavthakur0-0:
Thanks @jaypipes

